PR #41, opened 2018-08-08
Local scoring (aka Sparkless) using Aardpfark

tovbinm:
Added a subproject that enables loading and scoring models without Spark context but locally using Aardpfark (PFA for Spark) and Hadrian libraries instead. This allows orders of magnitude faster scoring times compared to Spark.

dbml-local, ml-local and a custom runtime.

**Additional Context**
Sample usage:
```scala
import com.salesforce.op.local._
val model = workflow.loadModel("/path/to/model")
val scoreFn = model.scoreFunction
val score: Map[String, Any] = scoreFn(Map("age" -> 18, "name" -> "Peter"))
```
Test results (single thread running on MacBook Pro i7 3.5Ghz):
```
Scored 6000000 records in 239s
Average time per record: 0.0399215ms
```

**TODO**: 
- add a type to ScoreFunction
- consider using [mleap](https://github.com/combust/mleap).


local/src/main/scala/com/salesforce/op/local/OpWorkflowRunnerLocal.scala
```diff
@@ -0,0 +1,98 @@
...
+
+          case (e: PFAEngine[AnyRef, AnyRef], i) =>
+            val stage = stagesWithIndex.find(_._2 == i).map(_._1.asInstanceOf[OpPipelineStage[_]]).get
+            val outName = stage.getOutputFeatureName
+            val inputName = stage.getInputFeatures().collect {
+              case f if f.isSubtypeOf[OPVector] => f.name
+            }.head
+            val vector = r(inputName).asInstanceOf[Vector].toArray
+            val input = s"""{"$inputName":${vector.mkString("[", ",", "]")}}"""
+            val res = e.action(e.jsonInput(input)).toString
```
tovbinm: @MLnick is using json is the most efficient way to call engine action?

manuzhang:
Have you looked at https://github.com/combust/mleap ?

tovbinm:
@manuzhang last time I checked spark support wast that great, but it seems better now. Whats your experience with it?

manuzhang:
@tovbinm We find MLeap supports more models than Aardpfark and have been using it to export pipeline stages into a single online model for Spark 2.2.0 and 2.3.1

features/src/main/scala/com/salesforce/op/utils/spark/RichRow.scala
```diff
@@ -116,6 +116,27 @@ object RichRow {
     def getFeatureType[T <: FeatureType](f: TransientFeature)(implicit conv: FeatureTypeSparkConverter[T]): T =
       conv.fromSpark(getAny(f.name))
 
+    /**
+     * Converts row to a [[collection.mutable.Map]]
+     *
+     * @return a [[collection.mutable.Map]] with row contents
+     */
+    def toMutableMap: collection.mutable.Map[String, Any] = {
```
leahmcguire: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/Row.scala#L354

tovbinm: Oook, so my function is faster, because `getValuesMap` calls `def getAs[T](fieldName: String): T = getAs[T](fieldIndex(fieldName))` for each value, while my function operates on indices.

local/src/main/scala/com/salesforce/op/local/OpWorkflowRunnerLocal.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Load the model & prepare a score function for local scoring
+   *
+   * @param params params to use during scoring
+   * @return score function for local scoring
+   */
+  def score(params: OpParams): ScoreFunction = {
+    require(params.modelLocation.isDefined, "Model location must be set in params")
+    val model = workflow.loadModel(params.modelLocation.get)
```
leahmcguire: will the standard load method work on spark models that use parquet storage without a spark context?

tovbinm: None of the spark ml readers require the context explicitly, but I will need to verify, cause they might get/create spark context inside. Do you have a model in mind that I can check against?

local/src/test/scala/com/salesforce/op/local/OpWorkflowRunnerLocalTest.scala
```diff
@@ -0,0 +1,120 @@
...
+class OpWorkflowRunnerLocalTest extends FlatSpec with PassengerSparkFixtureTest with TestCommon {
+
+  val log = LoggerFactory.getLogger(this.getClass)
+
+  val features = Seq(height, weight, gender, description, age).transmogrify()
+  val survivedNum = survived.occurs()
+
+  // TODO: remove .map[Text] once Aardpfark supports null inputs for StringIndexer
+  val indexed = description.map[Text](v => if (v.isEmpty) Text("") else v)
+    .indexed(handleInvalid = StringIndexerHandleInvalid.Skip)
```
leahmcguire: We should add a spark model that uses complicated serialization - maybe PCA since that uses parquet

local/src/main/scala/com/salesforce/op/local/OpWorkflowRunnerLocal.scala
leahmcguire: maybe try PCA

tovbinm: oh snap, they simply create a spark context internally when loading models 🤦‍♂️ https://github.com/apache/spark/blob/5264164a67df498b73facae207eda12ee133be7d/mllib/src/main/scala/org/apache/spark/ml/feature/PCA.scala#L212

tovbinm: well, we also use spark context when reading the model & stages - https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala#L61
and
https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/stages/OpPipelineStageReader.scala#L63

build.gradle
```diff
@@ -88,6 +88,8 @@ configure(allProjs) {
         commonsValidatorVersion = '1.6'
         commonsIOVersion = '2.6'
         scoveragePluginVersion = '1.3.1'
+        hadrianVersion = '0.8.5'
+        aardpfarkVersion = '0.1.0-SNAPSHOT'
```
leahmcguire: why are we pulling in a shapshot?

PR #55, opened 2018-08-14
New model selector interface

leahmcguire:
**Related issues**
Changing model selector interface to use new more flexible model selector class.

Changing model selector interface to use new more flexible model selector class.

Leaving both interfaces. It would be confusing.

**Additional context**


michaelweilsalesforce:
@leahmcguire This is not be related to this PR, but I'm trying to understand the type safety in the models to try.
Here is an example : What happens if a (witty) user decides to do something this : 
`**BinaryClassification**ModelSelector.withCrossValidation(modelsAndParameters  = Seq(new **OpLinearRegression()** -> Array.empty)...` ?
In a word, it is a ClassificationModelSelector trying a regression model. Will it break? Will it do a kind of regression Model selector?

Should we consider in the future enforcing a type 'Classification' for `OpLogisticRegression`, `OpRandomForest`,... and a type `Regression` for `OpLinearRegression`, ... ?

leahmcguire:
It is an not part of this PR but you are correct @michaelweilsalesforce there is no compile time type check for this now. It will fail at runtime because the evaluator will not find a raw prediction / probability. In order to support users being able to define their own estimators I had to relax the type checks. Any estimator that takes a label and feature vector and returns a prediction will try to run. So the default models and models that can be turned on by name are all of the correct type. A user can mess it up if they try :-P Good eye :-)

core/src/main/scala/com/salesforce/op/evaluators/OpEvaluatorBase.scala
```diff
@@ -61,28 +61,28 @@ trait OpHasLabelCol[T <: FeatureType] extends Params {
...
 
 /**
- * Trait for predictionCol param
+ * Trait for internal flattened predictionCol param
  */
-trait OpHasPredictionCol[T <: FeatureType] extends Params {
-  final val predictionCol: Param[String] = new Param[String](this, "predictionCol", "prediction column name")
-  setDefault(predictionCol, "prediction")
+trait OpHasPredictionValueCol[T <: FeatureType] extends Params {
+  final val predictionValueCol: Param[String] = new Param[String](this, "predictionCol", "prediction column name")
```
kinfaikan: `"predictionCol"` -> `"predictionValueCol"`?

core/src/main/scala/com/salesforce/op/evaluators/OpEvaluatorBase.scala
```diff
@@ -61,28 +61,28 @@ trait OpHasLabelCol[T <: FeatureType] extends Params {
 /**
  * Trait for predictionCol which contains all output results param
  */
-trait OpHasFullPredictionCol extends Params {
-  final val fullPredictionCol: Param[String] = new Param[String](this, "fullPredictionCol", "prediction column name")
+trait OpHasPredictionCol extends Params {
+  final val predictionCol: Param[String] = new Param[String](this, "fullPredictionCol", "prediction column name")
```
kinfaikan: Update name and doc of the param?
Add `setDefault`?

core/src/main/scala/com/salesforce/op/evaluators/OpEvaluatorBase.scala
```diff
@@ -194,17 +194,17 @@ private[op] abstract class OpClassificationEvaluatorBase[T <: EvaluationMetrics]
    * @return data formatted for use with the evaluator
    */
   protected def makeDataToUse(data: Dataset[_], labelColName: String): Dataset[_] = {
-    if (isSet(fullPredictionCol) &&
-      !(isSet(predictionCol) && data.schema.fieldNames.contains(getPredictionCol))) {
-      val fullPredictionColName = getFullPredictionCol
+    if (isSet(predictionCol) &&
+      !(isSet(predictionValueCol) && data.schema.fieldNames.contains(getPredictionValueCol))) {
```
kinfaikan: `data.columns.contains(getPredictionValueCol)`

core/src/main/scala/com/salesforce/op/evaluators/OpEvaluatorBase.scala
```diff
@@ -253,11 +253,11 @@ abstract class OpRegressionEvaluatorBase[T <: EvaluationMetrics]
...
```
kinfaikan: `data.columns.contains(getPredictionValueCol)`

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -97,6 +169,13 @@ case object BinaryClassificationModelSelector {
    *                            Caution : setting that param to true might impact the runtime
    * @param parallelism         level of parallelism used to schedule a number of models to be trained/evaluated
    *                            so that the jobs can be run concurrently
+   * @param modelTypesToUse     list of model types to run grid search on must from supported types in
+   *                            BinaryClassificationModelsToTry (OpLogisticRegression, OpRandomForestClassifier,
+   *                            OpGBTClassifier, OpLinearSVC, OpDecisionTreeClassifier, OpNaiveBayes)
+   * @param modelsAndParameters pass in an explicit list pairs of estimators and the accompanying hyper parameters to
```
kinfaikan: hyper parameters -> hyperparameters

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -106,90 +185,50 @@ case object BinaryClassificationModelSelector {
...
-  ): BinaryClassificationModelSelector = {
-    new BinaryClassificationModelSelector(
+    trainTestEvaluators: Seq[OpBinaryClassificationEvaluatorBase[_ <: EvaluationMetrics]],
+    modelTypesToUse: Seq[BinaryClassificationModelsToTry],
+    modelsAndParameters: Seq[(EstimatorType, Array[ParamMap])]
+  ): ModelSelector[ModelType, EstimatorType] = {
+    val modelStrings = modelTypesToUse.map(_.entryName)
+    val modelsToUse =
+      if (modelsAndParameters == defaultModelsAndParams || modelTypesToUse != modelNames) modelsAndParameters
+        .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
```
kinfaikan: To use a proper subset of the default models, one has to specify `modelsAndParameters` explicitly?

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -106,90 +185,50 @@ case object BinaryClassificationModelSelector {
...
-  validator: OpValidator[ProbClassifierModel, ProbClassifier],
-  splitter: Option[Splitter],
-  evaluators: Seq[OpBinaryClassificationEvaluatorBase[_ <: EvaluationMetrics]],
-  uid: String = UID[Stage1BinaryClassificationModelSelector],
-  stage2uid: String = UID[SwTernaryTransformer[_, _, _, _, _]],
-  stage3uid: String = UID[SwQuaternaryTransformer[_, _, _, _, _, _]]
-) extends Stage1ClassificationModelSelector(validator, splitter, evaluators, uid, stage2uid, stage3uid)
+object BinaryClassificationModelsToTry extends Enum[BinaryClassificationModelsToTry] {
+  val values = findValues
+  case object OpLogisticRegression extends BinaryClassificationModelsToTry
```
kinfaikan: Should we associate each BinaryClassificationModelsToTry with the corresponding estimator and default params?
```
case object OpLogisticRegression extends BinaryClassificationModelsToTry {
  val estimator = new OpLogisticRegression()
  val params = new ParamGridBuilder()
      .addGrid(estimator.fitIntercept, DefaultSelectorParams.FitIntercept)
      .addGrid(estimator.elasticNetParam, DefaultSelectorParams.ElasticNet)
      .addGrid(estimator.maxIter, DefaultSelectorParams.MaxIterLin)
      .addGrid(estimator.regParam, DefaultSelectorParams.Regularization)
      .addGrid(estimator.standardization, DefaultSelectorParams.Standardized)
      .addGrid(estimator.tol, DefaultSelectorParams.Tol)
      .build()
}
```
Also, we might let users define custom BinaryClassificationModelsToTry?

core/src/main/scala/com/salesforce/op/evaluators/OpEvaluatorBase.scala
leahmcguire: yes and `fullPredictionCol` -> `prediction col`

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
leahmcguire: I can make the base class public for BinaryClassificationModelsToTry. And I could add a sub value to have the class they are associated with rather than relying on the name...

core/src/main/scala/com/salesforce/op/evaluators/OpEvaluatorBase.scala
leahmcguire: there is no default that can be set for this

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorNames.scala
```diff
@@ -0,0 +1,121 @@
...
+  val Desired = "desiredFraction"
+  val UpSample = "upSamplingFraction"
+  val DownSample = "downSamplingFraction"
+  val idColName = "rowId"
+  val LabelsKept = "labelsKept"
+  val LabelsDropped = "labelsDropped"
+
+  type ModelType = Model[_ <: Model[_]] with OpTransformer2[RealNN, OPVector, Prediction]
+  type EstimatorType = Estimator[_ <: Model[_]] with OpPipelineStage2[RealNN, OPVector, Prediction]
+
```
tovbinm: @leahmcguire totally minor - wanna add `type ModelSelector = ModelSelector[ModelType, EstimatorType]`?

PR #62, opened 2018-08-16
Added Memory Appender, Log assertions in OpSparkListner

ajayborra:
- Added Memory Appender, Log assertions for OpSparkListner

### DIFF
--- a/utils/src/main/scala/com/salesforce/op/test/MemoryAppender.scala
+++ b/utils/src/main/scala/com/salesforce/op/test/MemoryAppender.scala
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2017, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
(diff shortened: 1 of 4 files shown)


utils/src/main/scala/com/salesforce/op/test/MemoryAppender.scala
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+package com.salesforce.op.test
+
+import org.apache.log4j.{AppenderSkeleton, Level, spi}
+
+import scala.collection.mutable
+
+class MemoryAppender extends AppenderSkeleton {
+  private val _logs = new mutable.HashSet[spi.LoggingEvent]
```
tovbinm: 1. no need to prefix with `_`
2. make this an `ArrayBuffer`

utils/src/main/scala/com/salesforce/op/test/MemoryAppender.scala
```diff
@@ -0,0 +1,74 @@
+/*
```
tovbinm: define this appender in the same file as `OpSparkListenerTest` - we dont need to expose it as it's currently only useful in one test.

utils/src/main/scala/com/salesforce/op/test/TestSparkContext.scala
```diff
@@ -44,6 +44,15 @@ trait TestSparkContext extends TempDirectoryTest with TestCommon {
     classOf[com.salesforce.op.test.PassengerCSV]
   )
 
+  val sparkLogAppender: MemoryAppender = {
```
tovbinm: we dont need this in every test. let's only use it in OpSparkListenerTest

utils/src/test/resources/log4j.properties
```diff
@@ -12,3 +12,4 @@ log4j.logger.org.eclipse.jetty=ERROR
 
 # TransmogrifAI logging
 log4j.logger.com.salesforce.op=ERROR
+log4j.logger.com.salesforce.op.utils.spark.OpSparkListener=INFO
```
tovbinm: you can set this `log4j.logger.com.salesforce.op.utils.spark.OpSparkListener=INFO` programmatically in OpSparkListenerTest

utils/src/main/scala/com/salesforce/op/test/MemoryAppender.scala
```diff
@@ -0,0 +1,74 @@
...
+   */
+  def logs: mutable.HashSet[spi.LoggingEvent] = _logs
+
+  /**
+   * Check if a log exists in the log collection
+   * @param logLevel The log level of the message
+   * @param logMessage The log message
+   * @return Boolean of log existence
+   */
+  def logExists(logLevel: Level, logMessage: String): Boolean = {
```
tovbinm: this seems a bit redundant. you can simply do this check in the test itself.

utils/src/main/scala/com/salesforce/op/test/MemoryAppender.scala
ajayborra: Was using hash sets for constant inset and lookup times. Array buf's seems to be a good choice as well, will change it.

tovbinm: Hashset would dedup and loose order, thereforce we cannot use it.

ajayborra: Good point.

ajayborra:
Closing this PR in favor of https://github.com/salesforce/TransmogrifAI/pull/69. As the link b/w this PR and the branch on old fork broke, post making repo public.

PR #67, opened 2018-08-17
Improved Coverage for RichDataType

himsmittal:
- Improved Coverage for RichDataType
Added test case for checking nullability for ArrayType , MapType

### DIFF
--- a/features/src/test/scala/com/salesforce/op/utils/spark/RichDatasetTest.scala
+++ b/features/src/test/scala/com/salesforce/op/utils/spark/RichDatasetTest.scala
@@ -234,6 +234,16 @@ class RichDatasetTest extends FlatSpec with TestSparkContext {
     in(ds) noneOf "f2" should ((x: Double) => x > 3.5)
   }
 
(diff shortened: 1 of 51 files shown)


features/src/test/scala/com/salesforce/op/utils/spark/RichDatasetTest.scala
```diff
@@ -234,6 +234,16 @@ class RichDatasetTest extends FlatSpec with TestSparkContext {
     in(ds) noneOf "f2" should ((x: Double) => x > 3.5)
   }
 
+  it should "only compare types and ignore nullability" in {
```
tovbinm: Minor correction: lets replace `it` in this test case with `Spec(com.salesforce.op.utils.spark.RichDataType.getClass)`.

himsmittal: Changed usage of "it" to reflect RichDataType.
Thanks

PR #71, opened 2018-08-20
Adding unit test for FeatureHistory

gokulsfdc:
- Added unit tests for FeatureHistory class

### DIFF
--- a/utils/src/test/scala/com/salesforce/op/FeatureHistoryTest.scala
+++ b/utils/src/test/scala/com/salesforce/op/FeatureHistoryTest.scala
@@ -45,7 +45,7 @@ class FeatureHistoryTest extends FlatSpec with TestCommon {
   val stage1 = "stage1"
   val stage2 = "stage2"
 
-  Spec(FeatureHistory.getClass) should "convert a feature history to metadata" in {
+  Spec[FeatureHistory] should "convert a feature history to metadata" in {
     val featureHistory = FeatureHistory(originFeatures = Seq(feature1, feature2), stages = Seq(stage1, stage2))
 
     val featureHistoryMetadata = featureHistory.toMetadata
@@ -57,7 +57,7 @@ class FeatureHistoryTest extends FlatSpec with TestCommon {
     featureHistoryMetadata.getStringArray(StagesKey) shouldBe Array(stage1, stage2)
   }
 
-  it should "merge two featurehistory" in {
+  it should "merge two instances" in {
     val featureHistory1 = FeatureHistory(originFeatures = Seq(feature1), stages = Seq(stage1))
     val featureHistory2 = FeatureHistory(originFeatures = Seq(feature2), stages = Seq(stage2))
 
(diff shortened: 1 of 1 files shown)


utils/src/test/scala/com/salesforce/op/FeatureHistoryTest.scala
```diff
@@ -0,0 +1,115 @@
...
+
+@RunWith(classOf[JUnitRunner])
+class FeatureHistoryTest extends FlatSpec with TestCommon {
+
+  val feature1 = "feature1"
+  val feature2 = "feature2"
+  val stage1 = "stage1"
+  val stage2 = "stage2"
+
+  Spec(FeatureHistory.getClass) should "convert a feature history to metadata" in {
```
tovbinm: Since there is an actual class you can do: Spec[FeatureHistory] instead

utils/src/test/scala/com/salesforce/op/FeatureHistoryTest.scala
```diff
@@ -0,0 +1,115 @@
...
+
+    featureHistoryMetadata.contains(OriginFeatureKey) shouldBe true
+    featureHistoryMetadata.contains(StagesKey) shouldBe true
+
+    featureHistoryMetadata.getStringArray(OriginFeatureKey) shouldBe Array(feature1, feature2)
+    featureHistoryMetadata.getStringArray(StagesKey) shouldBe Array(stage1, stage2)
+  }
+
+  it should "merge two featurehistory" in {
+    val featureHistory1 = FeatureHistory(originFeatures = Seq(feature1), stages = Seq(stage1))
```
tovbinm: No need to repeat featurehistory in the test name. Simply "merge two instances"

utils/src/test/scala/com/salesforce/op/FeatureHistoryTest.scala
```diff
@@ -0,0 +1,115 @@
...
+    val featureHistoryCombined = featureHistory1.merge(featureHistory2)
+    featureHistoryCombined.originFeatures shouldBe Seq(feature1, feature2)
+    featureHistoryCombined.stages shouldBe Seq(stage1, stage2)
+  }
+
+  it should "create a metadata for a featurehistory map" in {
+    val featureHistory1 = FeatureHistory(originFeatures = Seq(feature1), stages = Seq(stage1))
+    val featureHistory2 = FeatureHistory(originFeatures = Seq(feature2), stages = Seq(stage2))
+
+    val map = Map[String, FeatureHistory](("1" -> featureHistory1), ("2" -> featureHistory2))
```
tovbinm: No need to specify map type.

utils/src/test/scala/com/salesforce/op/FeatureHistoryTest.scala
tovbinm: Same in other tests @gokulsfdc

tovbinm:
Please remove “featurehistory” from test names.

PR #72, opened 2018-08-20
Adding Array of metadataSeq in RichMap

gokulsfdc:
In RichMap, for Array values, we are currently accepting booleanSeq, intSeq, longSeq, doubleSeq and stringSeq. 
- Adding one more datatype to the list - MetadataSeq.
- Adding test case for the new type added.
- Adding test case for withSummary() as well for improved coverage

### DIFF
--- a/utils/src/test/scala/com/salesforce/op/utils/spark/RichMetadataTest.scala
+++ b/utils/src/test/scala/com/salesforce/op/utils/spark/RichMetadataTest.scala
@@ -85,6 +85,11 @@ class RichMetadataTest extends FlatSpec with TestCommon {
     val mergedMetadata = meta1.deepMerge(map2.toMetadata)
     mergedMetadata.json shouldBe Serialization.write(mergedMap)
 
+    val m1 = Map("1" -> Array(Map("val" -> "a").toMetadata)).toMetadata
+    val m2 = Map("1" -> Array(Map("val" -> "b").toMetadata)).toMetadata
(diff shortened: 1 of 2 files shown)


utils/src/test/scala/com/salesforce/op/utils/spark/RichMetadataTest.scala
```diff
@@ -85,6 +85,11 @@ class RichMetadataTest extends FlatSpec with TestCommon {
     val mergedMetadata = meta1.deepMerge(map2.toMetadata)
     mergedMetadata.json shouldBe Serialization.write(mergedMap)
 
+    val m1 = Map("1" -> Array(Map("val" -> "a").toMetadata)).toMetadata
```
tovbinm: Lets add it as a separate test case

utils/src/test/scala/com/salesforce/op/utils/spark/RichMetadataTest.scala
```diff
@@ -100,6 +105,17 @@ class RichMetadataTest extends FlatSpec with TestCommon {
     summaryMeta.containsSummaryMetadata shouldBe true
   }
 
+  it should "create summary for a given metadata" in {
+    val richMetaData = RichMetadata(meta1)
```
tovbinm: No need to explicitely wrap metadata into RichMetadata. It should world implicitly.

tovbinm:
Ooops, I should have waited for the tests to complete.

PR #76, opened 2018-08-21
Lm/keep dist

leahmcguire:
**Related issues**
Keeping feature distribution information calculated in raw feature filter for use in metrics logging and feature engineering

When raw feature filter is called it now returns the feature distribution information calculated on the training data. These distributions are kept in the workflow and added to the raw features so that they can be accessed later.

Could just write the information out directly from raw feature filter.


features/src/main/scala/com/salesforce/op/features/FeatureDistributionBase.scala
```diff
@@ -0,0 +1,68 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.salesforce.op.features
+
+
+/**
+ * Keeps the distribution information for features
+ */
+trait FeatureDistributionBase {
```
tovbinm: @leahmcguire why do we need the trait? if we do, let's name it `FeatureDistributionLike`.

leahmcguire: ok - It seemed nicer to make the trait than to move all of the classes in raw feature filter here. though I am willing to hear counter arguments :-P

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -108,7 +107,7 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
    * Will set the blacklisted features variable and if list is non-empty it will
    * @param features list of features to blacklist
    */
-  private[op] def setBlacklist(features: Array[OPFeature]): Unit = {
+  private[op] def setBlacklist(features: Array[OPFeature], distributions: Seq[FeatureDistribution]): Unit = {
```
kinfaikan: Please update doc.

features/src/main/scala/com/salesforce/op/features/FeatureDistributionBase.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * name of the feature
+   */
+  val name: String
+
+  /**
+   * key map key associated with distribution (when the feature is a map)
```
kinfaikan: `key map key` -> `map key`

features/src/main/scala/com/salesforce/op/features/FeatureDistributionBase.scala
```diff
@@ -0,0 +1,68 @@
...
+   */
+  val count: Long
+
+  /**
+   * number of empties seen in feature
+   */
+  val nulls: Long
+
+  /**
+   * binned counts of feature values (hashed for strings, evently spaced bins for numerics)
```
kinfaikan: `evently` -> `evenly`

features/src/main/scala/com/salesforce/op/features/FeatureDistributionBase.scala
tovbinm: Oh, I missed the part that it's located in a separate project. Makes sense then.
Let's rename it then to match our convention and also lets replace vals in the trait with defs.

PR #88, opened 2018-08-24
Stable key order for map vectorizers

tovbinm:
**Related issues**
Map vectorizers do not order the keys properly.

- Sorted map keys in `MapVectorizerFuns.getKeyValues` function.
- bonus: use op specs base classes for the vectorizers

N/A


core/src/main/scala/com/salesforce/op/stages/impl/feature/OPMapVectorizer.scala
```diff
@@ -124,14 +123,14 @@ class IntegralMapVectorizer[T <: OPMap[Long]](uid: String = UID[IntegralMapVecto
   def setFillWithMode(shouldFill: Boolean): this.type = set(withConstant, !shouldFill)
 
   override def fillByKey(dataset: Dataset[Seq[T#Value]]): Seq[Map[String, Double]] = {
```
gerashegalov: nit: it looks like fillByKey in `RealMapVectorizer` and `IntegralMapVectorizer` have a very small diff. Maybe refactor.

PR #89, opened 2018-08-25
Added value to OpVectorColumnMetadata for numeric column descriptors

leahmcguire:
**Related issues**
We use the indicatorGroup and indicatorValue in OpVectorColumnMetadata not being empty to decide when we need to run sanity checker categorical tests in metadata. This is a problem for cases where we have information about a derived field that we want to represent but the field is not a boolean indicator (we have no place to put them). We have hacked around this by just putting this information in the indicatorValue anyway and not having anything in the indicator group. However this doesn't work if we have the same transformation for maps (which we need to add for feature complete). 

Right now our column metadata is structured like this

```
case class OpVectorColumnMetadata
(
  parentFeatureName: Seq[String],
  parentFeatureType: Seq[String],
  indicatorGroup: Option[String],
  indicatorValue: Option[String],
  index: Int = 0
)
```
 we will add a value to hold numeric column name information and rename indicatorGroup to grouping to make it's usage more clear.
```
case class OpVectorColumnMetadata
(
  parentFeatureName: Seq[String],
  parentFeatureType: Seq[String],
  grouping: Option[String],
  indicatorValue: Option[String] = None, // for categoricals
  descriptorValue: Option[String] = None, // for continuous things that need description 
  // eg geolocation lat, lon, acc or datetime unit circle timeperiod_x timeperiod_y
  index: Int = 0
) {
  assert(indicatorValue.isEmpty || descriptorValue.isEmpty)
}
```

put in an check for parent feature type to exclude specific types from the sanity checker categorical tests


tovbinm:
I propose to keep the name `indicatorGroup` (instead of `grouping`) to allow backward compatibility for existing models.

leahmcguire:
Existing models will not load with this release anyway because of the model selector change

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorColumnMetadata.scala
```diff
@@ -75,6 +79,9 @@ case class OpVectorColumnMetadata
   assert(parentFeatureName.length == parentFeatureType.length,
     s"must provide both type and name for every parent feature," +
       s" names: $parentFeatureName and types: $parentFeatureType do not have the same length")
+  assert(indicatorValue.isEmpty || descriptorValue.isEmpty, "cannot have both an indicatorValue and a," +
```
tovbinm: `cannot have both an indicatorValue and descriptorValue.`

leahmcguire:
@Jauntbox thoughts on indicatorGroup versus grouping?

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorColumnMetadata.scala
tovbinm: Since we are making breaking changes anyways, I keep thinking that a base trait + two specific classes makes even more sense. This way we would avoid asserts like this ^
```scala
trait OpVectorColumnMetadata
(
  def parentFeatureName: Seq[String]
  def parentFeatureType: Seq[String]
  def grouping: Option[String]
  def index: Int
)
case class CategoricalVectorColumnMetadata(
  parentFeatureName: Seq[String],
  parentFeatureType: Seq[String],
  grouping: Option[String],
  indicatorValue: Option[String],
  index: Int = 0
) extends OpVectorColumnMetadata

case class ContinuousVectorColumnMetadata(
  parentFeatureName: Seq[String],
  parentFeatureType: Seq[String],
  grouping: Option[String],
  descriptorValue: Option[String]
  index: Int = 0
) extends OpVectorColumnMetadata
```

leahmcguire: this would break our internal usage

leahmcguire: I will make a separate ticket do do this as part of metadata restructuring

Jauntbox:
I don't think we should continue using `indicatorGroup` since we want a more general grouping term that's appropriate for both indicator groups (those filled with binary indicator columns) and other groups (eg. x/y coords for datetimes, stuff we may add in the future like distances to pre-determined cluster centroids). I think `grouping` is a good general replacement.

The only case I have questions about now is how we would deal with cases where we may split a single feature into multiple groupings. I can't think of any examples where we'd want multiple indicator groups or multiple numeric groups, but we might have some in the future. One way is to generate multiple different separate `grouping` values from the original feature based on the transformation applied. I'm assuming that's the way we'd go?

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
```diff
@@ -465,10 +465,10 @@ class SanityChecker(uid: String = UID[SanityChecker])
 
       // Only perform Cramer's V calculation on columns that have an indicatorGroup and indicatorValue defined (right
```
Jauntbox: Update comment to match new name

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
```diff
@@ -319,23 +319,6 @@ case object SanityCheckerSummary {
     )
   }
 
-  @deprecated("CategoricalStats replaced by Array[CategoricalGroupStats]", "3.3.0")
```
Jauntbox: yay!

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
tovbinm: @leahmcguire I think this is the only one left. try grepping the codebase `indicatorGroup`

PR #93, opened 2018-08-27
updated string indexer enum to match spark

leahmcguire:
**Related issues**
OpStringIndexer fails when NoFilter is set for handle invalid because this value doesnt exist in spark

Update enum to match spark supported valuese

### DIFF
--- a/core/src/main/scala/com/salesforce/op/dsl/RichTextFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichTextFeature.scala
@@ -256,10 +256,10 @@ trait RichTextFeature {
      */
     def indexed(
       unseenName: String = OpStringIndexerNoFilter.UnseenNameDefault,
-      handleInvalid: StringIndexerHandleInvalid = StringIndexerHandleInvalid.NoFilter
+      handleInvalid: StringIndexerHandleInvalid = StringIndexerHandleInvalid.Keep
     ): FeatureLike[RealNN] = {
       handleInvalid match {
-        case StringIndexerHandleInvalid.NoFilter => f.transformWith(
+        case StringIndexerHandleInvalid.Keep => f.transformWith(
           new OpStringIndexerNoFilter[T]().setUnseenName(unseenName)
         )
         case _ => f.transformWith(new OpStringIndexer[T]().setHandleInvalid(handleInvalid))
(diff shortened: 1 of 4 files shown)


core/src/test/scala/com/salesforce/op/stages/impl/feature/OpStringIndexerTest.scala
```diff
@@ -0,0 +1,57 @@
...
+import com.salesforce.op.test.TestSparkContext
+import org.apache.spark.ml.feature.StringIndexer
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class OpStringIndexerTest extends FlatSpec with TestSparkContext{
+
+  Spec[OpStringIndexer[_]] should "correctly set the wrapped spark stage params" in {
```
tovbinm: please move `OpStringIndexer` test portion from `OpStringIndexerNoFilterTest` to here as well

core/src/main/scala/com/salesforce/op/dsl/RichTextFeature.scala
```diff
@@ -256,7 +256,7 @@ trait RichTextFeature {
      */
     def indexed(
       unseenName: String = OpStringIndexerNoFilter.UnseenNameDefault,
-      handleInvalid: StringIndexerHandleInvalid = StringIndexerHandleInvalid.NoFilter
+      handleInvalid: StringIndexerHandleInvalid = StringIndexerHandleInvalid.Keep
```
tovbinm: why change the default?

core/src/main/scala/com/salesforce/op/stages/impl/feature/OpStringIndexer.scala
```diff
@@ -32,6 +32,7 @@ package com.salesforce.op.stages.impl.feature
 
 import com.salesforce.op.UID
 import com.salesforce.op.features.types._
+import com.salesforce.op.stages.impl.feature.StringIndexerHandleInvalid.{Keep, Skip}
```
tovbinm: I think you can do `import com.salesforce.op.stages.impl.feature.StringIndexerHandleInvalid._`
then `assert(Seq(Skip, Error, Keep).contains(value), ...)`

Issue #90, opened 2018-08-25
Build & tests parallelism - Part I


tovbinm:
**Problem**
Currently our build & tests are sequential, perhaps it would be better to have it parallelized.

**Solution**
- Parallelize build & tests using Gradle (if not possible, leverage CircleCI parallelism)
- Evaluate performance difference

**Alternatives**
N/A

ajayborra:
@tovbinm Raised a PR with optimization to gradle. We could try further optimization on CircleCI as well.

tovbinm:
thank you!

PR #94, opened 2018-08-27
Enabled parallelization properties for gradle

ajayborra:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/90

Enabling parallel build flag in gradle config.

**Additional context**
- Few minutes improvements on travis build system since Travis seems to be a dual core container.
- Improved local build speeds.
- Using `Runtime.runtime.availableProcessors()` in build.gradle can consume all cores and causes machine hangs on local builds, Hence using `Runtime.runtime.availableProcessors().intdiv(2)` instead.
- Increased test timeout for a flaky test due to parallel exec wait times of test threads on shared h/w resources.

**Build parallelization profile on an 8 core mac**

** Top time consuming tasks**


build.gradle
```diff
@@ -360,6 +360,18 @@ configure(subProjs) {
     }
 }
 
+allprojects {
+    tasks.withType(Test) {
```
tovbinm: let's move this to `gradle/tests.gradle` file.

features/src/test/scala/com/salesforce/op/features/types/FeatureTypeSparkConverterTest.scala
```diff
@@ -85,7 +85,7 @@ class FeatureTypeSparkConverterTest
   }
   property("create a feature type instance and back in a timely fashion") {
     forAllConcurrentCheck[FeatureTypeSparkConverter[_ <: FeatureType]](
-      numThreads = 10, numInvocationsPerThread = 10000, atMost = 10.seconds,
+      numThreads = 10, numInvocationsPerThread = 10000, atMost = 15.seconds,
```
tovbinm: why the change?

utils/src/test/scala/com/salesforce/op/utils/reflection/ReflectionUtilsTest.scala
```diff
@@ -223,7 +223,7 @@ class ReflectionUtilsTest extends FlatSpec with Matchers {
     val elapsedReflect = measure(method.apply().asInstanceOf[Int])
     val actual = measure(myClass.getValuePerf)
 
-    elapsedReflect should be <= 10 * actual
+    elapsedReflect should be <= 15 * actual
```
tovbinm: same here - why the change?

tovbinm:
nice, let's see how well this works 🤞

features/src/test/scala/com/salesforce/op/features/types/FeatureTypeSparkConverterTest.scala
ajayborra: Seems to be that when parallel threads are scheduled for running tests, wait time on the CPU increased a bit. Tested this on my fork with travis and have seen this test taking ~13 sec on few occasions.

gradle/tests.gradle
```diff
@@ -64,3 +64,15 @@ tasks.withType(Test) {
     }
 
 }
+
+allprojects {
```
tovbinm: you dont need an extra section. update the one above `tasks.withType(Test) { ... }`

gradle/tests.gradle
```diff
@@ -23,7 +23,13 @@ tasks.withType(Test) {
         showStackTraces true
         showStandardStreams true
     }
-    maxParallelForks = 1
+    // Default forks config applied to local builds
+    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
```
tovbinm: `Runtime.runtime.availableProcessors() / 2 ?: 1`

gradle/tests.gradle
```diff
@@ -23,7 +23,13 @@ tasks.withType(Test) {
...
+
+    // Forks config for cloud build systems (Eg. travis)
+    if (System.getenv('BUILD_TYPE').equals("CLOUD")) {
```
tovbinm: `System.getenv('BUILD_TYPE') == 'CLOUD'`

tovbinm:
@ajayborra just by looking into build times - I don't see any improvements really ;(

ajayborra:
@tovbinm As described in the additional context section above, Travis/CircleCI seem to be providing dual-core containers so there is not much parallelism we can achieve. We can see improvements on the local machine where we have 4/8 cores with the config in this PR. Also seems like build times are inconsistent, Travis on my fork is giving build time in lower 30's and on this repo it's giving higher 30s
obs

Furthermore, observation based on profiling indicate that our core module is taking up most of the time in the build even on 8 core machines, In the attached image above for example out of 16 min build time 13 min ( 5m20s compile + 7m40s build ) is consumed by core module. So unless we split up these test in core module further, we may not be able to achieve faster builds even on 4+ core machines.

Checking the pricing pages Travis/CircleCI don't seem to be providing large containers with 4 or 8 cores. Alternatively, As you have already recommended on the issue description we can use [CircleCI Parallelization](https://circleci.com/docs/1.0/setting-up-parallelism/) and Travis equivalent [Travis Build Matirx](https://docs.travis-ci.com/user/speeding-up-the-build/) to split the tests on the core module to reduce the 7m40s build time. Not completely sure, But we may be able to improve the 5m20s compile time by setting the following prop `org.gradle.configureondemand=true`.

Let me know your thoughts on how we can proceed with this further!

ajayborra:
Planning to experiment further on Travis build matrix approach since it already configured on my fork and we can extrapolate the ideas to CircleCI. On a similar note, Is there any reason for using the dual build system (both CircleCI & Travis)?

tovbinm:
@ajayborra thank you for looking into this. 

We are using TravisCI and CircleCI for redundancy. So far Travis was more stable, but CircleCI offers more features. Eventually we might want to have only one, but since they are both free - why not ;)

tovbinm:
I can also suggest looking into tuning local Spark context settings, for instance try setting `local[*]` in `TestSparkContext` and see if improves he build times and also setting `spark.default.parallelism` to a higher value.

Another direction is to persist input data for model selectors in tests. For instance in `RegressionModelSelectorTest`:
```scala
val data = sc.parallelize(rawData).toDF("label", "features").persist() // add persist
println("number of partitions: " + data.rdd.getNumPartitions) // also print out number of partitions it should match "spark.default.parallelism"
```

.circleci/config.yml
```diff
@@ -1,28 +1,67 @@
...
 jobs:
-  build:
+  checkout_code_and_dependencies:
+    machine: true
+    working_directory: ~/repo
+    steps:
+      - checkout
+      - run:
+          name: Download project code and dependency jars
+          command: ./gradlew resolveAllDependencies
```
tovbinm: what's the purpose of this step?

.circleci/config.yml
```diff
@@ -31,17 +70,22 @@ jobs:
             find . -type d -regex ".*/build/test-results/testScoverage" -exec bash -c 'f=$(echo {} | cut -d/ -f2); cp -r {} ~/junit.xml/$f' \;
             find . -type d -regex ".*/build/reports/tests/testScoverage" -exec bash -c 'f=$(echo {} | cut -d/ -f2); cp -r {} ~/junit/$f' \;
           when: always
+      - save_cache:
+          key: gcache-{{ .Branch }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "build.gradle" }}-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+          paths:
+            - ~/junit.xml
+            - ~/junit
+  save_assets:
```
tovbinm: `save_assets` tasks seems redundant, since each task uploads it's own assets.

.circleci/config.yml
tovbinm: I think if you do that, then you should:
1. clone the repo
2. resolve dependencies
3. cache the repo
4. cache the gradle dependencies

Then in subsequent steps:
1. restore repo and gradle caches
2. build

tovbinm: I dont think its a good idea to cache test results but better upload it from the task itself.

ajayborra: Good point. Makes sense to upload at task level itself instead of cache, merge and upload.

.circleci/config.yml
```diff
@@ -1,28 +1,69 @@
...
+
 jobs:
-  build:
+  checkout_code_and_dependencies:
+    machine: true
+    working_directory: ~/repo
+    steps:
+      - checkout
+      - run:
+          name: Download project code and dependency jars
```
tovbinm: `name: Clone repo & resolve dependencies`

.circleci/config.yml
```diff
@@ -1,28 +1,69 @@
...
           keys:
+            - repo-{{ .Environment.CIRCLE_SHA1 }}
             - gcache-{{ .Branch }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "build.gradle" }}-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
-
       - run:
           name: Java version
           command: java -version
       - run:
-          name: Build core project
+          name: Test core module
```
tovbinm: `name: Build core project`

.circleci/config.yml
```diff
@@ -1,28 +1,69 @@
...
+    steps:
+      - restore_cache:
+          keys:
+            - repo-{{ .Environment.CIRCLE_SHA1 }}
+            - gcache-{{ .Branch }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "build.gradle" }}-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+      - run:
+          name: Java version
+          command: java -version
+      - run:
+          name: Test non core modules
```
tovbinm: `name: Build non core projects`

.travis.yml
```diff
@@ -9,8 +9,14 @@ env:
     - GRADLE_OPTS="-Xmx256m"
 install:
   - true
-script: 
-  - ./gradlew --no-daemon reportScoverage
+matrix:
+  include:
+    - name: "Core Module Build"
```
tovbinm: `name: "Build core project"`

.travis.yml
```diff
@@ -9,8 +9,14 @@ env:
...
+      env: BUILD_PARAMETERS='core:reportScoverage'
+    - name: "Non-Core Modules Build"
```
tovbinm: `name: "Build non core projects"`

tovbinm:
@ajayborra Travis seem to run sequentially. Dont you need to add something like:
```
 global:
    - CI_NODE_TOTAL=2
  matrix:
    - CI_NODE_INDEX=0
    - CI_NODE_INDEX=1
```
?

ajayborra:
Let me try that options. As a general note, Travis is providing two metrics for build runtime 
```
 Ran for 28 min 29 sec <- this is the actual runtime according to the wall clock.
 Total time 40 min 18 sec <- this is the sum of the time for which all instances ran.
```
On the listing views like build history/pull request, it is showing the `Total time`. When we go to the build details we see the above-mentioned metrics.

tovbinm:
you might be right! CI_NODE_TOTAL etc is related to rspec

tovbinm:
TravisCI looks good! what about CircleCI I dont see report from it for some reason... 🤔

tovbinm:
Awesome work, man!!

PR #99, opened 2018-08-28
Adjust bin values for text features in RFF

sxd929:
**Related issues**
Not enough hashing space for RFF of text may reduce the js distance (weaken the signal)

- Add sum and total count to summary
- Allow user provided method to calculate the bin number based on data summary as parameter to calculate the distribution

Alternative method like Jaccard similarity, word and sentence embedding might be considered

**Additional context**
N/A


core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -182,18 +180,25 @@ private[op] object FeatureDistribution {
...
+    bins: Int
   ): (Array[Double], Array[Double]) = {
     values match {
-      case Left(seq) => Array(sum.min, sum.max) -> hasher.transform(seq).toArray // TODO use summary info to pick hashes
+      case Left(seq) => {
+        val minBins = bins
+        val maxBins = MaxBins
+        val numBins = math.min(math.max(bins, sum.max/AvgBinValue), maxBins).toInt
+
+        val hasher: HashingTF = new HashingTF(numFeatures = numBins)
```
tovbinm: do we have to create hasher every time or perhaps we can create it once?

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -182,18 +180,25 @@ private[op] object FeatureDistribution {
...
-    bins: Int,
-    hasher: HashingTF
+    bins: Int
   ): (Array[Double], Array[Double]) = {
     values match {
-      case Left(seq) => Array(sum.min, sum.max) -> hasher.transform(seq).toArray // TODO use summary info to pick hashes
+      case Left(seq) => {
+        val minBins = bins
+        val maxBins = MaxBins
+        val numBins = math.min(math.max(bins, sum.max/AvgBinValue), maxBins).toInt
```
tovbinm: `sum.max / AvgBinValue`

tovbinm: `.toInt` perhaps specify the explicit rounding instead.

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
sxd929: the hashing dimension can be different for different features, the minimum number would be `bins`
but we can have a shared one with numFeatures = bins, and use that for every case if there is not too many tokens; OR we can create a couple shared hashers with different scales, and choose one based on the scale of token numbers
What do you think? I was assuming creating a hasher for every feature will not be very resource consuming

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
sxd929: fixed! thanks!!

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
tovbinm: let see if we can reuse the hashing function without creating `HashingTF` everytime.

tovbinm: and I think we can. See `HashingTF.transform` and `HashingTF` object

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -142,6 +143,13 @@ case class FeatureDistribution
 private[op] object FeatureDistribution {
 
   val MaxBins = 100000
+  val AvgBinValue = 5000
+  val MaxTokenLowerLimit = 10
+  val getBins = (sum: Summary, bins: Int) => {
```
tovbinm: let's add move this method to Summary class:
```
private[op] case class Summary(min: Double, max: Double) {
   
   // TODO: docs
   def bins(bins: Int): Int = {
       if (sum.max < MaxTokenLowerLimit) bins
       else math.min(math.max(bins, sum.sum / AvgBinValue), MaxBins).toInt
   }

}

Comment (sxd929) [reply core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala]: fixed!

File: core/src/main/scala/com/salesforce/op/filters/Summary.scala [hunk 4]
```diff
@@ -35,18 +35,36 @@ import com.twitter.algebird.Monoid
...
+ * @param count number of doubles for double, number of texts for text
  */
-private[op] case class Summary(min: Double, max: Double)
+private[op] case class Summary(min: Double, max: Double, sum: Double, count: Double) {
+  /**
+   * to calculate the hashing size for RFF (compare js distance) for Text
+   * @param bins default bins of RFF
+   * @return
+   */
+  def getBinsText(bins: Int): Int = bins
```
leahmcguire: The user should be able to pass the formula for how to compute the text feature bin size into RawFeatureFilter as a parameter. We should provide a default which for now can be no formula but later can be updated

tovbinm: Done.

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
```diff
@@ -72,30 +67,47 @@ class FeatureDistributionTest extends FlatSpec with PassengerSparkFixtureTest wi
...
     val values: Array[Map[String, ProcessedSeq]] = Array(
       Map("A" -> Left(Seq("male", "female"))),
       Map("A" -> Right(Seq(1.0)), "B" -> Right(Seq(1.0))),
       Map("B" -> Right(Seq(0.0))))
-    val summary = Array(
-      Map("A" -> Summary(0.0, 1.0), "B" -> Summary(0.0, 5.0)),
-      Map("A" -> Summary(-1.6, 10.6), "B" -> Summary(0.0, 3.0)),
-      Map("B" -> Summary(0.0, 0.0)))
+    val summaries = Array(
+      Map("A" -> Summary(0.0, 2.0, 100.0, 10), "B" -> Summary(0.0, 5.0, 10.0, 10)),
```
Jauntbox: Why did the max change for just the `A` key here?

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
```diff
@@ -107,15 +119,15 @@ class FeatureDistributionTest extends FlatSpec with PassengerSparkFixtureTest wi
       else d.distribution.length shouldBe 2
     }
     distribs(0).nulls shouldBe 0
-    distribs(0).summaryInfo should contain theSameElementsAs Array(0.0, 1.0)
+    distribs(0).summaryInfo should contain theSameElementsAs Array(0.0, 2.0, 100.0, 10.0)
```
Jauntbox: Same here, why did the max change?

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
sxd929: This is hardcoded, I changed it since "A" is "male", "female", so it should be 2 instead of 1

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
sxd929: this one reads the info from hardcoded 'summaries' above

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -179,40 +179,46 @@ private[op] object FeatureDistribution {
...
-          val hist = binned.fold(new Array[Double](bins))(_ + _)
-          splits -> hist
-        } else {
-          val same = seq.map(v => if (v == sum.max) 1.0 else 0.0).sum
-          val other = seq.map(v => if (v != sum.max) 1.0 else 0.0).sum
-          Array(sum.min, sum.max) -> Array(same, other)
+    textBinsFormula: (Summary, Int) => Int
+  ): (Array[Double], Array[Double]) = values match {
+    case Left(seq) =>
+      val numBins = textBinsFormula(summary, bins)
```
leahmcguire: why would the formula need the bins?

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -514,7 +517,8 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
     maxJSDivergence: Double = 0.90,
     maxCorrelation: Double = 0.95,
     correlationType: CorrelationType = CorrelationType.Pearson,
-    protectedFeatures: Array[OPFeature] = Array.empty
+    protectedFeatures: Array[OPFeature] = Array.empty,
+    textBinsFormula: (Summary, Int) => Int = RawFeatureFilter.textBinsFormula
```
leahmcguire: why the int?

tovbinm: it's the bins.

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
tovbinm: Right now we set is as `def textBinsFormula(s: Summary, bins: Int) = bins`.
I can image default `bins` can be returned in some cases.

tovbinm: what would you recommend then? @leahmcguire

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
leahmcguire: Yeah but that is not what we want the formula to look like long term

tovbinm: so what is your suggestion on the inputs / outputs then?

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
sxd929: bins is the default number (e.g. 100) there might be cases that we will just want to fall back to the defaults

PR #103, opened 2018-08-30
Add distributions calculated in RawFeatureFilter to ModelInsights

Jauntbox:
**Related issues**
n/a

Add distributions calculated in RawFeatureFilter to ModelInsights. Note that currently, raw features that aren't explicitly blacklisted by RFF, but are not used because they are inputs to explicitly blacklisted features are not present as raw features in the model, nor in ModelInsights. However, the distributions of such features are still present in the model and accessible via `getRawFeatureDistributions()`.

n/a

**Additional context**
n/a


core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -474,4 +477,22 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
     f0InDer3.variance shouldBe Some(3.3)
   }
 
+  it should "include raw feature distribution information calculated in RawFeatureFilter when it's included" in {
```
tovbinm: I think you can omit `when it's included. Perhaps `include raw feature distribution info calculated in RawFeatureFilter`

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -474,4 +477,22 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
...
+    val wfRawFeatureDistributions = modelWithRFF.getRawFeatureDistributions()
+    val wfDistributionsGrouped = wfRawFeatureDistributions.groupBy(_.name)
+
+    // Currently, raw features that aren't explicitly blacklisted, but are not used because they are inputs to
```
tovbinm: should this go to RawFeatureFilter docs?

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -474,4 +477,22 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
...
+    // explicitly blacklisted features (eg. weight is explicitly blacklisted here, which means that height will
+    // not be added as a raw feature even though it's not explicitly blacklisted itself.
+    val insights = modelWithRFF.modelInsights(predWithMaps)
+    insights.features.foreach(f =>
+      f.distributions should contain theSameElementsAs wfDistributionsGrouped.getOrElse(f.featureName, Array.empty)
+    )
+  }
+
+  it should "not include raw feature distribution information if RawFeatureFilter is not used" in {
```
tovbinm: perhaps `not include raw feature distribution info if RawFeatureFilter is not used`

Jauntbox: I'm missing what's different here... you want "info" instead of "information"?

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
Jauntbox: Sure I can move it there

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
tovbinm: yeah ;) the point is - try to make the test name a bit shorter.

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -311,8 +312,10 @@ case class Discrete(domain: Seq[String], prob: Seq[Double]) extends LabelInfo
  * @param featureName     name of raw feature insights are about
  * @param featureType     type of raw feature insights are about
  * @param derivedFeatures sequence containing insights for each feature derived from the raw feature
+ * @param distributions   distribution information for the raw feature (if calculated in RawFeatureFilter)
  */
-case class FeatureInsights(featureName: String, featureType: String, derivedFeatures: Seq[Insights])
+case class FeatureInsights(featureName: String, featureType: String, derivedFeatures: Seq[Insights],
```
tovbinm: can you perhaps reformat?
```scala
case class FeatureInsights
(
   featureName: String,
   featureType: String,
   derivedFeatures: Seq[Insights],
   distributions: Seq[FeatureDistributionLike] = Seq.empty
)
```

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -311,8 +312,10 @@ case class Discrete(domain: Seq[String], prob: Seq[Double]) extends LabelInfo
...
+  distributions: Seq[FeatureDistributionLike] = Seq.empty)
```
leahmcguire: make type `FeatureDistribution`

tovbinm: @leahmcguire why not the trait?

leahmcguire: it is the class everywhere else

Issue #101, opened 2018-08-29
Create new BinaryClassificationEvaluator that computes relationship between binned probabilities and outcomes


leahmcguire:
**Problem**
We would like a new Evaluator for binary classification which bins the probabilities predicted for the data and compares it to the average outcome of values that fell in that bin. 

**Solution**
This estimator should extend the `OpBinaryClassificationEvaluatorBase` class. It should return a new class extending `EvaluationMetrics` which contains the bins, the counts in each bin, the average outcome of records in each bin and the brier score https://en.wikipedia.org/wiki/Brier_score. An
example of this kind of estimator can be found here: https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala

PR #119, opened 2018-09-07
Adding binaryclassification bin score evaluator

gokulsfdc:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/101

Creating a evaluator for Binary Classification which provides statistics about the predicted scores. This evaluator creates the specified number of bins and computes the following details for each bin.
1. Total number of data points.
2. Average Score.
3. Average Conversion rate.
4. Bin Centers.

Overall BrierScore for the scores is also computed, which is a default metric as well.

Added unit tests for the evaluator


core/src/test/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluatorTest.scala
```diff
@@ -0,0 +1,122 @@
...
+    metrics.AverageConversionRate shouldBe Seq(0.0, 0.0, 0.0, 1.0)
+  }
+
+  it should "return the empty bin metrics for numBins == 0" in {
+    val df = spark.createDataFrame(dataset_test).toDF(predictionLabel, labelName)
+
+    val metrics = new OpBinaryClassifyBinEvaluator(numBins = 0)
+      .setLabelCol(labelName).setPredictionCol(predictionLabel).evaluateAll(df)
+
+    metrics.BrierScore shouldBe 0.0
```
tovbinm: case classes has equals implemented so you can simply do `metrics shouldBe BinaryClassificationBinMetrics(0.0, Seq(), Seq(), Seq(), Seq())` here and everywhere in tests

core/src/test/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluatorTest.scala
```diff
@@ -0,0 +1,122 @@
...
+      .setLabelCol(labelName).setPredictionCol(predictionLabel).evaluateAll(df)
+
+    BigDecimal(metrics.BrierScore).setScale(3, BigDecimal.RoundingMode.HALF_UP).toDouble shouldBe 0.098
+    metrics.BinCenters shouldBe Seq(0.125, 0.375, 0.625, 0.875)
+    metrics.NumberOfDataPoints shouldBe Seq(2, 0, 1, 2)
+    metrics.AverageScore shouldBe Seq(0.003205, 0.0, 0.7, 0.99999)
+    metrics.AverageConversionRate shouldBe Seq(0.0, 0.0, 0.0, 1.0)
+  }
+
+  it should "return the empty bin metrics for numBins == 0" in {
```
tovbinm: `it should "error on invalid num of bins"`

core/src/test/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluatorTest.scala
```diff
@@ -0,0 +1,122 @@
...
+      .setLabelCol(labelName).setPredictionCol(predictionLabel).evaluateAll(df)
+
+    metrics.BrierScore shouldBe 0.0
+    metrics.BinCenters shouldBe Seq.empty[Double]
+    metrics.NumberOfDataPoints shouldBe Seq.empty[Long]
+    metrics.AverageScore shouldBe Seq.empty[Double]
+    metrics.AverageConversionRate shouldBe Seq.empty[Double]
+  }
+
+  it should "return the bin metrics for skewed data" in {
```
tovbinm: `it should "evaluate bin metrics for skewed data"`

core/src/test/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluatorTest.scala
```diff
@@ -0,0 +1,122 @@
...
+      .setLabelCol(labelName).setPredictionCol(predictionLabel).evaluateAll(df)
+
+    metrics.BrierScore shouldBe 7.294225500000013E-4
+    metrics.BinCenters shouldBe Seq(0.1, 0.3, 0.5, 0.7, 0.9)
+    metrics.NumberOfDataPoints shouldBe Seq(0, 0, 0, 0, 4)
+    metrics.AverageScore shouldBe Seq(0.0, 0.0, 0.0, 0.0, 0.98617)
+    metrics.AverageConversionRate shouldBe Seq(0.0, 0.0, 0.0, 0.0, 1.0)
+  }
+
+  it should "return the default metric as BrierScore" in {
```
tovbinm: `it should "evaluate the default metric as BrierScore"`

core/src/test/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluatorTest.scala
```diff
@@ -0,0 +1,122 @@
...
+  val dataset_skewed = Seq(
+    (Map("probability_1" -> 0.99999, "probability_0" -> 0.0001, "prediction" -> 1.0), 1.0),
+    (Map("probability_1" -> 0.99999, "probability_0" -> 0.0001, "prediction" -> 1.0), 1.0),
+    (Map("probability_1" -> 0.9987, "probability_0" -> 0.001, "prediction" -> 1.0), 1.0),
+    (Map("probability_1" -> 0.946, "probability_0" -> 0.0541, "prediction" -> 1.0), 1.0)
+  )
+
+  val emptyDataSet = Seq.empty[(Map[String, Double], Double)]
+
+  Spec[OpBinaryClassifyBinEvaluator] should "return the bin metrics" in {
```
tovbinm: `should "evaluate the bin metrics"`

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluator.scala
```diff
@@ -0,0 +1,170 @@
...
+ * Total number of data points per bin
+ * Average Score per bin
+ * Average Conversion rate per bin
+ * Bin Centers for each bin
+ * BrierScore for the overall dataset is also computed, which is a default metric as well.
+ *
+ * @param name            name of default metric
+ * @param isLargerBetter  is metric better if larger
+ * @param uid             uid for instance
+ */
```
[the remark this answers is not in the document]
gokulsfdc: This evaluator returns 5 diff values. Should there be 5 factory methods? or only for Brier Score, which is the defaultmetric?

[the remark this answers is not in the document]
tovbinm: only the Brier Score. @leahmcguire wdyt?

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluator.scala
```diff
@@ -0,0 +1,164 @@
...
+  @transient private lazy val log = LoggerFactory.getLogger(this.getClass)
+
+  def getDefaultMetric: BinaryClassificationBinMetrics => Double = _.BrierScore
+
+  override def evaluateAll(data: Dataset[_]): BinaryClassificationBinMetrics = {
+    val labelColumnName = getLabelCol
+    val dataProcessed = makeDataToUse(data, labelColumnName)
+
+    import dataProcessed.sparkSession.implicits._
+    val rdd = dataProcessed.select(getPredictionValueCol, labelColumnName).as[(Double, Double)].rdd
```
leahmcguire: you dont seem to use this rdd beyond checking if the dataset is empty

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluator.scala
```diff
@@ -0,0 +1,164 @@
...
+          case Row(prob: Double, label: Double) => (prob, label)
+        }
+
+      if (numBins == 0) {
+        log.error("numBins is set to 0. Returning empty metrics")
+        BinaryClassificationBinMetrics(0.0, Seq.empty[Double], Seq.empty[Long], Seq.empty[Double], Seq.empty[Double])
+      } else {
+        // Find the significant digit to which the scores needs to be rounded, based of numBins.
+        val significantDigitToRoundOff = math.log10(numBins).toInt + 1
+        val scoreAndLabelsRounded = for {i <- scoreAndLabels}
```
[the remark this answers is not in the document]
leahmcguire: I think my explanation on the doc may have made this more complicated than it needs to be. 
The probabilities will be between 0 and 1 so use that information to compute the bins and their centers. Make a function `binFn` that takes in the probability and returns the bin. Then you can simply do:
`scoreAndLabels.map{ case (score, label) => (binFun(score), (score, label, 1L) }.reduceByKey(_ + _)
.map{ case (bin, (scoreSum, labelSum, count)) => (bin, scoreSum/count, labelSum/count, count)  }`
(you will need to `import com.twitter.algebird.Operators._` for the reduceByKey to work without specifying everything)

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassifyBinEvaluator.scala
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * @param BinCenters            center of each bin
+ * @param NumberOfDataPoints    total number of data points in each bin
+ * @param AverageScore          average score in each bin
+ * @param AverageConversionRate average conversion rate in each bin
+ * @param BrierScore            brier score for overall dataset
+ */
+case class BinaryClassificationBinMetrics
+(
+  BrierScore: Double,
```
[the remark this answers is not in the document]
leahmcguire: So for some reason all of our metrics do not follow this convention @tovbinm. We should figure out what we want and then make them all consistent.

tovbinm:
please take a look into the failing test
```
[-] BinaryClassificationModelSelector should fit and predict for default models (497.062 secs)
com.salesforce.op.stages.impl.classification.BinaryClassificationModelSelectorTest > BinaryClassificationModelSelector should fit and predict for default models FAILED
    java.lang.ArrayIndexOutOfBoundsException: -175
        at com.salesforce.op.evaluators.OpBinScoreEvaluator$$anonfun$7.apply(OpBinScoreEvaluator.scala:100)
        at com.salesforce.op.evaluators.OpBinScoreEvaluator$$anonfun$7.apply(OpBinScoreEvaluator.scala:96)
        at scala.collection.IndexedSeqOptimized$class.foldl(IndexedSeqOptimized.scala:57)
        at scala.collection.IndexedSeqOptimized$class.foldLeft(IndexedSeqOptimized.scala:66)
        at scala.collection.mutable.ArrayOps$ofRef.foldLeft(ArrayOps.scala:186)
        at com.salesforce.op.evaluators.OpBinScoreEvaluator.evaluateAll(OpBinScoreEvaluator.scala:96)
        at com.salesforce.op.evaluators.OpBinScoreEvaluator.evaluateAll(OpBinScoreEvaluator.scala:59)
        at com.salesforce.op.stages.impl.selector.HasEval$$anonfun$1.apply(ModelSelectorNames.scala:92)
        at com.salesforce.op.stages.impl.selector.HasEval$$anonfun$1.apply(ModelSelectorNames.scala:89)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
        at scala.collection.immutable.List.foreach(List.scala:392)
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
        at scala.collection.immutable.List.map(List.scala:296)
        at com.salesforce.op.stages.impl.selector.HasEval$class.evaluate(ModelSelectorNames.scala:89)
        at com.salesforce.op.stages.impl.selector.ModelSelector.evaluate(ModelSelector.scala:74)
        at com.salesforce.op.stages.impl.selector.ModelSelector.fit(ModelSelector.scala:171)
        at com.salesforce.op.stages.impl.classification.BinaryClassificationModelSelectorTest$$anonfun$8.apply(BinaryClassificationModelSelectorTest.scala:186)
        at com.salesforce.op.stages.impl.classification.BinaryClassificationModelSelectorTest$$anonfun$8.apply(BinaryClassificationModelSelectorTest.scala:171)
```
https://travis-ci.com/salesforce/TransmogrifAI/jobs/145162156
I guess more unit tests needed?

leahmcguire:
Yeah so what is happening is that not every model type produces a probablity column. And there is a special case in the makeData call that will fall back to the rawPrediciton value instead of the probability. This is not bounded so you will actually need to do a reduce to get the min and max values for the bin range.

gokulsfdc:
So do you mean, default min max value to compute bin should be [0,1]
If probability goes out of bounds due to the fallback , then this range needs to be altered accordingly

PR #120, opened 2018-09-07
Specify categorical variables in metadata

michaelweilsalesforce:
**Related issues**

**Additional context**

### DIFF
--- a/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
@@ -32,18 +32,16 @@ package com.salesforce.op.stages.impl.feature
 
 import org.apache.spark.ml.attribute.AttributeGroup
 import org.apache.spark.sql.types.StructField
-import org.junit.runner.RunWith
-import org.scalatest.Matchers
-import org.scalatest.junit.JUnitRunner
+import org.scalatest.{Assertion, Matchers}
 
-object AttributeTestUtils extends Matchers{
-
+trait AttributeAsserts {
+  self: Matchers =>
   /**
    * Assert if attributes are nominal or not
    * @param schema
    * @param expectedNominal Expected array of booleans. True if the field is nominal, false if not.
    */
-  final def assertNominal(schema: StructField, expectedNominal: Array[Boolean]): Unit = {
+  final def assertNominal(schema: StructField, expectedNominal: Array[Boolean]): Assertion = {
     val attributes = AttributeGroup.fromStructField(schema).attributes.get
     attributes.map(_.isNominal) shouldBe expectedNominal
   }
--- a/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
@@ -44,6 +44,6 @@ trait AttributeAsserts {
    */
   final def assertNominal(schema: StructField, expectedNominal: Array[Boolean]): Assertion = {
     val attributes = AttributeGroup.fromStructField(schema).attributes
-    attributes.map(_.map(_.isNominal)) shouldBe Some(expectedNominal)
+    attributes.map(_.map(_.isNominal).toSeq) shouldBe Some(expectedNominal.toSeq)
   }
 }
--- a/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
@@ -38,11 +38,12 @@ trait AttributeAsserts {
   self: Matchers =>
   /**
    * Assert if attributes are nominal or not
+   *
    * @param schema
    * @param expectedNominal Expected array of booleans. True if the field is nominal, false if not.
    */
   final def assertNominal(schema: StructField, expectedNominal: Array[Boolean]): Assertion = {
-    val attributes = AttributeGroup.fromStructField(schema).attributes.get
-    attributes.map(_.isNominal) shouldBe expectedNominal
+    val attributes = AttributeGroup.fromStructField(schema).attributes
+    attributes.map(_.map(_.isNominal)) shouldBe Some(expectedNominal)
   }
 }
--- a/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeTestUtils.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeTestUtils.scala
@@ -1,3 +1,33 @@
+/*
+ * Copyright (c) 2017, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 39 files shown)


core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeTestUtils.scala
```diff
@@ -0,0 +1,50 @@
...
+
+package com.salesforce.op.stages.impl.feature
+
+import org.apache.spark.ml.attribute.AttributeGroup
+import org.apache.spark.sql.types.StructField
+import org.junit.runner.RunWith
+import org.scalatest.Matchers
+import org.scalatest.junit.JUnitRunner
+
+object AttributeTestUtils extends Matchers{
```
tovbinm: 1. better use a trait - it will be cleaner and simpler to use.
2. make return type the `Assertion`
here it is:
```scala
trait AttributeAsserts {
   self: Matchers =>
   final def assertNominal(schema: StructField, expectedNominal: Array[Boolean]): Assertion = ???
}
```
then mixin it like this:
```
class BinaryVectorizerTest extends OpTransformerSpec[OPVector, BinaryVectorizer] with AttributeAsserts { ... }
```

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
```diff
@@ -72,6 +75,8 @@ class OpVectorMetadata private
     newColumns: Array[OpVectorColumnMetadata]
   ): OpVectorMetadata = OpVectorMetadata(name, newColumns, history)
 
+  val textTypes = Seq(MultiPickList, MultiPickListMap, Text, TextArea, TextAreaMap, TextMap, Binary, BinaryMap,
```
tovbinm: 1. how is `Binary` and `BinaryMap` are testTypes?
2. also instead please use `FeatureType.shortTypeName[Text]` etc

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
```diff
@@ -31,15 +31,18 @@
 package com.salesforce.op.utils.spark
 
 import com.salesforce.op.FeatureHistory
+import com.salesforce.op.features.types.{Binary, BinaryMap, MultiPickList, MultiPickListMap, Text, TextArea, TextAreaMap, TextList, TextMap}
```
tovbinm: `import com.salesforce.op.features.types._`

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
michaelweilsalesforce: I want the package name as well. E.g `com.salesforce.features.types. MultiPickList`

As it is specified in `OpVectorColumnMetadata.parentFeatureType`

tovbinm: Then FeatureType.typeName

core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
```diff
@@ -0,0 +1,48 @@
...
+
+trait AttributeAsserts {
+  self: Matchers =>
+  /**
+   * Assert if attributes are nominal or not
+   * @param schema
+   * @param expectedNominal Expected array of booleans. True if the field is nominal, false if not.
+   */
+  final def assertNominal(schema: StructField, expectedNominal: Array[Boolean]): Assertion = {
+    val attributes = AttributeGroup.fromStructField(schema).attributes.get
```
tovbinm: avoid `.get`, rather do `attributes.map(_.map(_.isNominal)) shouldBe Some(expectedNominal)`

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
```diff
@@ -72,6 +75,10 @@ class OpVectorMetadata private
     newColumns: Array[OpVectorColumnMetadata]
   ): OpVectorMetadata = OpVectorMetadata(name, newColumns, history)
 
+  val categoricalTypes = Seq(FeatureType.typeName[MultiPickList], FeatureType.typeName[MultiPickListMap],
+    FeatureType.typeName[Text], FeatureType.typeName[TextArea], FeatureType.typeName[TextAreaMap],
+    FeatureType.typeName[TextMap], FeatureType.typeName[Binary], FeatureType.typeName[BinaryMap],
+    FeatureType.typeName[TextList])
```
leahmcguire: picklist? Combo box? country, state, city, id

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
```diff
@@ -82,12 +89,22 @@ class OpVectorMetadata private
     val groupedCol = columns
       .groupBy(c => (c.parentFeatureName, c.parentFeatureType, c.grouping, c.indicatorValue, c.descriptorValue))
     val colData = groupedCol.toSeq
-      .map{ case (_, g) => g.head -> g.map(_.index) }
-    val colMeta = colData.map{ case (c, i) => c.toMetadata(i) }
-    new MetadataBuilder()
+      .map { case (_, g) => g.head -> g.map(_.index) }
+    val colMeta = colData.map { case (c, i) =>
+      c.toMetadata(i)
+    }
```
leahmcguire: nit unnecessary new line hanging => is bad

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
```diff
@@ -82,12 +89,22 @@ class OpVectorMetadata private
...
+    val meta = new MetadataBuilder()
       .putMetadataArray(OpVectorMetadata.ColumnsKey, colMeta.toArray)
       .putMetadata(OpVectorMetadata.HistoryKey, FeatureHistory.toMetadata(history))
       .build()
+    val attributes = columns.map { c =>
```
tovbinm:
```scala
val attributes = columns.map {
    case c if c.indicatorValue.isDefined || categoricalTypes.exists(c.parentFeatureType.contains) =>
        BinaryAttribute.defaultAttr.withName(c.makeColName()).withIndex(c.index)
    case c =>
        NumericAttribute.defaultAttr.withName(c.makeColName()).withIndex(c.index)
}
```

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
michaelweilsalesforce: Oh yeah If it is only hashing + count, let's remove all these Text types. Do we only do hashing to Combo box, country, state, city, id?

leahmcguire: we do pivot - so they should be picked up automatically. I think we also do pivot on multiPickList. So you may want to remove the categorical types check completely and only rely on the indicatorValue

tovbinm:
@michaelweilsalesforce please update the description of the PR to clarify the reasons and the solution.

PR #122, opened 2018-09-10
Spark packages instructions

tovbinm:


tovbinm:
@mayukhb @leahmcguire should we also add this paragraph somewhere? https://github.com/databricks/spark-avro#with-spark-shell-or-spark-submit

PR #125, opened 2018-09-11
ModelSelector UID Fix

leahmcguire:
**Related issues**
Because objects are initialized on first use having the models to try defined as values in the Selector interfaces means that a different number of stages are initialized if the Selectors are used twice in the same JVM. This means that the selectors cannot be used in unit tests that save and then load the model by creating a new workflow.

Changed the values to defs for default models so the same number of stages are created every time the model selector is called

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
@@ -50,7 +50,7 @@ case object BinaryClassificationModelSelector {
     MTT.OpRandomForestClassifier, MTT.OpGBTClassifier, MTT.OpLinearSVC)
   // OpNaiveBayes and OpDecisionTreeClassifier off by default
 
-  private val defaultModelsAndParams: Seq[(EstimatorType, Array[ParamMap])] = {
+  private def defaultModelsAndParams: Seq[(EstimatorType, Array[ParamMap])] = {
     val lr = new OpLogisticRegression()
     val lrParams = new ParamGridBuilder()
       .addGrid(lr.fitIntercept, DefaultSelectorParams.FitIntercept)
@@ -146,7 +146,7 @@ case object BinaryClassificationModelSelector {
     stratify: Boolean = ValidatorParamDefaults.Stratify,
     parallelism: Int = ValidatorParamDefaults.Parallelism,
     modelTypesToUse: Seq[BinaryClassificationModelsToTry] = modelNames,
-    modelsAndParameters: Seq[(EstimatorType, Array[ParamMap])] = defaultModelsAndParams
+    modelsAndParameters: Seq[(EstimatorType, Array[ParamMap])] = Seq.empty
   ): ModelSelector[ModelType, EstimatorType] = {
     val cv = new OpCrossValidation[ModelType, EstimatorType](
       numFolds = numFolds, seed = seed, validationMetric, stratify = stratify, parallelism = parallelism
@@ -187,7 +187,7 @@ case object BinaryClassificationModelSelector {
     stratify: Boolean = ValidatorParamDefaults.Stratify,
     parallelism: Int = ValidatorParamDefaults.Parallelism,
     modelTypesToUse: Seq[BinaryClassificationModelsToTry] = modelNames,
-    modelsAndParameters: Seq[(EstimatorType, Array[ParamMap])] = defaultModelsAndParams
+    modelsAndParameters: Seq[(EstimatorType, Array[ParamMap])] = Seq.empty
   ): ModelSelector[ModelType, EstimatorType] = {
     val ts = new OpTrainValidationSplit[ModelType, EstimatorType](
       trainRatio = trainRatio, seed = seed, validationMetric, stratify = stratify, parallelism = parallelism
@@ -206,7 +206,9 @@ case object BinaryClassificationModelSelector {
   ): ModelSelector[ModelType, EstimatorType] = {
     val modelStrings = modelTypesToUse.map(_.entryName)
     val modelsToUse =
-      if (modelsAndParameters == defaultModelsAndParams || modelTypesToUse != modelNames) modelsAndParameters
+      if (modelsAndParameters.isEmpty) defaultModelsAndParams
+        .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      else if (modelTypesToUse != modelNames) modelsAndParameters
         .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
       else modelsAndParameters
     new ModelSelector(
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
@@ -206,10 +206,13 @@ case object BinaryClassificationModelSelector {
   ): ModelSelector[ModelType, EstimatorType] = {
     val modelStrings = modelTypesToUse.map(_.entryName)
     val modelsToUse =
+      // if no models are specified use the defaults and filter then by the named models to use
       if (modelsAndParameters.isEmpty) defaultModelsAndParams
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -206,7 +206,9 @@ case object BinaryClassificationModelSelector {
   ): ModelSelector[ModelType, EstimatorType] = {
     val modelStrings = modelTypesToUse.map(_.entryName)
     val modelsToUse =
-      if (modelsAndParameters == defaultModelsAndParams || modelTypesToUse != modelNames) modelsAndParameters
+      if (modelsAndParameters.isEmpty) defaultModelsAndParams
+        .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      else if (modelTypesToUse != modelNames) modelsAndParameters
```
tovbinm: if user only passes `modelTypesToUse = Seq(OpLogisticRegression)` the model selector wont work, since `modelsAndParameters` will be initialized as `Seq.empty`

leahmcguire: not true - if the user only passes modelTypesToUse = Seq(OpLogisticRegression) the first condition will be hit

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -50,7 +50,7 @@ case object BinaryClassificationModelSelector {
     MTT.OpRandomForestClassifier, MTT.OpGBTClassifier, MTT.OpLinearSVC)
   // OpNaiveBayes and OpDecisionTreeClassifier off by default
 
-  private val defaultModelsAndParams: Seq[(EstimatorType, Array[ParamMap])] = {
+  private def defaultModelsAndParams: Seq[(EstimatorType, Array[ParamMap])] = {
```
vpatryshev: Does it produce a mutable sequence? Is not it supposed to be a constant?

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -206,8 +206,13 @@ case object BinaryClassificationModelSelector {
...
     val modelStrings = modelTypesToUse.map(_.entryName)
     val modelsToUse =
-      if (modelsAndParameters == defaultModelsAndParams || modelTypesToUse != modelNames) modelsAndParameters
+      // if no models are specified use the defaults and filter by the named models to use
+      if (modelsAndParameters.isEmpty) defaultModelsAndParams
         .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      // if models to use has been specified and the models have been specified filter the models by the names
+      else if (modelTypesToUse != modelNames) modelsAndParameters
+        .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      // else just use the specified models
```
vpatryshev: The presence of comments kind of demonstrates that the code is too quirky.
See what you can do.
First, don't use `modelStrings`. It's misleading. Instead, there must be a method, in `BinaryClassificationModelsToTry`, or, rather, in its superclass (or, better, in a trait) that checks whether it's compatible with `modelTypes. So you can use that method, and write, instead of `modelStrings.contains(e.getClass.getSimpleName)`, `e.isCompatibleWith(modelTypesToUse)`. 
Second, comparing two sequences is weird. If we are dealing with sets, it's one thing; if we are dealing with lists, it's another. Do you mean here a set or a list? What's the reason of this comparison? If these collections are being passed around, it makes more sense to introduce a wrapper class for `Seq[(EstimatorType, Array[ParamMap])]`, call it `ModelsAndParameters`, and throw in functionality into it.
And have a good test coverage. A sequence of `if`s makes the code hard to perceive, and hard to prove that it is actually correct.

core/src/main/scala/com/salesforce/op/stages/impl/classification/MultiClassificationModelSelector.scala
```diff
@@ -49,7 +49,7 @@ case object MultiClassificationModelSelector {
   private[op] val modelNames: Seq[MultiClassClassificationModelsToTry] = Seq(MTT.OpLogisticRegression,
     MTT.OpRandomForestClassifier) // OpDecisionTreeClassifier and OpNaiveBayes off by default
 
-  private val defaultModelsAndParams: Seq[(EstimatorType, Array[ParamMap])] = {
+  private def defaultModelsAndParams: Seq[(EstimatorType, Array[ParamMap])] = {
     val lr = new OpLogisticRegression()
```
vpatryshev: Is it a duplicate of the code above?
Why not have a trait holding these things?

core/src/main/scala/com/salesforce/op/stages/impl/classification/MultiClassificationModelSelector.scala
```diff
@@ -184,8 +184,13 @@ case object MultiClassificationModelSelector {
...
     val modelsToUse =
-      if (modelsAndParameters == defaultModelsAndParams || modelTypesToUse != modelNames) modelsAndParameters
+      // if no models are specified use the defaults and filter by the named models to use
+      if (modelsAndParameters.isEmpty) defaultModelsAndParams
         .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      // if models to use has been specified and the models have been specified filter the models by the names
+      else if (modelTypesToUse != modelNames) modelsAndParameters
+        .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      // else just use the specified models
       else modelsAndParameters
```
vpatryshev: Is it a total duplicate? Please, please, extract this common functionality to a trait or something.

core/src/main/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelector.scala
```diff
@@ -195,8 +196,13 @@ case object RegressionModelSelector {
...
     val modelStrings = modelTypesToUse.map(_.entryName)
     val modelsToUse =
-      if (modelsAndParameters == defaultModelsAndParams || modelTypesToUse != modelNames) modelsAndParameters
+    // if no models are specified use the defaults and filter by the named models to use
+      if (modelsAndParameters.isEmpty) defaultModelsAndParams
         .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      // if models to use has been specified and the models have been specified filter the models by the names
+      else if (modelTypesToUse != modelNames) modelsAndParameters
+        .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+      // else just use the specified models
```
vpatryshev: Oh, the third time the same code. Should be fixed.
Also, it needs tests.

core/src/main/scala/com/salesforce/op/stages/impl/classification/MultiClassificationModelSelector.scala
leahmcguire: this is not a duplicat they are different levels of control on the models used

core/src/main/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelector.scala
leahmcguire: this has tests in the interface for all three constructor objects

leahmcguire:
@vpatryshev I extracted the common private constructor into a trait - the rest of the functions, while at first glance look like duplicates are different because we want the type checks on the user interface.

leahmcguire:
tests for these classes are:
https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelectorTest.scala
https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/stages/impl/classification/MultiClassificationModelSelectorTest.scala
https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala

PR #127, opened 2018-09-12
added check for multipicklist other class as not being binary in metadata

leahmcguire:
**Related issues**
creation of spark attribute groups was mislabeling multipicklist other column as binary. this made tree models fail

multipicklist other is now a numeric not binary


features/src/main/scala/com/salesforce/op/utils/spark/OpVectorColumnMetadata.scala
```diff
@@ -155,6 +161,7 @@ object OpVectorColumnMetadata {
   val DescriptorValueKey = "descriptor_value"
   val IndicesKey = "indices"
   val NullString = "NullIndicatorValue"
+  val OtherString = "OTHER"
```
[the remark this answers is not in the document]
leahmcguire: I changed the TransmogrifierDefaults to point to this

michaelweilsalesforce:
@leahmcguire should we add in `OpRandomForestTest` a unit test on a data that contains almost all possible feature transformation we have?
I can help contribute to it...

core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
```diff
@@ -42,8 +43,11 @@ trait AttributeAsserts {
    * @param schema
    * @param expectedNominal Expected array of booleans. True if the field is nominal, false if not.
    */
-  final def assertNominal(schema: StructField, expectedNominal: Array[Boolean]): Assertion = {
+  final def assertNominal(schema: StructField, expectedNominal: Array[Boolean], output: Array[OPVector]): Assertion = {
```
tovbinm: please update docs for the `assertNominal` function

core/src/test/scala/com/salesforce/op/stages/impl/feature/AttributeAsserts.scala
```diff
@@ -42,8 +43,11 @@ trait AttributeAsserts {
...
     val attributes = AttributeGroup.fromStructField(schema).attributes
+    output.foreach{
+      x => x.value.toArray.zip(expectedNominal).foreach{ case (v , n) => if (n) (v == 0.0 || v == 1.0) shouldBe true }
```
tovbinm: should we also add a check when `n == false`? is so, then better:
```scala
for {
    x <- output
    (value, nominal) <- x.value.toArray.zip(expectedNominal)
} if (nominal) (v == 0.0 || v == 1.0) shouldBe true else ???
```

leahmcguire: there are no constraints on if n is false

tovbinm: so it can be NaN, -/+Inf as well?

PR #128, opened 2018-09-12
adding feature importance for new model types

leahmcguire:


leahmcguire:
it would not be simple - they would be pretty heavy duty (see long running) tests

tovbinm:
I mean we can unit test just the `getModelContributions` and other model isights functions without actually training the models but just instantiating them, e.g:
```getModelContributions(Some(new GBTClassificationModel(...))```

PR #133, opened 2018-09-15
Cleaning up the OpBinScoreEvaluator class a little bit

tovbinm:
Removed some redundant ctor arguments and renames & idents.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/evaluators/Evaluators.scala
+++ b/core/src/main/scala/com/salesforce/op/evaluators/Evaluators.scala
@@ -54,11 +54,8 @@ object Evaluators {
 
     /*
      * Brier Score for the prediction
-     *
-     * @param numOfBins number of bins to produce
      */
-    def brierScore(numOfBins: Int = 100): OpBinScoreEvaluator =
-      new OpBinScoreEvaluator(numOfBins = numOfBins)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/salesforce/op/evaluators/Evaluators.scala
```diff
@@ -54,10 +54,11 @@ object Evaluators {
 
     /*
      * Brier Score for the prediction
+     *
+     * @param numOfBins number of bins to produce
      */
-    def brierScore(): OpBinScoreEvaluator =
-      new OpBinScoreEvaluator(
-        name = BinaryClassEvalMetrics.brierScore, isLargerBetter = true)
+    def brierScore(numOfBins: Int = 100): OpBinScoreEvaluator =
```
gokulsfdc: brierScore is independent of numBins.

tovbinm: Then perhaps we should not even expose it. Why is it 100 by default?

gokulsfdc: numBins are needed to compute other stats in OpBinScoreEvaluator like averageScore per bin.
As we are interested only in brierScore here, we don't need to pass numBins.

tovbinm: got it. removed.

Issue #115, opened 2018-09-06
No feature type tag mapping for Spark type IntegerType


liuzhenhai93:
**Describe the bug**
spark sql type integer is not mapped to featuretype
in the source code,type is mapped as bellow.
def featureTypeTagOf(sparkType: DataType, isNullable: Boolean): WeakTypeTag[_ <: FeatureType] = sparkType match {
    case DoubleType if !isNullable => weakTypeTag[types.RealNN]
    case DoubleType => weakTypeTag[types.Real]
 // **add this line  and tweak method fromSparkFn in FeatureTypeSparkConverter.scala can fix the bug**
  // case IntegerType => weakTypeTag[types.Integral]

    case LongType => weakTypeTag[types.Integral]
    case ArrayType(StringType, _) => weakTypeTag[types.TextList]
    case StringType => weakTypeTag[types.Text]
    case BooleanType => weakTypeTag[types.Binary]
    ......
    case _ => throw new IllegalArgumentException(s"No feature type tag mapping for Spark type $sparkType")
  }
 **and change  line 153 ( in method fromSparkFn)  in FeatureTypeSparkConverter.scala to:** 
 case wt if wt <:< weakTypeOf[t.Integral] => (value: Any) => {
        var result: Any = FeatureTypeDefaults.Integral.value
        if (value == null) {result = FeatureTypeDefaults.Integral.value}
        else if (value.isInstanceOf[Integer]) {result = Some(value.asInstanceOf[Int].toLong)}
        else {result = Some(value.asInstanceOf[Long])}
        result
        }

**To Reproduce**
package com.jd.spark
import com.salesforce.op._
import com.salesforce.op.readers._
import com.salesforce.op.features._
import com.salesforce.op.features.types._
import com.salesforce.op.stages.impl.classification._
import org.apache.spark.SparkConf
import org.apache.spark.sql.SparkSession
import org.apache.spark.sql.functions.udf
case class Passenger(
  id: Int,
  survived: Int,
  pClass: Option[Int],
  name: Option[String],
  sex: Option[String],
  age: Option[Double],
  sibSp: Option[Int],
  parCh: Option[Int],
  ticket: Option[String],
  fare: Option[Double],
  cabin: Option[String],
  embarked: Option[String])
object TestSalesForce {
  def main(args: Array[String]) = {
    implicit val spark = SparkSession.builder.config(new SparkConf().setMaster("local")).enableHiveSupport()
      .getOrCreate()
[...]

PR #135, opened 2018-09-19
Added support for DateType and TimestampType primitive spark types

ajayborra:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/115

- Added conversions for DateType and TimestampType spark types for completeness.
- Better error messages for non supported spark types.


features/src/main/scala/com/salesforce/op/features/types/FeatureTypeSparkConverter.scala
```diff
@@ -147,14 +147,36 @@ case object FeatureTypeSparkConverter {
...
+          case null => FeatureTypeDefaults.Integral.value
+          case _: Short => Some(value.asInstanceOf[Short].toLong)
+          case _: Int => Some(value.asInstanceOf[Int].toLong)
+          case _: Long => Some(value.asInstanceOf[Long])
+          case _ => throw new IllegalArgumentException(s"Integral type mapping is not defined for ${value.getClass}")
+        }
       case wt if wt <:< weakTypeOf[t.Binary] => (value: Any) =>
         if (value == null) FeatureTypeDefaults.Binary.value else Some(value.asInstanceOf[Boolean])
 
+      // Date & Time
```
tovbinm: Why do you need special handling for `Date` and `DateTime`? were you planning to take care of `TimestampType` spark type?

features/src/main/scala/com/salesforce/op/features/types/FeatureTypeSparkConverter.scala
```diff
@@ -147,14 +147,36 @@ case object FeatureTypeSparkConverter {
 
       // Numerics
       case wt if wt <:< weakTypeOf[t.RealNN] => (value: Any) =>
-        if (value == null) None else Some(value.asInstanceOf[Double])
+        value match {
+          case null => None
+          case _: Float => Some(value.asInstanceOf[Float].toDouble)
```
tovbinm: `case v: Float => Some(v.toDouble)` - and the same everywhere.

features/src/main/scala/com/salesforce/op/features/types/FeatureTypeSparkConverter.scala
tovbinm: Since Date and DateTime types in TransmogrifAI inherit from Integral types, their values are captured by this case `case wt if wt <:< weakTypeOf[t.Integral]` and it works, because `<:<` operator on type tags check inheritance.

tovbinm: If you would like to mape spark types `DateType` and `TimestampType` into TransmogrifAI `Date` and `DateTime` types accordingly then you should move the cases `case wt if wt <:< weakTypeOf[t.Date]` and `case wt if wt <:< weakTypeOf[t.DateTime]` before the numeric checks.

tovbinm:
so I think we can start by adding support for `DateType` and `TimestampType` and as for int, short etc let's discuss it.

ajayborra:
> In any case, the main problem with this approach can result in loosing precision. If spark type is say float, we convert it to double, then user applies the operation on doubles, but then we store it back into Dataframe as float loosing precision. Example:
> 
> ```scala
> scala> val f = Float.MaxValue
> f: Float = 3.4028235E38
> scala> val d = f.oDouble + f.toDouble
> d: Double = 6.805646932770577E38
> scala> val f2 = d.toFloat 
> f2: Float = Infinity
> ```
> 
> same applies to short, int etc.

It is true that dropping of the precision can be a problem introducing vanishing gradients etc..., In your opinion what is the best approach to handle this types?
There are few approaches that we can take here,
- Auto convert to largest range type and warn the user about the precision problems involved?
- Maybe provide a way to turn on/off auto conversions as a parameter?
- Avoid auto convert and recommend user to choose the type with the largest range in the category by throwing an error when we receive lower range? (Eg. Recommend Long when we see a Short )

tovbinm:
I think for a start we can add nicer error message for these types in `FeatureSparkTypes.featureTypeTagOf`. Something like this:
```
case IntegerType => throw new IllegalArgumentException("Spark IntegerType is currently not supported. Please use LongType instead.")
    
```

tovbinm:
I would rather have user explicitly make a conversion from int -> long, float -> double rather than taking care of precision issues.

ajayborra:
Good point. Will update the PR with changes.

tovbinm:
@ajayborra please add test cases for these changes in `FeatureTypeSparkConverterTest`.

tovbinm:
@ajayborra are you planning to be working on this one or? ;)

features/src/main/scala/com/salesforce/op/features/types/FeatureTypeSparkConverter.scala
```diff
@@ -145,13 +145,41 @@ case object FeatureTypeSparkConverter {
...
+        if (value == null) FeatureTypeDefaults.DateTime.value else Some(value.asInstanceOf[Long])
+      case wt if wt <:< weakTypeOf[t.Date] => (value: Any) =>
+        value match {
+          case null => FeatureTypeDefaults.Date.value
+          case v: Int => Some(v.toLong)
+          case v: Long => Some(v)
+          case _ => throw new IllegalArgumentException(s"Date type mapping is not defined for ${value.getClass}")
+        }
+
+      // Numerals
```
tovbinm: ```// Numerics```

features/src/test/scala/com/salesforce/op/features/FeatureSparkTypeTest.scala
```diff
@@ -0,0 +1,99 @@
...
+    (MapType(StringType, DoubleType, valueContainsNull = true), weakTypeTag[types.RealMap]),
+    (MapType(StringType, LongType, valueContainsNull = true), weakTypeTag[types.IntegralMap]),
+    (MapType(StringType, BooleanType, valueContainsNull = true), weakTypeTag[types.BinaryMap]),
+    (MapType(StringType, ArrayType(StringType, containsNull = true), valueContainsNull = true),
+      weakTypeTag[types.MultiPickListMap]),
+    (MapType(StringType, ArrayType(DoubleType, containsNull = true), valueContainsNull = true),
+      weakTypeTag[types.GeolocationMap])
+  )
+
+  val sparkNonNullableTypeToTypeTagMappings = Seq(
```
tovbinm: RealNN is non nullable, why is it Real here?

features/src/test/scala/com/salesforce/op/features/FeatureSparkTypeTest.scala
```diff
@@ -0,0 +1,99 @@
...
+    (DoubleType, weakTypeTag[types.Real]), (FloatType, weakTypeTag[types.Real])
+  )
+
+  Spec(FeatureSparkTypes.getClass) should "assign appropriate feature type tags for valid types" in {
+    sparkTypeToTypeTagMappings.foreach(mapping =>
```
tovbinm: avoid _1 and _2, better name args:
```scala
for { (sparkType, featureTypeTag) <- sparkTypeToTypeTagMappings } {
    FeatureSparkTypes.featureTypeTagOf(sparkType, isNullable = false) shouldBe featureTypeTag
}

File: features/src/test/scala/com/salesforce/op/features/types/FeatureTypeSparkConverterTest.scala [hunk 6]
```diff
@@ -52,6 +52,12 @@ class FeatureTypeSparkConverterTest
   )
   val bogusNames = Gen.alphaNumStr
 
+  val naturalNumbers = Table("NaturalNumbers", Byte.MaxValue, Short.MaxValue, Int.MaxValue, Long.MaxValue)
```
tovbinm: Are you sure that this generates numbers in these ranges correctly? what about negatives and special values?

features/src/test/scala/com/salesforce/op/features/types/FeatureTypeSparkConverterTest.scala
```diff
@@ -94,4 +100,100 @@ class FeatureTypeSparkConverterTest
...
+  property("converts timestamp denoted using long ranges to long range datetime feature") {
+    forAll(dateTimeValues)(dt =>
+      FeatureTypeSparkConverter[DateTime]().fromSpark(dt).value.get shouldBe a[java.lang.Long]
+    )
+  }
+
+  property("converts string to text feature type") {
+    val text = FeatureTypeSparkConverter[Text]().fromSpark("Simple")
+    text shouldBe a[Text]
+    text.value.get shouldBe a[String]
```
tovbinm: `text.value shouldBe Some("Simple")`

features/src/test/scala/com/salesforce/op/features/types/FeatureTypeSparkConverterTest.scala
```diff
@@ -94,4 +100,100 @@ class FeatureTypeSparkConverterTest
...
+  }
+
+  property("converts a Boolean to Binary feature type") {
+    val bool = FeatureTypeSparkConverter[Binary]().fromSpark(false)
+    bool shouldBe a[Binary]
+    bool.value.get shouldBe a[java.lang.Boolean]
```
tovbinm: `bool.value shouldBe Some(false)`

Issue #136, opened 2018-09-21
Titanic evaluate throws an error on 0.3.4 and master branches


tovbinm:
@chatbotnlp Sep 19 15:14
Hello team , I am trying to run simple titanic example Train and score method are working fine but when i try to run Evaluate method i am getting error.
```
Exception in thread "main" org.apache.spark.sql.AnalysisException: cannot resolve 'probability' given input columns: [parch, key, name_2-stagesApplied_OPVector_00000000000f, age_1-stagesApplied_OPVector_00000000000d, age-cabin-embarked-name-pClass-parch-
```
full log https://pastebin.com/mhWWht6k

PR #137, opened 2018-09-21
Set probability column for evaluation

tovbinm:
Fixes issue #136


helloworld/src/main/scala/com/salesforce/hw/titanic/OpTitanic.scala
```diff
@@ -86,6 +86,7 @@ object OpTitanic extends OpAppWithRunner with TitanicFeatures {
     .setLabelCol(survived)
     .setPredictionCol(pred)
     .setRawPredictionCol(raw)
+    .setProbabilityCol(prob)
```
leahmcguire: ooo - is this what we have in master?? 'cause this is NOT up to date with the Evaluators version currently in master

tovbinm: master is working against TAI 0.3.4. I will update 0.4.0 release branch as well.

PR #139, opened 2018-09-24
changed README.md template to generate correct shell commands

neuhausler:
Change README.md template to be able to copy/paste shell commands from the generated README.md


tovbinm:
@neuhausler thank you for he contribution. are you sure it hase to be `\\`? we used single back slash everywhere before. See - https://github.com/salesforce/TransmogrifAI/tree/master/helloworld#train

neuhausler:
If you use single \ your template-preprocessor will just concat those two lines. If you use \\ it will add a single \ to the end of the line.
If you look at your template for for example "Scoring" the last two line will just have \n at the end without the single \.  Therefor if you copy past it to your shell the shell will interpret it as a multiple command instead of concatenating them. So you either add a single \ to have those concatenated by the pre-processor, or your add \\, on which the processor will remove one of those, and the shell will afterwards command those lines based on the single \.

.. btw, I replace the \ by \\ for the train command just to be consistent with the other commands.

PR #146, opened 2018-10-02
added logging of response distribution to RFF

leahmcguire:


core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -283,6 +283,7 @@ class RawFeatureFilter[T]
     assert(trainData.count() > 0, "RawFeatureFilter cannot work with empty training data")
     val trainingSummary = computeFeatureStats(trainData, rawFeatures)
     log.info("Computed summary stats for training features")
+    log.debug(trainingSummary.responseDistributions.mkString("\n"))
```
tovbinm: perhaps guard these two lines with `if (log.isDebugEnabled)`?

PR #147, opened 2018-10-04
fixed model insights exception when features are excluded from sanity…

leahmcguire:
… checker correlation calculations

**Related issues**
Model insights used to throw runtime exception if a feature was not in the sanity checker correlation computation. However some features can be excluded from these calculations for performance so this was a problem.

If the feature is not found in the sanity checker calculations a None is returned for correlation.


core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -152,6 +153,22 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
     insights.stageInfo.keys.size shouldEqual 8
   }
 
+  it should "return model insights even when correlation is turned off for some features" in {
+    val featuresFinal = Seq(description.vectorize(10, false, 1, true),
```
tovbinm: can you please name the arguments to `vectorize` calls.

Issue #134, opened 2018-09-19
Replace AvroInOut with spark-avro or at least upgrade Hadoop API


tovbinm:
**Problem**
AvroInOut works great but it uses the old `org.apache.hadoop.mapred` Hadoop API

**Solution**
Consider replacing it with `spark-avro` or at least use the newer `org.apache.hadoop.mapreduce` Hadoop API

**Alternatives**
NA

**Additional context**
This change would require extensive testing to make sure we have the AvroInOut functionality the same.
Starting from adding unit tests and also running some real Spark jobs.

tovbinm:
Upon discussion with @himsmittal we decided to start from upgrading to the newer `org.apache.hadoop.mapreduce` Hadoop API, since `spark-avro` has several serious flaws, e.g. inability to use GenericRecord, issues with union types and namespaces.

tovbinm:
thank you for the contribution @himsmittal !!!

PR #150, opened 2018-10-05
Using MapReduce Api for Avro Read Write

himsmittal:
**Related issues**
Refer to issue(s) addressed in this pull request from [Issues]https://github.com/salesforce/TransmogrifAI/issues/134

Using Mapreduce API to read/write Avro.

Considered using Spark Avro, Spark Avro returns and writes DataFrames, but current implementation deals with RDD.

**Additional context**

### DIFF
--- a/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
+++ b/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
@@ -161,7 +161,7 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
     insights.label.rawFeatureName shouldBe Seq(survived.name)
     insights.label.rawFeatureType shouldBe Seq(survived.typeName)
     insights.label.stagesApplied.size shouldBe 1
-    insights.label.sampleSize shouldBe Some(4.0)
+    insights.label.sampleSize shouldBe Some(5.0)
     insights.features.size shouldBe 5
     insights.features.map(_.featureName).toSet shouldEqual rawNames
     ageInsights.derivedFeatures.size shouldBe 2
@@ -213,12 +213,13 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
     insights.label.rawFeatureName shouldBe Seq(survived.name)
     insights.label.rawFeatureType shouldBe Seq(survived.typeName)
     insights.label.stagesApplied.size shouldBe 1
-    insights.label.sampleSize shouldBe Some(4.0)
+    insights.label.sampleSize shouldBe Some(5.0)
     insights.features.size shouldBe 5
     insights.features.map(_.featureName).toSet shouldEqual rawNames
     ageInsights.derivedFeatures.size shouldBe 2
+    ageInsights.derivedFeatures(0).contribution.size shouldBe 1
+    ageInsights.derivedFeatures(1).contribution.size shouldBe 0
     ageInsights.derivedFeatures.foreach { f =>
-      f.contribution.size shouldBe 1
       f.corr.nonEmpty shouldBe true
       f.variance.nonEmpty shouldBe true
       f.cramersV.isEmpty shouldBe true
@@ -304,11 +305,10 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
         pretty should not include(m.modelName)
       }
     }
-    pretty should include("area under precision-recall | 0.0")
+    pretty should include("area under precision-recall | 1.0")
     pretty should include("Model Evaluation Metrics")
     pretty should include("Top Model Insights")
     pretty should include("Top Positive Correlations")
-    pretty should include("Top Negative Correlations")
     pretty should include("Top Contributions")
   }
 
--- a/utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
+++ b/utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
@@ -35,11 +35,12 @@ import java.net.URI
 import com.salesforce.op.utils.spark.RichRDD._
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
-import org.apache.avro.mapred._
+import org.apache.avro.mapred.AvroKey
+import org.apache.avro.mapreduce.{AvroJob, AvroKeyInputFormat, AvroKeyOutputFormat}
 import org.apache.avro.specific.SpecificData
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.hadoop.io.NullWritable
-import org.apache.hadoop.mapred.JobConf
+import org.apache.hadoop.mapreduce.Job
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.SparkSession
 
@@ -133,9 +134,9 @@ object AvroInOut {
     (implicit sc: SparkSession, ct: ClassTag[T]): RDD[T] = {
     def maybeCopy(r: T): T = if (deepCopy) SpecificData.get().deepCopy(r.getSchema, r) else r
 
-    val records = sc.sparkContext.hadoopFile(path,
-      classOf[AvroInputFormat[T]],
-      classOf[AvroWrapper[T]],
+    val records = sc.sparkContext.newAPIHadoopFile(path,
+      classOf[AvroKeyInputFormat[T]],
+      classOf[AvroKey[T]],
       classOf[NullWritable]
     )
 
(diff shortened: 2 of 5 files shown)


utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
```diff
@@ -35,11 +35,12 @@ import java.net.URI
 import com.salesforce.op.utils.spark.RichRDD._
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
-import org.apache.avro.mapred._
+import org.apache.avro.mapred.AvroKey
```
tovbinm: `mapreduce` still uses `AvroKey` from `mapred`. is this correct?

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
```diff
@@ -182,12 +177,12 @@ object AvroInOut {
...
      * @param path   Input directory where avro records should be written.
      * @param schema Avro schema string for records being written out.
-     * @param sc     Spark Session
      * @return
      */
-    def writeAvro(path: String, schema: String)(implicit sc: SparkSession): Unit = {
-      val jobConf = createJobConfFromContext(schema)
-      writeAvro(path)(jobConf)
+    def writeAvro(path: String, schema: String): Unit = {
+      val job: Job = Job.getInstance()
```
tovbinm: how do we pass in hadoop configuration now? before it was picked up from spark context like this:
```scala
 val jobConf = new JobConf(sc.sparkContext.hadoopConfiguration)
```

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -161,7 +161,7 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
     insights.label.rawFeatureName shouldBe Seq(survived.name)
     insights.label.rawFeatureType shouldBe Seq(survived.typeName)
     insights.label.stagesApplied.size shouldBe 1
-    insights.label.sampleSize shouldBe Some(4.0)
+    insights.label.sampleSize shouldBe Some(5.0)
```
tovbinm: @leahmcguire why do you think this was affected? same goes for `area under precision-recall | 1.0` below.

leahmcguire: These shouldn't be effected unless the data read in changed so we need to track down what is happening

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
gerashegalov: Since AvroWriter wraps the RDD it seems more correct to have a field
```scala
private implicit val job = Job.getInstance(rdd.sparkContext.hadoopConfiguration)
```
in AvroWriter

tovbinm: excellent point @gerashegalov thanks for looking into it.

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
himsmittal: class AvroKeyInputFormat<T> extends FileInputFormat<AvroKey<T>, NullWritable> . 
This uses AvroKey from mapred only.

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
```diff
@@ -154,6 +155,8 @@ object AvroInOut {
 
   implicit class AvroWriter[T <: GenericRecord](rdd: RDD[T]) {
 
+    private implicit val job = Job.getInstance(rdd.sparkContext.hadoopConfiguration)
```
tovbinm: On line `167` you can also remove `implicit job: Job` in `def writeAvro(path: String)`

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
```diff
@@ -153,27 +155,22 @@ object AvroInOut {
...
 
     /**
      * This method writes out RDDs of generic records as avro files to path.
      *
      * @param path Input directory where avro records should be written.
-     * @param jobConf job config
      * @return
      */
-    def writeAvro(path: String)(implicit jobConf: JobConf): Unit = {
+    def writeAvro(path: String): Unit = {
```
tovbinm: please make sure these functions are covered in `AvroInOutTest`, if not add tests.

tovbinm:
@himsmittal are you planning to work on this one? ;)

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
himsmittal: AvroInOutTest has one test for writeAvro already present. Will add some more tests.

himsmittal: Made this method private, Since the schema needs to be set before calling this method.

himsmittal: I guess so.
AvroKeyOutputFormat  has specific check for this 
if (null == writerSchema) {
      throw new IOException(
          "AvroKeyOutputFormat requires an output schema. Use AvroJob.setOutputKeySchema().");
    }

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
```diff
@@ -133,10 +134,11 @@ object AvroInOut {
     (implicit sc: SparkSession, ct: ClassTag[T]): RDD[T] = {
     def maybeCopy(r: T): T = if (deepCopy) SpecificData.get().deepCopy(r.getSchema, r) else r
 
-    val records = sc.sparkContext.hadoopFile(path,
-      classOf[AvroInputFormat[T]],
-      classOf[AvroWrapper[T]],
-      classOf[NullWritable]
+    val records = sc.sparkContext.newAPIHadoopFile(path,
+      classOf[AvroKeyInputFormat[T]],
```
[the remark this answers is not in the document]
himsmittal: While Reading it logs a warning : Using a reader schema equal to the writer schema.

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
himsmittal: @tovbinm @leahmcguire Did some debugging to figure out what's happening. The test uses AggregateAvroReader to read the data. Data has multiple rows for passengerId :4. Aggregated Reader aggregates the data for this particular key. Since the order in which the rows are fetched is changing, aggregation is resulting in change in Data. "Description" field is specifically changing to **"this is description stuff"** . Prior to the change, this field was aggregated as **"stuff this is description"**. Let me know if this could be the reason behind this change needed.

leahmcguire: @himsmittal as the data that you read in has 6 keys after generation:
```
+---+---------------+--------+---------------------------------------------------------------------------------------------------+
|key|     numericMap|survived|age-description-genderPL-height-numericMap-survived-weight_12-stagesApplied_Prediction_000000000027|
+---+---------------+--------+---------------------------------------------------------------------------------------------------+
|  4|  [Male -> 2.0]|   false|                                                                               [probability_1 ->...|
|  6|           null|    true|                                                                               [probability_1 ->...|
|  2|[Female -> 1.0]|    null|                                                                               [probability_1 ->...|
|  5|[Female -> 1.0]|    null|                                                                               [probability_1 ->...|
|  3|  [Male -> 1.0]|    null|                                                                               [probability_1 ->...|
|  1|[Female -> 1.0]|    null|                                                                               [probability_1 ->...|
+---+---------------+--------+---------------------------------------------------------------------------------------------------+
```
I think this may be due to the combination of the change in ordering and the randomness associated with the holdout set.

Can you add this `  modelWithRFF.score(keepIntermediateFeatures = true).show()` to the test at line 122 and make sure your output looks the same as the one posted above?

himsmittal: @leahmcguire added modelWithRFF.score(keepIntermediateFeatures = true).show() to ModelInsightsTest (line 122) . Getting the below output. The columns are different. Am I missing something here ?
```
+---+------------------------------------------------+---------------------------------------------------------------------------------------+--------------------------------------------+-------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------+
|key|numericMap_1-stagesApplied_OPVector_000000000018|age-description-genderPL-height-numericMap-weight_9-stagesApplied_OPVector_000000000019|survived_1-stagesApplied_RealNN_00000000001a|age-description-genderPL-height-numericMap-survived-weight_11-stagesApplied_OPVector_00000000001c|age-description-genderPL-height-numericMap-survived-weight_12-stagesApplied_Prediction_000000000027|
+---+------------------------------------------------+---------------------------------------------------------------------------------------+--------------------------------------------+-------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------+
|  4|                                       [2.0,0.0]|                                                                              [2.0,0.0]|                                         0.0|                                                                                            [0.0]|                                                                               [probability_1 ->...|
|  5|                                       [2.0,1.0]|                                                                              [2.0,1.0]|                                         0.0|                                                                                            [1.0]|                                                                               [probability_1 ->...|
|  6|                                       [2.0,1.0]|                                                                              [2.0,1.0]|                                         1.0|                                                                                            [1.0]|                                                                               [probability_1 ->...|
|  2|                                       [2.0,1.0]|                                                                              [2.0,1.0]|                                         0.0|                                                                                            [1.0]|                                                                               [probability_1 ->...|
|  3|                                       [1.0,0.0]|                                                                              [1.0,0.0]|                                         0.0|                                                                                            [0.0]|                                                                               [probability_1 ->...|
|  1|                                       [2.0,1.0]|                                                                              [2.0,1.0]|                                         0.0|                                                                                            [1.0]|                                                                               [probability_1 ->...|
+---+------------------------------------------------+---------------------------------------------------------------------------------------+--------------------------------------------+-------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------+
```

leahmcguire: I put in the wrong command :-P `modelWithRFF.score(keepRawFeatures = true).show()` is my printout. However this looks right so I think we are good to go. Thanks for checking!

himsmittal: Thanks :)

PR #152, opened 2018-10-08
Streaming histogram implementation

marcovivero:
**Related issues**
This is an enhancement.

Add additional StreamingHistogram implementation as specified here:

http://www.jmlr.org/papers/volume11/ben-haim10a/ben-haim10a.pdf

This is an alternative to the current histogram binning strategy performed in FeatureDistribution, we do a comparison on density estimation given by the two binning methods.


utils/src/main/java/com/salesforce/op/utils/stats/HistogramJavaUtils.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package com.salesforce.op.utils.stats;
+
+import java.io.Serializable;
+import java.util.Comparator;
+import java.util.TreeMap;
+import java.util.function.Function;
+
+final public class HistogramJavaUtils {
```
tovbinm: why do you need this? why not use `scala.collection.immutable.TreeMap`?

utils/src/main/scala/com/salesforce/op/utils/stats/StreamingHistogram.scala
```diff
@@ -0,0 +1,181 @@
...
+            }
+          }
+        } match {
+          case (s, (p, m), true) => s
+          case (s, (p, m), false) =>
+            s + { if (x == p) m / 2 else m }
+        }
+      case b if b.isEmpty => 0.0
+      case b =>
+        val (p, m) = b(0) // We must have exactly one element
```
tovbinm: why not have this if/else in the case?

utils/src/main/scala/com/salesforce/op/utils/stats/StreamingHistogram.scala
```diff
@@ -0,0 +1,181 @@
...
+    // Remove the old points from keyset
+    points.remove(q1)
+    points.remove(q2)
+
+    // Add new point to bins.
+    points.put((q1 * k1 + q2 * k2) / (k1 + k2), k1 + k2)
+  }
+
+  private def updatePoints(updates: (Double, Double)*): Unit = {
+    updates.foreach { case (point, ct) =>
```
tovbinm: why mutable?!

utils/src/test/scala/com/salesforce/op/utils/stats/StreamingHistogramTest.scala
```diff
@@ -0,0 +1,262 @@
...
+    val MeanAndVariance(streamingSumCDFMSEMean, streamingSumCDFMSEVar, _) =
+      meanAndVariance(mses.map(_.streamingSumCDFMSE))
+    val MeanAndVariance(streamingEmpiricalCDFMSEMean, streamingEmpiricalCDFMSEVar, _) =
+      meanAndVariance(mses.map(_.streamingEmpiricalCDFMSE))
+    val MeanAndVariance(equiDistSumCDFMSEMean, equiDistSumCDFMSEVar, _) =
+      meanAndVariance(mses.map(_.equiDistSumCDFMSE))
+    val MeanAndVariance(equiDistEmpiricalCDFMSEMean, equiDistEmpiricalCDFMSEVar, _) =
+      meanAndVariance(mses.map(_.equiDistEmpiricalCDFMSE))
+
+    println("-" * 50)
```
tovbinm: extract printing into a helper function in the spec + replace println with log.info

utils/src/test/scala/com/salesforce/op/utils/stats/StreamingHistogramTest.scala
```diff
@@ -0,0 +1,262 @@
...
+
+  private def round(x: Double): Double = math.round(x * 100).toDouble / 100
+}
+
+object StreamingHistogramTest {
+
+  type Distribution[T] = Density[T] with Rand[T]
+
+  case class MixtureDistribution(
+      d1: Distribution[Double] with HasCdf,
```
tovbinm: where `HasCdf` trait is coming from?

utils/src/test/scala/com/salesforce/op/utils/stats/StreamingHistogramTest.scala
```diff
@@ -0,0 +1,262 @@
...
+    cdfTest("Beta(5, 1)", beta1, grid, sampleSize, 400, 50)
+    cdfTest("Beta(1, 5)", beta2, grid, sampleSize, 100, 50)
+    cdfTest("Beta(1, 5)", beta2, grid, sampleSize, 200, 50)
+    cdfTest("Beta(1, 5)", beta2, grid, sampleSize, 400, 50)
+    cdfTest("0.5 * Beta(5, 1) + 0.5 * Beta(1, 5)", mixture, grid, sampleSize, 100, 50)
+    cdfTest("0.5 * Beta(5, 1) + 0.5 * Beta(1, 5)", mixture, grid, sampleSize, 200, 50)
+    cdfTest("0.5 * Beta(5, 1) + 0.5 * Beta(1, 5)", mixture, grid, sampleSize, 400, 50)
+  }
+
+  private def cdfTest(
```
tovbinm: why there is so many private nested methods in this tests? it makes it very difficult to follow what actually the test does. can you please simplify if and use more of scalatest matchers syntax?

utils/src/test/scala/com/salesforce/op/utils/stats/StreamingHistogramTest.scala
tovbinm: breeze. I see.

utils/src/main/java/com/salesforce/op/utils/stats/HistogramJavaUtils.java
```diff
@@ -0,0 +1,46 @@
...
+    final public static <T> TreeMap<Double, T> getTreeMap() {
+        return new TreeMap<Double, T>(getComparator());
```
tovbinm: you should be good to use TreeMap since it's already has a default serializer defined - https://github.com/EsotericSoftware/kryo/blob/kryo-parent-3.0.3/src/com/esotericsoftware/kryo/Kryo.java#L207

marcovivero:
@leahmcguire I was primarily trying to avoid bringing in any additional dependencies into the project.

utils/src/main/scala/com/salesforce/op/utils/stats/StreamingHistogram.scala
```diff
@@ -0,0 +1,206 @@
...
+   * Performs sum algorithm outlined in paper evaluated at some input point, which approximates the number of
+   * points less than or equal to the input point.
+   *
+   * @param x input point
+   * @return approximate number of points that are less than or equal to the input point
+   */
+  private[stats] def sum(bins: Array[Bin], x: Double): Double =
+    bins.sliding(2).foldLeft((0.0, EmptyBin, false)) { case ((s, _, done), arr) =>
+      arr match {
+        case Array((p, m)) =>
```
kinfaikan: Do we need this case?

utils/src/main/scala/com/salesforce/op/utils/stats/StreamingHistogram.scala
```diff
@@ -0,0 +1,206 @@
...
+          val newS = if (x == p) m / 2
+            else if (x > m) m
+            else 0.0
+
+          (newS, EmptyBin, true)
+
+        case Array((p1, m1), (p2, m2)) =>
+          if (x >= p1 && x < p2 && !done) {
```
kinfaikan: Slightly simplified?
```
if (done || x < p1) {
  (s, (p2, m2), true)
} else if (x >= p1 && x < p2) {
  val mx = m1 + ((m2 - m1) / (p2 - p1)) * (x - p1)
  val newS = s + (m1 / 2) + ((m1 + mx) / 2) * ((x - p1) / (p2 - p1))
  (newS, (p2, m2), true)
} else {
  (s + m1, (p2, m2), false)
}
```

utils/src/main/scala/com/salesforce/op/utils/stats/StreamingHistogram.scala
```diff
@@ -0,0 +1,206 @@
...
+   */
+  final def sumCDF(x: Double): Double = StreamingHistogram.sumCDF(getBins, x)
+
+  /**
+   * @return empirical CDF estimate
+   */
+  final def empiricalCDF(x: Double): Double = StreamingHistogram.empiricalCDF(getBins, x)
+
+  private[this] def mergePoints(): Unit = if (points.size > math.max(maxBins, 2)) {
+    val (q1, q2) = points.descendingKeySet.descendingIterator.asScala.sliding(2).map(_ match {
```
kinfaikan: Counts are not considered in points merging?

leahmcguire:
small dependancies for tested and maintained code are better than reinventing the wheel.

marcovivero:
@leahmcguire: per @tovbinm's suggestion I grabbed out the code from Cassandra implementation and modified a bit for this use case.

utils/src/main/scala/com/salesforce/op/utils/kryo/OpKryoRegistrator.scala
```diff
@@ -67,6 +70,15 @@ class OpKryoRegistrator extends KryoRegistrator {
       classOf[GenericData.Array[_]],
       new GenericJavaCollectionSerializer(classOf[java.util.ArrayList[_]])
     )
+
+    // Streaming histogram registration
+    doClassRegistration(kryo)(
```
tovbinm: hmm, we should not be doing this for `TreeMap` and `mutable.WrappedArray`, as those should be already registered with Kryo.

utils/src/main/scala/com/salesforce/op/utils/stats/HistogramUtils.scala
```diff
@@ -0,0 +1,94 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.salesforce.op.utils.stats
+
+import scala.collection.JavaConverters._
+
+object HistogramUtils {
```
tovbinm: docs please: for object and methods.

utils/src/main/scala/com/salesforce/op/utils/kryo/OpKryoRegistrator.scala
marcovivero: This flag causes a failure if these aren't explicitly registered here:

https://github.com/salesforce/TransmogrifAI/pull/152/files#diff-36f3a0a7cf47675bc10aac982dab7fecR51

tovbinm: This is weird and to my understanding should not be the case. Let's find time to look at together.

leahmcguire:
@marcovivero if this is copy pasted code the license file needs to reflect that

utils/src/main/java/com/salesforce/op/utils/stats/StreamingHistogram.java
```diff
@@ -0,0 +1,300 @@
...
+import com.google.common.base.Objects;
+
+/**
+ * Histogram that can be constructed from streaming of data.
+ *
+ * The algorithm is taken from following paper:
+ * Yael Ben-Haim and Elad Tom-Tov, "A Streaming Parallel Decision Tree Algorithm" (2010)
+ * http://jmlr.csail.mit.edu/papers/volume11/ben-haim10a/ben-haim10a.pdf
+ *
+ * This implementation is from Apache Cassandra roject, release 3.11.
```
leahmcguire: project

utils/src/test/scala/com/salesforce/op/utils/stats/StreamingHistogramTest.scala
```diff
@@ -0,0 +1,250 @@
...
+      s"[bins = $maxBins, sample size = $histogramSampleSize, iterations = $numResults]")
+    println("-" * 50)
+    println("Streaming histogram density MSE mean and variance: " +
+      s"${result.streamingDensityMSE.mean}, ${result.streamingDensityMSE.variance}")
+    println("Streaming histogram cdf MSE mean and variance: " +
+      s"${result.streamingCdfMSE.mean}, ${result.streamingCdfMSE.variance}")
+    println("Equidistant histogram density MSE mean and variance: " +
+      s"${result.equiDistDensityMSE.mean}, ${result.equiDistDensityMSE.variance}")
+    println("Equidistant histogram cdf MSE mean and variance: " +
+      s"${result.equiDistCdfMSE.mean}, ${result.equiDistCdfMSE.variance}")
```
leahmcguire: please add some asserts and change these prints to logs that can be turned on or off

utils/src/main/java/com/salesforce/op/utils/stats/StreamingHistogram.java
```diff
@@ -0,0 +1,300 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.salesforce.op.utils.stats;
+
+import java.io.IOException;
```
tovbinm: no need to `import java.io.IOException;`

utils/src/test/scala/com/salesforce/op/utils/stats/StreamingHistogramTest.scala
```diff
@@ -0,0 +1,250 @@
...
+      (0 to 10).map(k => (k.toDouble, 1.0))
+  }
+
+  it should "correctly approximate standard normal distribution" in {
+    val sampleSize = 500
+    val gaussian1 = Gaussian(0, 1)
+    val gaussian2 = Gaussian(5, 5)
+    val mixture = MixtureDistribution(gaussian1, gaussian2, 0.8, sampleSize)
+
+    cdfTest("Gaussian(0, 1)", gaussian1, histogramSampleSize, 100, mcSampleSize, 50)
```
tovbinm: let's not run every test 50 times. that is very slow. let's add this as val and maybe do just 3.

utils/src/main/scala/com/salesforce/op/utils/kryo/OpKryoRegistrator.scala
```diff
@@ -67,6 +70,15 @@ class OpKryoRegistrator extends KryoRegistrator {
...
+      classOf[StreamingHistogram],
+      classOf[StreamingHistogramBuilder],
+      classOf[StreamingHistogramComparator],
+      classOf[TreeMap[_, _]],
```
tovbinm: do this instead
```scala
 // Streaming histogram registration
    doClassRegistration(kryo)(
      classOf[StreamingHistogram],
      classOf[StreamingHistogramBuilder],
      classOf[StreamingHistogramComparator]
    )
kryo.register(classOf[TreeMap[_, _]], new TreeMapSerializer())
 doClassRegistration(kryo)(OpKryoClasses.WrappedArrays: _*)
```
where

```
scala
private[op] case object OpKryoClasses {
   

  val WrappedArrays: Seq[Class[_]] = Seq(
    MWrappedArray.make(Array[Boolean]()).getClass,
    MWrappedArray.make(Array[Byte]()).getClass,
    MWrappedArray.make(Array[Char]()).getClass,
    MWrappedArray.make(Array[Double]()).getClass,
    MWrappedArray.make(Array[Float]()).getClass,
    MWrappedArray.make(Array[Int]()).getClass,
    MWrappedArray.make(Array[Long]()).getClass,
    MWrappedArray.make(Array[String]()).getClass,
    MWrappedArray.make(Array[Short]()).getClass
  )
```

utils/src/test/scala/com/salesforce/op/utils/stats/StreamingHistogramTest.scala
```diff
@@ -0,0 +1,267 @@
...
+    pdf(5.0) shouldEqual 0.0
+  }
+
+  // Checks that it does well for well-behaved distributions
+  it should "correctly approximate standard normal distribution" in {
+    val sampleSize = 500
+    val gaussian = Gaussian(0, 1)(RandBasis.mt0)
+    val distributionName = "Gaussian(0, 1)"
+
+    val result75 = distributionTestResult(distributionName, gaussian, histogramSampleSize, 75, mcSampleSize, 5)
```
tovbinm: please add `val numResults = 5` in the head of the test.

utils/src/main/java/com/salesforce/op/utils/stats/StreamingHistogram.java
```diff
@@ -0,0 +1,299 @@
+/*
```
tovbinm: I think you need to add `// scalastyle:off header.matches` line here.

tovbinm: and also include it to the exclusion list - https://github.com/salesforce/TransmogrifAI/blob/master/build.gradle#L205

marcovivero: This class won't go through `scalastyle` since in `src/main/java` so I don't think that's needed.

Issue #151, opened 2018-10-08
sanitychecker seems do not work for realmap featuretype


liuzhenhai93:
**Describe the bug**
sanitychecker seems do not work for realmap featuretype

**Logs or screenshots**
```
18/10/08 10:57:58 ERROR ApplicationMaster: User class threw exception: java.lang.IllegalArgumentException: requirement failed: Vector column 99 has multiple null indicator fields: ArrayBuffer((99,({
  "parentFeatureName" : [ "delv_way_cd_count" ],
  "parentFeatureType" : [ "com.salesforce.op.features.types.RealMap" ],
  "grouping" : "99",
  "indicatorValue" : "NullIndicatorValue",
  "index" : 87,
  "nullIndicator" : true
},87)), (99,({
  "parentFeatureName" : [ "pay_mode_cd_count" ],
  "parentFeatureType" : [ "com.salesforce.op.features.types.RealMap" ],
  "grouping" : "99",
  "indicatorValue" : "NullIndicatorValue",
  "index" : 126557,
  "nullIndicator" : true
},126557)))
java.lang.IllegalArgumentException: requirement failed: Vector column 99 has multiple null indicator fields: ArrayBuffer((99,({
  "parentFeatureName" : [ "delv_way_cd_count" ],
  "parentFeatureType" : [ "com.salesforce.op.features.types.RealMap" ],
  "grouping" : "99",
  "indicatorValue" : "NullIndicatorValue",
  "index" : 87,
  "nullIndicator" : true
},87)), (99,({
  "parentFeatureName" : [ "pay_mode_cd_count" ],
  "parentFeatureType" : [ "com.salesforce.op.features.types.RealMap" ],
  "grouping" : "99",
  "indicatorValue" : "NullIndicatorValue",
  "index" : 126557,
  "nullIndicator" : true
},126557)))
	at com.salesforce.op.stages.impl.preparators.SanityChecker$$anonfun$makeColumnStatistics$2.apply(SanityChecker.scala:286)
	at com.salesforce.op.stages.impl.preparators.SanityChecker$$anonfun$makeColumnStatistics$2.apply(SanityChecker.scala:284)
	at scala.collection.immutable.HashMap$HashMap1.foreach(HashMap.scala:221)
	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:428)
	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:428)
[...]

PR #153, opened 2018-10-08
Added test and removed dead code for Sanity Checker dealing with maps with same key

leahmcguire:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/151

Looks like this error was caused by some dead code in sanity checker. I removed this code and added a test for maps with the same keys.


core/src/test/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerTest.scala
```diff
@@ -611,6 +611,19 @@ class SanityCheckerTest extends OpEstimatorSpec[OPVector, BinaryModel[RealNN, OP
...
+    val mapData = textRawData.map{
+      case (i, t, tm) => (i, t, tm, tm, tm.value.map{ case (k, v) => k -> math.random }.toRealMap)
+    }
+    val (mapDataFrame, id, target, textMap1, textMap2, doubleMap) = TestFeatureBuilder(
+      "id", "target", "textMap1", "textMap2", "doubleMap", mapData)
+    val targetResponse: FeatureLike[RealNN] = target.copy(isResponse = true)
+    val features = Seq(id, target, textMap1, textMap2, doubleMap).transmogrify()
+    val checked = targetResponse.sanityCheck(features)
+    val output = new OpWorkflow().setResultFeatures(checked).transform(mapDataFrame)
+    output.show()
```
tovbinm: 1. please add some actual checks. perhaps check metadata produced?
2. please remove `output.show()`

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
```diff
@@ -274,18 +274,6 @@ class SanityChecker(uid: String = UID[SanityChecker])
...
-    // build map from indicator group name to any null indicator it may have
-    val nullGroups = for {
-      col <- metaCols
-      if col.isNullIndicator
-        grouping <- col.grouping
-      } yield (grouping, (col, col.index))
-
-    nullGroups.groupBy(_._1).foreach {
-      case (group, cols) =>
-        require(cols.length == 1, s"Vector column $group has multiple null indicator fields: $cols")
```
tovbinm: since you are removing this check, also remove this ignored test - https://github.com/salesforce/TransmogrifAI/blob/f37113b4dbbd9078b3315ebfe0ab36975f83493a/core/src/test/scala/com/salesforce/op/stages/impl/preparators/BadFeatureZooTest.scala#L107

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
```diff
@@ -274,18 +274,6 @@ class SanityChecker(uid: String = UID[SanityChecker])
...
     val numCorrIndices = corrIndices.length
 
-    // build map from indicator group name to any null indicator it may have
-    val nullGroups = for {
-      col <- metaCols
-      if col.isNullIndicator
-        grouping <- col.grouping
-      } yield (grouping, (col, col.index))
-
-    nullGroups.groupBy(_._1).foreach {
```
Jauntbox: Do you think we should take out the whole check for multiple null indicators right now?

Afaik, we still have an issue when a single parent feature is vectorized with two different unary stages (eg. if you were to vectorize a real with both the normal `RealVectorizer` and also bucketize). SanityChecker will still group these into a single contingency matrix since their grouping will just be the parent feature name, so I think we should leave this check in until we fix that issue.

leahmcguire: Do you really think it is better to fail the whole flow when that happens? I would be better to automatically drop one than fail the flow...

leahmcguire: Or even have some incorrect statistics (in my opinion)

Jauntbox: Can we at least keep it in as a warning we log so that we can monitor when it's happening?

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorColumnMetadata.scala
```diff
@@ -151,6 +151,12 @@ case class OpVectorColumnMetadata // TODO make separate case classes extending t
     if (hasParentOfSubType[OPMap[_]]) parentFeatureName.map(p => grouping.map(p + "_" + _).getOrElse(p))
     else parentFeatureName
 
+  /**
+   * Get the feature grouping qualified by the parent feature name
+   * @return Optional string of feature grouping
+   */
+  def featureGroup(): Option[String] = grouping.map(g => s"${parentFeatureName.mkString}${g}")
```
tovbinm: dont you want to do something similar to `s"${parentFeatureName.mkString("_")}${grouping.map("_" + _).getOrElse("")}"` (same as we do in [`makeColName()`](https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/utils/spark/OpVectorColumnMetadata.scala#L125) function)

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
leahmcguire: @Jauntbox I fixed it so that it will remove duplicates from the categorical calculations

core/src/test/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerTest.scala
```diff
@@ -35,9 +35,14 @@ import com.salesforce.op.features.FeatureLike
 import com.salesforce.op.features.types._
 import com.salesforce.op.stages.MetadataParam
 import com.salesforce.op.stages.base.binary.{BinaryEstimator, BinaryModel}
+<<<<<<< HEAD
 import com.salesforce.op.stages.impl.feature.{HashSpaceStrategy, RealNNVectorizer, SmartTextMapVectorizer}
 import com.salesforce.op.test.{OpEstimatorSpec, TestFeatureBuilder, TestSparkContext}
 import com.salesforce.op.utils.json.JsonUtils
+=======
```
tovbinm: @leahmcguire merge conflit?

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorColumnMetadata.scala
```diff
@@ -151,6 +151,12 @@ case class OpVectorColumnMetadata // TODO make separate case classes extending t
     if (hasParentOfSubType[OPMap[_]]) parentFeatureName.map(p => grouping.map(p + "_" + _).getOrElse(p))
     else parentFeatureName
 
+  /**
+   * Get the feature grouping qualified by the parent feature name
+   * @return Optional string of feature grouping
+   */
+  def featureGroup(): Option[String] = grouping.map(g => s"${parentFeatureName.mkString("_")}_$g")
```
tovbinm: @leahmcguire repeating my previous comment here -  dont you want to do something similar to `s"${parentFeatureName.mkString("_")}${grouping.map("_" + _).getOrElse("")}"` (same as we do in [`makeColName()`](https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/utils/spark/OpVectorColumnMetadata.scala#L125) function)

tovbinm: Especially because you do call `.get` here - https://github.com/salesforce/TransmogrifAI/pull/153/files#diff-55c6247553ff142cb1657c1bac3c728fR459

leahmcguire: no because before calling that get there is a check on whether this is a None which determines if the categorical stats are computed

leahmcguire: actually the check is directly on the group - I will change this

PR #167, opened 2018-10-26
Time based aggregators

leahmcguire:
Made aggregators for first and last event value for all feature types.


features/src/main/scala/com/salesforce/op/aggregators/TimeBasedAggregator.scala
```diff
@@ -0,0 +1,250 @@
...
+import com.twitter.algebird.{Monoid, MonoidAggregator}
+
+import scala.reflect.runtime.universe.WeakTypeTag
+
+private[op] abstract class TimeBasedAggregator[T <: FeatureType]
+(
+  val compareFun: (Long, Long) => Boolean,
+  val timeZero: Long,
+  val emptyValue: T#Value
+)(
```
tovbinm: indentation is very very weird in these classes

features/src/main/scala/com/salesforce/op/aggregators/TimeBasedAggregator.scala
```diff
@@ -0,0 +1,250 @@
...
+package com.salesforce.op.aggregators
+
+import com.salesforce.op.features.types._
+import com.twitter.algebird.{Monoid, MonoidAggregator}
+
+import scala.reflect.runtime.universe.WeakTypeTag
+
+private[op] abstract class TimeBasedAggregator[T <: FeatureType]
+(
+  val compareFun: (Long, Long) => Boolean,
```
tovbinm: Is there any particular reason in making this function a value argument?

features/src/main/scala/com/salesforce/op/aggregators/TimeBasedAggregator.scala
```diff
@@ -0,0 +1,250 @@
...
+  val timeZero: Long,
+  val emptyValue: T#Value
+)(
+  implicit val ttag: WeakTypeTag[T],
+  val ttvag: WeakTypeTag[T#Value]
```
tovbinm: I dont think you need `val ttvag: WeakTypeTag[T#Value]` at all.

features/src/main/scala/com/salesforce/op/aggregators/TimeBasedAggregator.scala
```diff
@@ -0,0 +1,250 @@
...
+)
+  extends MonoidAggregator[Event[T], (Long, T#Value), T] {
+  val ftFactory = FeatureTypeFactory[T]()
+
+  val monoid: Monoid[(Long, T#Value)] = new Monoid[(Long, T#Value)] {
+    val zero = timeZero -> emptyValue
```
tovbinm: I think you dont need to pass zeroValue, instead simply:
```
val zero = timeZero -> FeatureTypeDefaults.default[T].value
```

leahmcguire: I have to remove support for RealNN since it doesn't have a default - but that may be better anyway

tovbinm:
@leahmcguire scalastyle is no happy.

PR #168, opened 2018-10-27
Improve test coverage for VectorsCombiner and make vector aggregator efficient 

tovbinm:
**Related issues**
1. `UnionVector` monoid aggregator is not efficient for sparse vectors.
2. `VectorsCombiner` stage has incomplete functional test coverage.

1. Moved `VectorsCombiner.combine` method to `RichVector` object
2. Added `OpEstimatorSpec` base class to `VectorsCombinerTest` to improve the functional test coverage
3. Added `+`, `-`, `dot`,`combine` methods to `OPVector` + test `OPVectorTest`
4. Added `combine` method to `RichVector` + tests
5. Renamed `UnionVector` monoid aggregator to `CombineVector` and made use of `RichVector.combine` at reduve
6. Bonus: added `SumVector` monoid aggregator

I could relocate `VectorsCombiner.combine` as-is to `utils` subproject, but it felt valuable enough to add `+`, `-`,`dot` and `combine` to `OPVector` as well.

### DIFF
--- a/features/src/main/scala/com/salesforce/op/features/types/OPVector.scala
+++ b/features/src/main/scala/com/salesforce/op/features/types/OPVector.scala
@@ -30,6 +30,7 @@
 
 package com.salesforce.op.features.types
 
+import com.salesforce.op.utils.spark.RichVector._
 import org.apache.spark.ml.linalg._
 
 /**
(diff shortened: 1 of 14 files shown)


features/src/main/scala/com/salesforce/op/features/types/OPVector.scala
```diff
@@ -39,8 +40,37 @@ import org.apache.spark.ml.linalg._
...
+  def +(that: OPVector): OPVector = (value + that.value).toOPVector
+
+  /**
+   * Subtract vectors
+   *
+   * @param that another vector
+   * @throws IllegalArgumentException if the vectors have different sizes
+   * @return vector subtraction
+   */
+  def -(that: OPVector): OPVector = (value - that.value).toOPVector
```
leahmcguire: as long as you are adding addition and subtraction why not dot product and cross product?

features/src/main/scala/com/salesforce/op/utils/spark/RichVector.scala
```diff
@@ -64,6 +66,16 @@ object RichVector {
...
 
+    /**
+     * Combine multiple vectors into one
+     *
+     * @param that  another vector
+     * @param other other vectors
+     * @return result vector
+     */
+    def combine(that: Vector, other: Vector*): Vector =
+      com.salesforce.op.utils.spark.RichVector.combine(v +: that +: other)
```
leahmcguire: do you need the full import here?

features/src/test/scala/com/salesforce/op/features/types/OPVectorTest.scala
```diff
@@ -0,0 +1,90 @@
...
+    for {(v1, v2) <- vectors.zip(vectors)} {
+      (v1 + v2) shouldBe (v1.value + v2.value).toOPVector
+    }
+  }
+
+  it should "'-' subtract correctly" in {
+    for {(v1, v2) <- vectors.zip(vectors)} {
+      (v1 - v2) shouldBe (v1.value - v2.value).toOPVector
+    }
+  }
```
leahmcguire: maybe tests for the errors when the vectors have different lengths?

features/src/main/scala/com/salesforce/op/utils/spark/RichVector.scala
tovbinm: this is to distinguish the object from the implicit class.

tovbinm:
@leahmcguire done

PR #170, opened 2018-11-02
Fix ModelInsights for xgboost

kinfaikan:
**Related issues**
`OpWorkflowModel::modelInsights` fails for xgboost because feature scores are not extracted correctly

Convert feature score map to a vector in `ModelInsights::getModelInsights`
Handle null stage param values in `ModelInsights::getStageInfo`
Do not create a table when no validation results is found

NA

**Additional context**
NA


core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -170,8 +171,10 @@ case class ModelInsights
       val params = e.modelParameters.filterKeys(!excludedParams.contains(_))
       Seq("name" -> e.modelName, "uid" -> e.modelUID, "modelType" -> e.modelType) ++ params
     }).flatten.sortBy(_._1)
-    val table = Table(name = name, columns = Seq("Model Param", "Value"), rows = validationResults)
-    table.prettyString()
+    if (validationResults.nonEmpty) {
+      val table = Table(name = name, columns = Seq("Model Param", "Value"), rows = validationResults)
+      table.prettyString()
+    } else "No validation results found"
```
tovbinm: I would rather change the return type of this function to `Seq[String]` and return `Seq.empty` if `validationResults` are empty.

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -527,10 +530,9 @@ case object ModelInsights {
     blacklistedMapKeys: Map[String, Set[String]],
     rawFeatureDistributions: Array[FeatureDistribution]
   ): Seq[FeatureInsights] = {
-    val contributions = getModelContributions(model)
-
     val featureInsights = (vectorInfo, summary) match {
       case (Some(v), Some(s)) =>
+        val contributions = getModelContributions(model, Option(v.getColumnHistory().length))
```
tovbinm: did you mean `Option(v.getColumnHistory()).map(_.length)`? how do we even get null here?! and the same below.

core/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostParams.scala
```diff
@@ -81,6 +82,25 @@ case object OpXGBoost {
     }
   }
 
+  implicit class RichBooster(val booster: Booster) extends AnyVal {
+    /**
+     * Converts feature score map into a vector
```
tovbinm: docs please

core/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostParams.scala
```diff
@@ -81,6 +82,25 @@ case object OpXGBoost {
...
+     *
+     * @param featureVectorSize
+     * @return
+     */
+    def getFeatureScoreVector(featureVectorSize: Option[Int]): Vector = {
```
tovbinm: 1. why use `Vector` as return type for if you are doing to `toArray` anyways?
2. should we have a default value `featureVectorSize: Option[Int] = None`?

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -537,4 +537,15 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
     insights.features.foreach(f => f.distributions shouldBe empty)
   }
 
+  it should "return model insights for xgboost classification" in {
+    val prediction = new OpXGBoostClassifier().setInput(label, features).setSilent(1).getOutput()
+    val model = new OpWorkflow().setResultFeatures(prediction).setReader(dataReader).train()
+    noException should be thrownBy model.modelInsights(prediction)
```
tovbinm: can we do some more asserts here on the actual model insights? same below.

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -668,7 +674,8 @@ case object ModelInsights {
         case p if p.param.name == OpPipelineStageParamsNames.InputFeatures =>
           p.param.name -> p.value.asInstanceOf[Array[TransientFeature]].map(_.toJsonString()).mkString(", ")
         case p if p.param.name != OpPipelineStageParamsNames.OutputMetadata &&
-          p.param.name != OpPipelineStageParamsNames.InputSchema => p.param.name -> p.value.toString
+          p.param.name != OpPipelineStageParamsNames.InputSchema =>
+          p.param.name -> Option(p.value).map(_.toString).getOrElse("null")
```
tovbinm: not sure what we want here `"null"` seems like a weird choice ;) perhaps we should skip the nulls completely. @leahmcguire

leahmcguire: If this is to take care of cases when the parameter is unset then null makes sense...though I suppose we could flatmap and just not show those params

core/src/main/scala/com/salesforce/op/ModelInsights.scala
leahmcguire: It doesnt look like you are actually using any of the additional info in the column history - you can just use v.columns.length

core/src/main/scala/com/salesforce/op/ModelInsights.scala
kinfaikan: I will omit the null params.

core/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostParams.scala
tovbinm: @kinfaikan ^

PR #171, opened 2018-11-02
Mt/fd type

tovbinm:
**Related issues**
Scoring feature distributions are not exposed from `RawFeatureFilter`

Return training and scoring feature distributions in single collection from `RawFeatureFilter`, but add a type to distinguish between them.

- Added `type` enum field for `FeatureDistribution` with `Training` and `Scoring` possible values.
- Made sure OpWorkflow can read / write old feature distributions (without the `type`).
- Added tests

Add an alternative collection with scoring feature distributions everywhere in RawFeatureFilter, OpWorkflow etc.

**Additional context**
This is an important piece of information that we don't want to loose ;)

### DIFF
--- a/core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
+++ b/core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
@@ -82,7 +82,6 @@ case class FeatureDistribution
     )
     check("Name", name, fd.name)
     check("Key", key, fd.key)
-    check("Type", `type`, fd.`type`)
   }
 
   /**
@@ -103,7 +102,7 @@ case class FeatureDistribution
     val combinedDist = distribution + fd.distribution
     // summary info can be empty or min max if hist is empty but should otherwise match so take the longest info
     val combinedSummary = if (summaryInfo.length > fd.summaryInfo.length) summaryInfo else fd.summaryInfo
-    FeatureDistribution(name, key, count + fd.count, nulls + fd.nulls, combinedDist, combinedSummary)
+    FeatureDistribution(name, key, count + fd.count, nulls + fd.nulls, combinedDist, combinedSummary, `type`)
   }
 
   /**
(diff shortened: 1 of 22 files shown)


core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -135,57 +148,100 @@ case class FeatureDistribution
...
 
   val MaxBins = 100000
 
   implicit val semigroup: Semigroup[FeatureDistribution] = new Semigroup[FeatureDistribution] {
-    override def plus(l: FeatureDistribution, r: FeatureDistribution) = l.reduce(r)
+    override def plus(l: FeatureDistribution, r: FeatureDistribution): FeatureDistribution = l.reduce(r)
+  }
+
+  implicit val formats: Formats = DefaultFormats +
+    EnumEntrySerializer.json4s[FeatureDistributionType](FeatureDistributionType)
```
Jauntbox: So we have a bunch of these sitting in https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/stages/OpPipelineStageReadWriteShared.scala#L79-L86

Should we we add this one there instead? Or put the others in their corresponding companion objects?

features/src/main/scala/com/salesforce/op/features/FeatureLike.scala
```diff
@@ -87,6 +82,26 @@ trait FeatureLike[O <: FeatureType] {
...
+   * The distribution information of the feature
+   * (is a sequence because map features have distribution for each key)
+   */
+  val distributions: Seq[FeatureDistributionLike]
+
+  /**
+   * The distribution information of the feature computed during training
+   * (is a sequence because map features have distribution for each key)
+   */
+  final def trainingDistributions: Seq[FeatureDistributionLike] =
```
Jauntbox: This is a bit weird to have as a sequence now since there should only be one training distribution for a feature, right? I'm okay with keeping the type as is for simplicity, but we should probably put it in the documentation if we expect it to always be a Seq of length one. Is this right, @leahmcguire ?

Issue #56, opened 2018-08-15
Parquet readers


tovbinm:
**Problem**
Missing implementations for Parquet readers.

**Solution**
Implement Parquet data readers: simple, aggregate and conditional.

**Alternatives**
N/A

Issue #83, opened 2018-08-24
Moere test specs


tovbinm:
**Problem**
Currently there is no standard way for users to test out their readers, features, workflows and apps.

**Solution**
Similarly to `OpTransformer` and OpEstimatorSpec` we provide test facilities to allow testing readers, features, workflows and apps: `OpReaderSpec`, `OpFeatureSpec`, `OpWorkflowSpec` and `OpAppSpec` accordingly.

**Alternatives**
N/A

rachelwarren:
I am working on adding the OpReaderSpec

PR #172, opened 2018-11-04
Added aggregate & conditonal readers for parquet

ajayborra:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/56

Added Aggregate & Conditional Reader API for parquet format


tovbinm:
@ajayborra please add some tests. thanks.

readers/src/main/scala/com/salesforce/op/readers/ParquetProductReader.scala
```diff
@@ -56,3 +56,33 @@ class ParquetProductReader[T <: Product : Encoder]
...
+  * @param key      function for extracting key from record
+  * @param conditionalParams params for conditional aggregation
+  * @tparam T
+  */
+class ConditionalParquetProductReader[T <: Product : Encoder : WeakTypeTag]
+(
+  readPath: Option[String],
+  key: T => String,
+  val conditionalParams: ConditionalParams[T]
+)extends ParquetProductReader[T](readPath, key) with ConditionalDataReader[T]
```
tovbinm: please fix scalastyle warnings

ajayborra: @tovbinm Seems like we have are going to have [similar logic](https://github.com/salesforce/TransmogrifAI/blob/master/readers/src/test/scala/com/salesforce/op/readers/CSVProductReadersTest.scala#L119-L163) for testing these readers. Will look at the possibility of extracting these as Table Tests and update the PR.

vpatryshev: Any tests included? Why not?

tovbinm:
@ajayborra Rachel is working on adding a common spec for all the readers, so dont sweat it too much ;) https://github.com/salesforce/TransmogrifAI/issues/83

PR #175, opened 2018-11-06
Add default param grid for xgboost

kinfaikan:
**Related issues**
XGBoost does not have default param grid.

XGBoost has quite many hyperparameters. Consider only `numRound`, `eta`, `maxDepth`, and `minChildWeight` in default param grid.

NA

**Additional context**
NA


tovbinm:
@kinfaikan please add xgboost to model selector enums and tests

PR #176, opened 2018-11-07
Refactored splitter tests

tovbinm:
**Related issues**
Test code was too repetitive and was not testing splitter summary correctly.

**Describe the proposed solution**
- Added `SplitterSummaryAsserts` for asserting `SplitterSummary` classes 
- Refactored tests for `DataCutter`, `DataSplitter`, `DataBalancer` and `OpValidator` to use `SplitterSummaryAsserts`
- Minor idents and cleanups

**Describe alternatives you've considered**
NA

**Additional context**
TBD - also needs checks & tests for empty datasets - https://github.com/salesforce/TransmogrifAI/issues/161

PR #178, opened 2018-11-07
Evaluators check for empty data

tovbinm:
**Related issues**
Most of our evaluators never check for empty during `evaluate` call and fail miserably.

- Added checks for empty data in `Evaluators` factory, which return defaults in case of no data was provided
- DRY out `Evaluators` factory 
- Update `EvaluatorsTest` to use property based testing with Tables. Each evaluators is now being evaluated against non empty and an empty dataset.
- Fix evaluation metrics order in ModelInsights pretty print
- Camel case Brier Score from `brierScore` -> `BrierScore`

NA

**Additional context**
Follows up on PR #176

### DIFF
--- a/core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
+++ b/core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
@@ -1,0 +1,200 @@
+/*
+ * Copyright (c) 2017, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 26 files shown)


core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
```diff
@@ -0,0 +1,200 @@
...
+   * Convert metrics class to a map
+   * @return a map from metric name to metric value
+   */
+  def toMap: Map[String, Any] = JsonUtils.toMap(JsonUtils.toJsonTree(this))
+
+  /**
+   * Convert metrics into metadata for saving
+   * @return metadata
+   */
+  def toMetadata: Metadata = this.toMap.toMetadata
```
leahmcguire: I think we have a trait for this ...

core/src/main/scala/com/salesforce/op/evaluators/Evaluators.scala
```diff
@@ -85,8 +86,8 @@ object Evaluators {
         override def evaluate(dataset: Dataset[_]): Double = {
           import dataset.sparkSession.implicits._
           val dataUse = makeDataToUse(dataset, getLabelCol)
-          new MulticlassMetrics(dataUse.select(getPredictionValueCol, getLabelCol).as[(Double, Double)].rdd)
-            .precision(1.0)
+            .select(getPredictionValueCol, getLabelCol).as[(Double, Double)].rdd
+          if (dataUse.isEmpty()) 0.0 else new MulticlassMetrics(dataUse).precision(1.0)
```
leahmcguire: you pass this in in other places, why hardcode here?

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
```diff
@@ -122,24 +122,32 @@ private[op] class OpBinaryClassificationEvaluator
...
+    default: => Double
+  ): Double = {
     val labelColName = getLabelCol
     val dataUse = makeDataToUse(dataset, labelColName)
     new BinaryClassificationEvaluator()
       .setLabelCol(labelColName)
       .setRawPredictionCol(getRawPredictionCol)
       .setMetricName(metricName.sparkEntryName)
-      .evaluate(dataUse)
+      .evaluateOrDefault(dataUse, default = default)
```
leahmcguire: We should not change the interface - the default has a default value so why change the name??

core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
tovbinm: yes, this exactly what I am using here - enrichment for map - https://github.com/salesforce/TransmogrifAI/blob/master/utils/src/main/scala/com/salesforce/op/utils/spark/RichMetadata.scala#L204

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
tovbinm: it's an enrichment function for spark `Evaluator`

core/src/main/scala/com/salesforce/op/evaluators/Evaluators.scala
tovbinm: Fixed.

PR #179, opened 2018-11-08
Integrate Streaming Histogram into RawFeatureFilter.

marcovivero:
**Related issues**
No related issue, this is an enhancement.

Integrating StreamingHistogram with histogram generation which happens in RawFeatureFilter. Text distribution computation has also been separated from numeric distribution computations.


core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -64,3 +75,162 @@ case object Summary {
...
+    this
+  }
+}
+
+object TextSummary {
+  implicit val semigroup = new Semigroup[TextSummary] {
+    def plus(l: TextSummary, r: TextSummary): TextSummary = l.merge(r)
+  }
+}
+
```
sxd929: could you add some comments for the newly added classes, is this for all numerics? (compared to textSummary)?

core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -64,3 +75,162 @@ case object Summary {
     }
   }
 }
+
+class TextSummary(textFormula: TextSummary => Int) extends Serializable {
```
sxd929: I'm quite confused here, is old Summary class still in use? or it's completely replaced by the new ones (text summary and numeric summary)

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -371,6 +355,156 @@ object RawFeatureFilter {
...
+        summaries: Map[FeatureKey, HistogramSummary],
+        features: Map[FeatureKey, Seq[Double]]): Map[FeatureKey, HistogramSummary] =
+        summaries.keySet.union(features.keySet).map { key =>
+          val summary = summaries.get(key).getOrElse(new HistogramSummary(bins, bins * 10))
+          val points = features.get(key).getOrElse(Seq())
+
+          if (points.nonEmpty) key -> summary.update(points) else key -> summary
+        }.toMap
+
+      def updateTextSummaries(
```
sxd929: how would text bins formula be passed in to this?

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -182,95 +191,68 @@ class RawFeatureFilter[T]
...
+    toDrop.foreach { case (featureKey, (_, reasons)) =>
+      log.info(s"Dropping feature $featureKey because:\n\t${reasons.mkString("\n\t")}\n")
     }
-    logExcluded(
-      trainingNullLabelLeakers,
-      s"Features excluded because null indicator correlation (absolute) exceeded max allowed ($maxCorrelation)")
-
-    val scoringUnfilled =
-      if (scoringDistribs.nonEmpty) {
-        require(scoringDistribs.length == featureSize, "scoring and training features must match")
```
sxd929: what will happen if scoring Distribution is empty of feature size does not match now, is it checked somewhere?

core/src/main/scala/com/salesforce/op/stages/impl/selector/DefaultSelectorParams.scala
```diff
@@ -38,15 +38,15 @@ object DefaultSelectorParams {
   val MaxBin = Array(32) // bins for cont variables in trees - 32 is spark default
   val MinInstancesPerNode = Array(10, 100) // spark default 1
   val MinInfoGain = Array(0.001, 0.01, 0.1) // spark default 0
-  val Regularization = Array(0.001, 0.01, 0.1, 0.2) // spark default 0
+  val Regularization = Array(0.0001, 0.0015, 0.001, 0.01, 0.1) // spark default 0
```
sxd929: why did we make this change, are there any related experiments?

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -371,6 +355,156 @@ object RawFeatureFilter {
     bins
   }
 
+  private def getAllSummaries(bins: Int, allFeatures: RDD[AllFeatures]): AllSummaries = {
+    def apply(sum: AllSummaries, feat: AllFeatures): AllSummaries = {
+      val (totalCount, responseSummaries, numericSummaries, textSummaries) = sum
+      val (responseFeatures, numericFeatures, textFeatures) = feat
+
+      def updateNumericSummaries(
```
sxd929: this might be just my personal feelings, is this nested too much? e.g. three levels nested method

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -146,31 +150,36 @@ class RawFeatureFilter[T]
...
+      responseSummaries.map { case (key, sum) =>
+        key -> sum.getFeatureDistribution(key, totalCount, featureDistributionType)
+      }
+    val numericDistributions: Map[FeatureKey, FeatureDistribution] =
+      numericSummaries.map { case (key, sum) =>
+        key -> sum.getFeatureDistribution(key, totalCount, featureDistributionType)
+      }
+
+    // This will initialize existing text summary hashing TFs
+    textSummaries.foreach { case (_, textSum) => textSum.setHashingTF() }
```
kinfaikan: Perhaps call setHashingTF when creating a new TextSummary (i.e., in updateTextSummaries)?

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
kinfaikan: I think we should add textBinsFormula as input argument to both `getAllSummaries` and `updateTextSummaries`.

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -371,6 +355,156 @@ object RawFeatureFilter {
...
+        summaries: Map[FeatureKey, TextSummary],
+        features: Map[FeatureKey, Seq[String]]): Map[FeatureKey, TextSummary] =
+        summaries.keySet.union(features.keySet).map { key =>
+          val summary = summaries.get(key).getOrElse(new TextSummary(_ => bins))
+
+          features.get(key).map(text => key -> summary.update(text))
+            .getOrElse(key -> summary)
+        }.toMap
+
+        val newResponseSummaries = updateNumericSummaries(responseSummaries, responseFeatures)
```
kinfaikan: Indentation

core/src/main/scala/com/salesforce/op/filters/Summary.scala
kinfaikan: Summary class seems no longer needed.

core/src/main/scala/com/salesforce/op/stages/impl/selector/DefaultSelectorParams.scala
kinfaikan: The only related doc I can find: https://salesforce.quip.com/JhIiA1PLIYUY (streaming histogram results)

tovbinm:
@marcovivero are you not going to work on this or you are making a clean / smaller one?

Issue #154, opened 2018-10-08
Geolocation to location transformer


tovbinm:
**Problem**
We would like to be able to treat GeoLocation values as categoricals features. 

**Solution**
Add unary transformer to convert `Geolocation` values into `Country`.

**Alternatives**
N/A

ybh6336:
Is the idea to use a reverse geolocation service?

http://www.geonames.org/export/web-services.html#countrysubdiv
OR
https://developers.google.com/maps/documentation/geocoding/start?csw=1#ReverseGeocoding

tovbinm:
@ybh6336 the idea is to use some kind of a binary file containing the information of geolocation -> country code (@ybh6336 do you know of any good one?). But we would provide an interface for users to replace is with a remote service if needed.

ybh6336:
@tovbinm I stumbled across the following projects which do offline reverse geocoding. Do these look in line with what you were thinking?

https://github.com/AReallyGoodName/OfflineReverseGeocode (java)
https://github.com/thampiman/reverse-geocoder (python)

tovbinm:
Yes, but if possible simpler ;) For a start we would like to have a transformer to return a country code from a pair of coordinates.

devansh2001:
Hello! I'm just getting started with TransmogrifAI and I'd love to work on this issue. I see a couple of attempts marked 'Closed,' but this issue is still marked 'open.' Is this issue still accepting solutions?

tovbinm:
We closed those cause there were no use-case for us to support exploration of this direction further.

You can have a look of what I did https://github.com/salesforce/TransmogrifAI/pull/227. It was based on the idea of mapping each (lat,lon) coordinate to the nearest city, and then assuming the country of that city as the result. This approach would operate correctly on majority of coordinates, but still have some edge cases.

An alternative approach was to locate a encoded country borders map, convert it to convex polygons, then query which polygon contains each location. Unfortunately no good source of publicly accessible country borders was found.

devansh2001:
Got it, thanks so much! I'll get back to this if I find a reliable public service, and I'll look into other issues to fix or features to implement. :-)

PR #196, opened 2018-12-16
Possible solution for issue #154 (Geolocation to Country transformer)

ybh6336:
Uses reverse geolocation lookup code at https://github.com/AReallyGoodName/OfflineReverseGeocode

**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/154

The proposed solution, implemented in com.salesforce.op.stages.impl.feature.GeolocationToCountryTransformer, uses latitude and longitude of Geolocation to look up the country code. It uses an offline reverse geocode lookup code at https://github.com/AReallyGoodName/OfflineReverseGeocode. The database for offline lookup is in file core/src/main/resources/geolocation/cities1000.txt. This file has been downloaded from http://download.geonames.org/export/dump/.
The companion object, GeolocationToCountryTransformer, loads the file once, which is then referenced as a static field.

An alternative approach is to perform a reverse lookup using a remote service, like the ones listed below:
http://www.geonames.org/export/web-services.html#countrysubdiv
https://developers.google.com/maps/documentation/geocoding/start?csw=1#ReverseGeocoding
Drawbacks:
- needs API license
- latency
- would need a fallback offline approach

**Additional context**
None

### DIFF
--- a/core/src/main/java/geocode/GeoName.java
+++ b/core/src/main/java/geocode/GeoName.java
@@ -1,0 +1,116 @@
+/*
+The MIT License (MIT)
+[OSI Approved License]
+The MIT License (MIT)
+
+Copyright (c) 2014 Daniel Glasson
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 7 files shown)


ybh6336:
Unfortunately, https://github.com/AReallyGoodName/OfflineReverseGeocode has not been published to maven central, so had to copy code here (with copyright header and link to the original repository).

tovbinm:
@ybh6336 thank you for your contribution.

The dataset `cities1000.txt` is under `Creative Commons Attribution 4.0 License` (https://creativecommons.org/licenses/by/4.0). I need to verify with our legal department if it's ok to include or not.

ybh6336:
Thanks @tovbinm. I will look into `lucene-spatial3d`.

tovbinm:
Also have a look on this one - https://s2geometry.io/. They might have a java interface that we could use.

tovbinm:
@ybh6336 try this one using lucene. It's as simple as this: 1. index all cities 2. query then with radius around lat/lon - https://medium.com/@janakachathuranga/indexing-geographical-data-with-lucene-6-4ade78eac1ab

tovbinm:
@ybh6336 How are you doing?

I just got back an answer from our legal department. Unfortunately we wont be able to include this dataset in our repository. Unless we find another dataset with BSD-3 or Apache 2.0 license, create our own instead. 

Alternatively we can add instructions for users on how they can download and include the dataset themselves. Based on my previous experience, it would be cumbersome, error prone and would create bad user experience. So this is the least preferred option.

vpatryshev:
Is it legal then to include a link to this file, and load it on demand. In
this case it's not us, it's the user that loads the file, no?

Thanks,
-Vlad

tovbinm:
#227

Issue #198, opened 2018-12-17
The problem of Xgboost


jdlzy:
Has anyone encountered this problem?When I use xgboost to train the model !

the logs is:
```

2018-12-17 19:40:16 INFO  ApplicationMaster:54 - Final app status: FAILED, exitCode: 15, (reason: User class threw exception: java.lang.RuntimeException: Key 'trackerConf' has unsupported value TrackerConf(0,scala) of type ml.dmlc.xgboost4j.scala.spark.TrackerConf
	at com.salesforce.op.utils.spark.RichMetadata$RichMap$.com$salesforce$op$utils$spark$RichMetadata$RichMap$$unsupported$1(RichMetadata.scala:207)
	at com.salesforce.op.utils.spark.RichMetadata$RichMap$$anonfun$toMetadata$extension$1.apply(RichMetadata.scala:238)
	at com.salesforce.op.utils.spark.RichMetadata$RichMap$$anonfun$toMetadata$extension$1.apply(RichMetadata.scala:218)
	at scala.collection.TraversableOnce$$anonfun$foldLeft$1.apply(TraversableOnce.scala:157)
	at scala.collection.TraversableOnce$$anonfun$foldLeft$1.apply(TraversableOnce.scala:157)
	at scala.collection.immutable.HashMap$HashMap1.foreach(HashMap.scala:221)
	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:428)
	at scala.collection.TraversableOnce$class.foldLeft(TraversableOnce.scala:157)
	at scala.collection.AbstractTraversable.foldLeft(Traversable.scala:104)
	at com.salesforce.op.utils.spark.RichMetadata$RichMap$.toMetadata$extension(RichMetadata.scala:218)
	at com.salesforce.op.stages.impl.selector.ModelEvaluation.toMetadata(ModelSelectorSummary.scala:130)
	at com.salesforce.op.stages.impl.selector.ModelSelectorSummary$$anonfun$1.apply(ModelSelectorSummary.scala:90)
	at com.salesforce.op.stages.impl.selector.ModelSelectorSummary$$anonfun$1.apply(ModelSelectorSummary.scala:90)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
[...]

PR #199, opened 2018-12-18
Export model selector defaults + metadata fixes

tovbinm:
**Related issues**
1. It's impossible to reuse model selector default models & grids without copy pasting them from our code base.
2. Currently model selector fails to produce model selector summary metadata when one of the models contain a stage with unsupported params (example - #198). 

1. Expose model selector defaults on all model selectors (binary, multi, regression) to allow reuse.
2. Skip unsupported metadata values when producing model selector summary metadata.

N/A

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorFactory.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorFactory.scala
@@ -41,15 +41,23 @@ import org.apache.spark.ml.param.ParamMap
 private[op] trait ModelSelectorFactory {
 
   /**
-   * Subset of models to use
+   * Default model types and model parameters for problem type
    */
-  private[op] val modelNames: Seq[ModelsToTry]
+  trait ModelDefaults[M <: ModelsToTry] {
 
-  /**
-   * Default models and parameters
-   * @return defaults for problem type
-   */
-  protected def defaultModelsAndParams: Seq[(EstimatorType, Array[ParamMap])]
+    /**
+     * Subset of models to use in model selector
+     */
+    val modelTypesToUse: Seq[M]
+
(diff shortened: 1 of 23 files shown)


tovbinm:
@kinfaikan please review

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorFactory.scala
```diff
@@ -58,25 +66,33 @@ private[op] trait ModelSelectorFactory {
...
-        .filter{ case (e, p) => modelStrings.contains(e.getClass.getSimpleName) }
+    val modelTypeNames = modelTypesToUse.map(_.entryName).toSet
+    val modelsToUse = {
+      // if no models are specified use the defaults and filter by the named models to use
+      if (modelsAndParameters.isEmpty) {
+        modelDefaults.modelsAndParams.filter { case (e, p) => modelTypeNames(e.getClass.getSimpleName) }
+      }
+      // if models to use has been specified and the models have been specified - filter the models by the names
+      else if (
+        modelTypesToUse.distinct.sortBy(_.entryName) != modelDefaults.modelTypesToUse.distinct.sortBy(_.entryName)
```
kinfaikan: Why not simply use entire modelsAndParameters?

tovbinm: well, because it would yield incorrect results: `List(1,2) == List(2,1) // false`. That the reason I am dedupping and sorting here.

kinfaikan: I was thinking whether it made more sense to ignore modelTypes and modelDefaults when modelsAndParameters was not empty.

PR #206, opened 2019-01-10
Use ParamGridBuilder in model selector grids to allow modifications

tovbinm:
**Related issues**
It's difficult to reuse the model selector default grids, especially if one only wants to change a single param value.

We replace `Array[ParamMap]` with `ParamGridBuilder` to ease on that. For example user can now modify a single parameter `maxBins` like this and keep the rest of the grid the same:
```scala
val modelsAndParams = BinaryClassificationModelSelector.Defaults.modelsAndParams.map {
    case (e: OpRandomForestClassifier, grid) => e -> grid.addGrid(e.maxBins, Array(1, 2, 3)).build()
    case (e, grid) => e -> grid.build()
  }
val modelSelector = BinaryClassificationModelSelector.withCrossValidation(modelsAndParameters = modelsAndParams)
```

Many many others.

**Additional context**
This is still not over...

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
@@ -65,7 +65,7 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
      *
      * @return defaults for problem type
      */
-    def modelsAndParams: Seq[(EstimatorType, Array[ParamMap])] = {
+    def modelsAndParams: Seq[(EstimatorType, ParamGridBuilder)] = {
       val lr = new OpLogisticRegression()
       val lrParams = new ParamGridBuilder()
         .addGrid(lr.fitIntercept, DefaultSelectorParams.FitIntercept)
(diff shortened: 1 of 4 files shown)


leahmcguire:
@tovbinm this is not incompatible with doing random param search. This should be reverted and we should come up with a better interface for minor changes to the grid...
 https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilder.scala#L52****

tovbinm:
@leahmcguire oh, how so?

leahmcguire:
Because the RandomParamBuilder is not a ParamGridBuilder (the inheritance would not make sense since the addGrid method is meaningless). The commonality is that they both have build methods to produce an array of params.

tovbinm:
`addGrid` can be an empty method or perhaps user's can set a fixed grid with `addGrid` call?

another option to wrap `RandomParamBuilder` and `ParamGridBuilder` into a new type so we can pass either one of them.

leahmcguire:
ParamGridBuilder is a spark class - and it doesn't make any sense to have a grid in your random parameter search

leahmcguire:
Also this is a non-backwards compatible interface change.

Issue #205, opened 2019-01-10
Error in `transmogrifai gen` when field has an underscore


wdearden:
**Describe the bug**
I discovered this bug while using the Bootstrap Your First Project docs (https://docs.transmogrif.ai/en/stable/examples/Bootstrap-Your-First-Project.html) on a csv that I have. `transmogrifai gen` errors out when a field name has an underscore in it.

**To Reproduce**
Replace `Pclass` with `P_class` in `test-data/PassengerDataAll.avsc` and `Pclass` with `P_class` in `cli/passengers.answers`.
Run from the main `TransmogrifAI` directory:
```
./gradlew cli:shadowJar
alias transmogrifai="java -cp `pwd`/cli/build/libs/\* com.salesforce.op.cli.CLI"
transmogrifai gen --input `pwd`/test-data/PassengerDataAll.csv \
  --id passengerId --response survived \
  --schema `pwd`/test-data/PassengerDataAll.avsc \
  --answers cli/passengers.answers Titanic --overwrite

cd titanic
./gradlew compileTestScala installDist
```

**Expected behavior**
I expect the last line to compile without error.

**Logs or screenshots**
```
ch-nb-3908:titanic wdearden$ ./gradlew compileTestScala installDist

> Task :compileScala
Pruning sources from previous analysis, due to incompatible CompileSetup.
/Users/wdearden/repos/oss/TransmogrifAI/titanic/src/main/scala/com/salesforce/app/Features.scala:17: value getP_class is not a member of com.salesforce.app.schema.Passenger
    .extract(asPickList(_.getP_class))
                          ^
one error found

> Task :compileScala FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileScala'.
> Compilation failed

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 13s
3 actionable tasks: 3 executed
ch-nb-3908:titanic wdearden$ ./gradlew compileTestScala installDist

> Task :compileScala
Pruning sources from previous analysis, due to incompatible CompileSetup.
/Users/wdearden/repos/oss/TransmogrifAI/titanic/src/main/scala/com/salesforce/app/Features.scala:17: value getP_class is not a member of com.salesforce.app.schema.Passenger
    .extract(asPickList(_.getP_class))
                          ^
one error found

[...]

PR #208, opened 2019-01-13
Fixing code generation bug with underscores in names

vpatryshev:
**Related issues**
see [issue 205](https://github.com/salesforce/TransmogrifAI/issues/205)

Since avro converts names with underscores to CamelCase, we need to refer those CamelCase accessors as generated by avro. Test and test data added.

Could probably tweak avro code generation logic, but why.
We could also use CamelCase in our code, but this seems more confusing.

**Additional context**
Also, gradlew was tweaked to care about the cases when your jvm is not 8.0, and deprecated flags are not being used anymore.

### DIFF
--- a/cli/src/main/scala/com/salesforce/op/cli/CommandParser.scala
+++ b/cli/src/main/scala/com/salesforce/op/cli/CommandParser.scala
@@ -47,9 +47,12 @@ trait OpCli {
     else success
   }
 
+  private[cli] def tweak(file: File): File =
+    new File(file.getPath.replace("`pwd`/", ""))
+
   def fileExists(file: File): Either[String, Unit] = {
-    if (!file.exists()) failure(s"File '${file.getAbsolutePath}' not found")
-    else success
+    val f = tweak(file)
+    if (f.exists()) success else failure(s"File '${f.getAbsolutePath}' not found")
   }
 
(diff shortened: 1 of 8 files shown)


cli/src/main/scala/com/salesforce/op/cli/CommandParser.scala
```diff
@@ -47,9 +47,12 @@ trait OpCli {
     else success
   }
 
+  private[cli] def tweak(file: File): File =
+    new File(file.getPath.replace("`pwd`/", ""))
```
tovbinm: why do we need to hack around `pwd`?

vpatryshev:
Just makes life easier on some occasions.

Thanks,
-Vlad

cli/src/main/scala/com/salesforce/op/cli/CommandParser.scala
vpatryshev: Just to make developer's life easier on certain occasions.

vpatryshev: Makes life easier on some occasions. When one copy-pastes the path, for instance.

vpatryshev:
Can we merge it?

Issue #114, opened 2018-09-05
Add Helloworld to CI


tovbinm:
**Problem**
Currently [Helloworld](https://github.com/salesforce/TransmogrifAI/tree/master/helloworld) project is not CI verified and has to be tested manually prior to each release.

**Solution**
Add a CI build for Helloworld project.

**Alternatives**
NA

tovbinm:
thanks @rajdeepd for the contribution!

PR #210, opened 2019-01-21
CI build for HelloWorld

rajdeepd:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/114

Created a new circleci based configuration which checks for scalastyle and testcoverage

It will be easier for helloworld to have its own CI build.

**Additional context**
none


rajdeepd:
@tovbinm is this a bug in the test case? https://travis-ci.com/salesforce/TransmogrifAI/jobs/171776690

tovbinm:
@rajdeepd it's not running the helloworld build yet. The results seems to be from the regular build.

rajdeepd:
@tovbinm is there something else I need to add or this will suffice.

tovbinm:
@rajdeepd we have two options here:
1. configure helloworld as a separate build on Travis & Circle CI (this is what you started to do) by adding  build configs.
2. include helloworld build as part of the regular build (by modifying the build.gradle file in the root project). we can add a extra task that execute helloworld builds, which we will use on CI. E.g.:
```
./gradlew reportScoverage && ./gradlew publishToMavenLocal && ./gradlew buildHelloworld
```

helloworld/.circleci/config.yml
```diff
@@ -0,0 +1,91 @@
+version: 2
```
tovbinm: I dont think you need the `elloworld/.circleci/config.yml` file anymore

.circleci/config.yml
```diff
@@ -46,6 +46,25 @@ jobs:
           name: Clone repo & compile project
```
tovbinm: `name: Compile project`

.circleci/config.yml
```diff
@@ -46,6 +46,25 @@ jobs:
...
+            name: Scala style check for HelloWorld
+            command: cd helloworld && ./gradlew --parallel scalaStyle && cd ..
+  compile-hw:
+      <<: *machine-config
+      steps:
+        - checkout
+        - *open-jdk
+        - *restore-gradle-cache
+        - run:
+            name: Clone repo & compile project
```
tovbinm: `name: Compile Helloworld project`

.circleci/config.yml
```diff
@@ -46,6 +46,25 @@ jobs:
...
+        - *open-jdk
+        - run:
+            name: Scala style check for HelloWorld
+            command: cd helloworld && ./gradlew --parallel scalaStyle && cd ..
+  compile-hw:
+      <<: *machine-config
+      steps:
+        - checkout
+        - *open-jdk
+        - *restore-gradle-cache
```
tovbinm: I dont think you need to `*restore-gradle-cache` for Helloworld.

.circleci/config.yml
```diff
@@ -46,6 +46,25 @@ jobs:
...
+        - run:
+            name: Clone repo & compile project
+            command: cd helloworld && ./gradlew --parallel testScoverageClasses && cd ..
+        - *save-gradle-cache
```
tovbinm: also I dont think you want to save the cache for helloworld either

helloworld/.circleci/config.yml
rajdeepd: @tovbinm  i know that, will remove it

rajdeepd:
> @rajdeepd we have two options here:
> 
> 1. configure helloworld as a separate build on Travis & Circle CI (this is what you started to do) by adding  build configs.
> 2. include helloworld build as part of the regular build (by modifying the build.gradle file in the root project). we can add a extra task that execute helloworld builds, which we will use on CI. E.g.:
> 
> ```
> ./gradlew reportScoverage && ./gradlew publishToMavenLocal && ./gradlew buildHelloworld
> ```
> I think the second option is preferred, cause you would be able to test it locally and we wont need to replicate the CI build configs into helloworld project.
> 
> Does it make sense?

@tovbinm yes

.circleci/config.yml
rajdeepd: fixed

.circleci/config.yml
rajdeepd: fixed

.circleci/config.yml
rajdeepd: fixed

.circleci/config.yml
rajdeepd: fixed

rajdeepd:
@tovbinm sure!

tovbinm:
Thank you @rajdeepd Looking forward to it ;)

PR #213, opened 2019-02-04
Fix sorting in Prediction type for multiclass classification and add stronger tests

Jauntbox:
**Related issues**
None

Sorts keys by their final index - the actual index corresponding to the class, rather than the default string sort which jumbles classes together. Also adds many tests relevant to multiclass classification and threshold metrics.

N/A

**Additional context**
Previous threshold metrics (only those, not the metrics from Spark itself) were out of order when there were more that 10 classes due to this issue.


features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -317,7 +317,10 @@ class Prediction private[op](value: Map[String, Double]) extends RealMap(value)
     s"value map must only contain valid keys: '$PredictionName' or " +
       s"starting with '$RawPredictionName' or '$ProbabilityName'"
   )
-  private def keysStartsWith(name: String): Array[String] = value.keys.filter(_.startsWith(name)).toArray.sorted
+
+  // Need to make sure we sort the keys by their final index, which comes after an underscore in the apply function
+  private def keysStartsWith(name: String): Array[String] = value.keys.filter(_.startsWith(name)).toArray
+    .sortBy(s => s.substring(s.lastIndexOf("_") + 1).toInt)
```
tovbinm: `s.split('_').last.toInt`

core/src/test/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluatorTest.scala
```diff
@@ -129,6 +130,148 @@ class OpMultiClassificationEvaluatorTest extends FlatSpec with TestSparkContext
...
+      case (_, v) => v.head shouldBe 0L
+    }
+  }
+
+  // TODO: Add a way to robustly handle scores that tie (either many-way or complete ties)
+  ignore should "work on probability vectors where there are ties in probabilities" in {
+    val numClasses = 5
+    val numRows = 10
+    val vectors = Seq.fill[OPVector](numRows)(Array.fill(numClasses)(4.2).toOPVector)
+    println(s"vectors: $vectors")
```
tovbinm: please remove prints

tovbinm: or replace with log

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -317,7 +317,10 @@ class Prediction private[op](value: Map[String, Double]) extends RealMap(value)
     s"value map must only contain valid keys: '$PredictionName' or " +
       s"starting with '$RawPredictionName' or '$ProbabilityName'"
   )
-  private def keysStartsWith(name: String): Array[String] = value.keys.filter(_.startsWith(name)).toArray.sorted
+
+  // Need to make sure we sort the keys by their final index, which comes after an underscore in the apply function
+  private def keysStartsWith(name: String): Array[String] = value.keys.filter(_.startsWith(name)).toArray
+    .sortBy(_.split('_').last.toInt)
```
gerashegalov: depending how wide the keys are, but it would be more efficient to scan keys backwards to look for `_` with `lastIndexOf` to map to a `substring`. It would also produce less transient garbage: a single string instead of an array.

tovbinm: Too late for the party :)

PR #219, opened 2019-02-07
Limit label count query and in-memory footprint 

gerashegalov:
**Related issues**
In pathological datasets with a large number of labels DataCutter estimate is inefficient

Use a single label count query for potentially misconfigured training datasets with a long tail of low-cardinality labels. Use topK for n*log(k) instead of n*log(n) sort. Limit the number of labels materialized  for diagnostics in Java Heap

Not showing dropped labels :)


core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -150,6 +153,14 @@ private[impl] trait DataCutterParams extends Params {
 
   def getMaxLabelCategories: Int = $(maxLabelCategories)
 
+  final val maxDroppedLabelsRetained = new IntParam(this, "maxDroppedLabelsRetained",
+    "top labels that had to be dropped to protect against long tail of low cardinality labels",
+    ParamValidators.inRange(lowerBound = 2, upperBound = 1 << 30)
```
tovbinm: `lowerInclusive = false, upperInclusive = true`?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -88,9 +88,10 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
     val keep: Set[Double] =
       if (!isSet(labelsToKeep) || !isSet(labelsToDrop)) {
         val labels = data.map(r => r.getDouble(0) -> 1L)
-        val labelCounts = labels.groupBy(labels.columns(0)).sum(labels.columns(1)).persist()
```
tovbinm: no `persist` needed anymore?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -110,26 +111,28 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
-    val labelsKeep = labelCounts
+    val labelsOrderedByCount = labelCounts
       .filter(r => (r.getLong(1) / totalValues) >= minLabelFract)
       .sort(col(colCount).desc)
-      .take(maxLabels)
-      .map(_.getDouble(0))
-
-    val labelSet = labelsKeep.toSet
-    val labelsDropped = labelCounts.filter(r => !labelSet.contains(r.getDouble(0))).collect().map(_.getDouble(0)).toSet
+      .limit(maxLabels + maxDroppedLabelsRetained)
```
Jauntbox: Is the main improvement here that this sort/limit is "smart" in the sense that it knows not to sort the entire dataframe, but only find the top (maxLabels + maxDroppedLabelsRetained)?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -110,26 +111,28 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+      .collect()
 
+    val topLabels = labelsOrderedByCount.map(_.getDouble(0))
+    val labelSet = topLabels.take(maxLabels).toSet
+    val topLabelsDropped = topLabels.slice(maxLabels, maxLabels + maxDroppedLabelsRetained).toSet
     if (labelSet.nonEmpty) {
-      log.info(s"DataCutter is keeping labels: $labelSet and dropping labels: $labelsDropped")
+      log.info(s"DataCutter is keeping labels: $labelSet and dropping lower cardinality labels " +
```
Jauntbox: Since we're not outputting the entire dropped label set, we should probably include how many total labels were dropped for help debugging in the future. Is there a cheap way to do this other than `labelCounts.count()`?

gerashegalov:
Will be superceded by #263

PR #233, opened 2019-02-28
Tweaks to OpBinScoreEvaluator

shaeselix:
**Related issues**
N/A

Adding Lift, a new BinaryClassificationMetrics evaluator, to show how well scores predict positive labels in various score groupings / bands. As seen in the first plot of this article: https://www.kdnuggets.com/2016/03/lift-analysis-data-scientist-secret-weapon.html
A new parameter has been added to BinaryClassificationMetrics, LiftMetrics, that is filled with a Seq[LiftMetricBand], calculated with an RDD of scoreAndLabel in LiftEvaluator.

Rather than an evaluator, this could be an estimator. However, since it's a method of evaluation that summarizes a trained dataset, rather than estimating new values from data, IMO it belongs in evaluators. MultiClassificationMetrics has ThresholdMetrics, designed for a Confidence Plot, and this PR was emulated on that design.

**Additional context**


amateurhuman:
@shaeselix you should have an email invite to join the Salesforce org, or you can visit https://github.com/salesforce to accept. Once you've accepted, you can refresh the CLA check at https://cla.salesforce.com/status/salesforce/TransmogrifAI/pull/233

core/src/main/scala/com/salesforce/op/evaluators/LiftEvaluator.scala
```diff
@@ -0,0 +1,237 @@
...
+   * @param scoreAndLabels RDD[(Double, Double)] of BinaryClassification (score, label) tuples
+   * @param getScoreBands  function to calculate score bands, potentially using score distribution
+   * @return Seq of LiftMetricBand containers of Lift calculations
+   */
+  private[op] def liftMetricBands
+  (
+    scoreAndLabels: RDD[(Double, Double)],
+    getScoreBands: RDD[Double] => Seq[(Double, Double, String)]
+  ): Seq[LiftMetricBand] = {
+    val bands = getScoreBands(scoreAndLabels.map { case (score, _) => score })
```
tovbinm: `val scores = scoreAndLabels.map { case (score, _) => score }`

core/src/main/scala/com/salesforce/op/evaluators/LiftEvaluator.scala
```diff
@@ -0,0 +1,237 @@
...
+    val bandedLabels = scoreAndLabels.map { case (score, label) =>
+      (categorizeScoreIntoBand((score, bands)), label)
+    }.collect { case (Some(band), label) => (band, label) }
+    val perBandCounts = aggregateBandedLabels(bandedLabels)
+    val overallRate = overallLiftRate(perBandCounts)
+    bands.map({ case (lower, upper, band) =>
```
tovbinm: `bands.map { case (lower, upper, band) => ... }`

core/src/main/scala/com/salesforce/op/evaluators/LiftEvaluator.scala
```diff
@@ -0,0 +1,237 @@
...
+
+  /**
+   * function to return score bands for calculating lift
+   * Default: 10 equidistant bands for all 0.1 increments
+   * from 0.0 to 1.0
+   *
+   * @param scores RDD of scores. unused in this function
+   * @return sequence of (lowerBound, upperBound, bandString) tuples
+   */
+  private[op] def getDefaultScoreBands(scores: RDD[Double]):
```
tovbinm: redundant endline `private[op] def getDefaultScoreBands(scores: RDD[Double]): Seq[(Double, Double, String)] =`

core/src/main/scala/com/salesforce/op/evaluators/LiftEvaluator.scala
```diff
@@ -0,0 +1,237 @@
...
+   * @param scoreAndLabels RDD[(Double, Double)] of BinaryClassification (score, label) tuples
+   * @return Seq of LiftMetricBand containers of Lift calculations
+   */
+  def apply
+  (
+    scoreAndLabels: RDD[(Double, Double)]
+  ): Seq[LiftMetricBand] = {
+    liftMetricBands(
+      scoreAndLabels,
+      getDefaultScoreBands
```
tovbinm: so why not allow users to customize score bands?

core/src/main/scala/com/salesforce/op/evaluators/LiftEvaluator.scala
```diff
@@ -0,0 +1,237 @@
...
+import org.apache.spark.rdd.RDD
+
+/**
+ * Object to calculate Lift metrics for BinaryClassification problems
+ * Intended to build a Lift Plot.
+ *
+ * Algorithm for calculating a chart as seen here:
+ * https://www.kdnuggets.com/2016/03/lift-analysis-data-scientist-secret-weapon.html
+ */
+object LiftEvaluator {
```
tovbinm: The evaluator has to extend `OpBinaryClassificationEvaluatorBase`, `OpMultiClassificationEvaluatorBase` or `OpRegressionEvaluatorBase`

shaeselix:
@leahmcguire @tovbinm apologies for the extended reply to this. I was going through the BrierScore code and while they're calculating very similar things, I don't think they can be combined because the BrierScore calculations are bound by MinScore and MaxScore, and Lift Plots should be bound by 0.0 and 1.0. I was able to refactor it to extend `OpBinaryClassificationEvaluatorBase` and evaluate to a single value when a threshold for the score decision boundary is provided. Since F1 is also in `BinaryClassificationMetrics` and uses a threshold, I think it should be okay to also use one here.

leahmcguire:
@shaeselix - the reason the Brier score goes from min to max and not 0 to 1 is that not every classification model provides a probability as an output. This will fail if you try to run it on an SVM.

If that is really the difference in what you are trying to do then you should parameterize the Brier score so you can optionally set the min and max and throw an error if the data falls outside them.

shaeselix:
@leahmcguire @tovbinm okay, we figured out that, yes, this could all be expressed in the OpBinScoreEvaluator. Just made a few tweaks and updated tests. Also happy to contribute to the docs docs for this class, since it looks like they haven't been added yet.

tovbinm:
@shaeselix please do!! ;)

shaeselix:
@leahmcguire @tovbinm any issues here or can we merge?

core/src/main/scala/com/salesforce/op/evaluators/OpBinScoreEvaluator.scala
```diff
@@ -134,20 +143,30 @@ private[op] class OpBinScoreEvaluator
  * Metrics of BinaryClassificationBinMetrics
  *
  * @param BrierScore            brier score for overall dataset
+ * @param binSize               size of each bin
  * @param binCenters            center of each bin
  * @param numberOfDataPoints    total number of data points in each bin
+ * @param sumOfLabels           sum of the label in each bin
```
leahmcguire: or do you mean count? - then I request a name change :-)

core/src/main/scala/com/salesforce/op/evaluators/OpBinScoreEvaluator.scala
```diff
@@ -66,60 +67,68 @@ private[op] class OpBinScoreEvaluator
...
+        val diff = maxScore - minScore
+        val binCenters = for {i <- 0 until numOfBins}
+          yield minScore + ((diff * i) / numOfBins) + (diff / (2 * numOfBins))
+
+        val metrics = BinaryClassificationBinMetrics(
+          BrierScore = brierScoreSum / numberOfPoints,
+          binSize = diff / numOfBins,
+          binCenters = binCenters,
+          numberOfDataPoints = numberOfDataPoints,
+          sumOfLabels = sumOfLabels,
```
leahmcguire: why sum of labels, wouldn't count of labels be more useful?

core/src/main/scala/com/salesforce/op/evaluators/OpBinScoreEvaluator.scala
shaeselix: It's meant to be the count of the "yes" labels, calculated as the sum of the labels. Does that make sense?

leahmcguire: that makes sense for the name but then see my comment below. Why the sum of yes and not just the count of all (that combined with the average gives you the same info but in a more intuitive formate)

shaeselix: Isn't `numberOfDataPoints` the count of all?

leahmcguire: ah - good point :-) is what you want really the sum of the labels or the count of the positive labels? I dont think it is possible with the current models but say someone ran binary classification and fed in -1 and 1 instead of 0, 1 what do you want to see?

shaeselix: Yes, what I'm wanting is the count of the positive labels. This could be reconstructed by multiplying the `averageConversionRate` with the `numberOfDataPoints` but figured it would be cleaner and less subject to round off error by having a parameter for it explicitly.

I assumed your -1/1 example wasn't possible but I can refactor the code to make sure it's only counting positive labels rather than just summing. Would `numberOfPositiveLabels` be an appropriate param name?

shaeselix: Okay, fix is in! Thanks for the review!

tovbinm:
@shaeselix are you planning to address [this](https://github.com/salesforce/TransmogrifAI/pull/233#pullrequestreview-221424426) in a separate PR then?

shaeselix:
@tovbinm I'm happy to add it in another PR. I would say that BrierScore != Lift. My understanding of lift as a singular metric is that it requires a threshold for a decision boundary. That can be a parameter in the `Evaluators.BinaryClassification.lift(threshold)`, or it can simply be assumed to be 0.5, as it appears the `f1()`, `error()`, etc also assume.

PR #238, opened 2019-03-06
Random param builder for random hyperparameter search in model selectors

leahmcguire:
Made a new param builder that can be used with model selectors in place of spark param grid builder. This builder will generate params randomly within the specified ranges and distributions (which is superior to grid search)


tovbinm:
@leahmcguire please let us if this PR is ready for review

core/src/main/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilder.scala
```diff
@@ -0,0 +1,195 @@
...
+import com.salesforce.op.stages.impl.selector.Distribution.{Exponential, Subset, Uniform}
+import enumeratum._
+import org.apache.spark.ml.param._
+
+import scala.util.Random
+import scala.collection.mutable
+
+sealed abstract class Distribution extends EnumEntry with Serializable
+
+private object Distribution extends Enum[Distribution] {
```
tovbinm: don't make `Distribution` object private unless you have a very good reason.

core/src/main/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilder.scala
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * Adds double param with range of values in uniform distribution
+   *
+   * @param param parameter to add
+   * @param min   minimum value for param
+   * @param max   maximum value for param
+   */
+  def uniform(param: DoubleParam, min: Double, max: Double): this.type = {
+    assert(min < max, "min must be less than max")
```
tovbinm: use `require` instead of `assert`, as assertions can be skipped at runtime and requires cannot.

core/src/main/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilder.scala
```diff
@@ -0,0 +1,195 @@
...
+  val values: Seq[Distribution] = findValues
+  case object Uniform extends Distribution
+  case object Subset extends Distribution
+  case object Exponential extends Distribution // TODO inverse exponential?
+}
+
+/**
+ * Builder for a param sets used in random search-based model selection.
+ */
+class RandomParamBuilder {
```
tovbinm: consider allowing users to pass an instance of `Random` and/or seed

core/src/main/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilder.scala
```diff
@@ -0,0 +1,195 @@
...
+import scala.util.Random
+import scala.collection.mutable
+
+sealed abstract class Distribution extends EnumEntry with Serializable
+
+private object Distribution extends Enum[Distribution] {
+  val values: Seq[Distribution] = findValues
+  case object Uniform extends Distribution
+  case object Subset extends Distribution
+  case object Exponential extends Distribution // TODO inverse exponential?
```
tovbinm: are you planning to add `inverse exponential`?

core/src/test/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilderTest.scala
```diff
@@ -0,0 +1,103 @@
...
+
+@RunWith(classOf[JUnitRunner])
+class RandomParamBuilderTest extends FlatSpec with TestSparkContext {
+
+  private val lr = new OpLogisticRegression()
+  private val rf = new OpRandomForestClassifier()
+  private val xgb = new OpXGBoostClassifier()
+
+
+  it should "build a param grid of the desired length with one param variable" in {
```
tovbinm: `Spec[RandomParamBuilder] should ".."`

core/src/main/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilder.scala
leahmcguire: I cant decide, it is only for parameters that are bounded by 0 and 1 where you want to have more fine grained exploration close to 1...so not sure how useful it is

core/src/main/scala/com/salesforce/op/stages/impl/selector/RandomParamBuilder.scala
leahmcguire: it is only used inside the object - why would I make it public?

tovbinm:
@leahmcguire some scala style issues, otherwise lgtm

PR #241, opened 2019-03-12
Pivot with max cardinality percentage

michaelweilsalesforce:
**Related issues**
Some pivots do not necessary make sense when the cardinality is too high.

Set a max cardinality percentage for pivot in OneHotVectorizer, TextMapVectorizer and MultiPickListMap Vectorizer. Number of uniques is computed with HyperLogLog Monoid.

**Additional context**
Default value will be 1.00. Then Experiments will be run in order to determine a good default.
Max cardinality percentage is implemented for MultiPickList(Map) Vectorizers. this may be questionable.
Max  cardinality percentage is not in SmartText(Map)Vectorizer, because the param maxCardinality already exists, and it is used to differentiate a category from a text.


leahmcguire:
@mweilsalesforce can you please add a description of what this PR is doing?

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextMapPivotVectorizer.scala
```diff
@@ -52,16 +55,39 @@ class TextMapPivotVectorizer[T <: OPMap[String]]
...
+        val kryoSerializer = new KryoSerializer(new SparkConf()).newInstance()
+        k -> hll.toHLL(kryoSerializer.serialize(v).array())
+      })).reduce {
+        (a, b) => a.zip(b).map { case (m1, m2) => (m1 ++ m2).map {
+          case (k, v) => k -> (v + m1.getOrElse(k, hll.zero)) } }
+      }
+
+    val percentFilter = countUniques.flatMap(_.map{ case (k, v) =>
+      k -> (v.estimatedSize / n < $(maxPercentageCardinality))}.toSeq).toMap
+    val filteredDataset = filterHighCardinality(dataset, percentFilter)
```
leahmcguire: Rather than repeating this block of code can you make it a mixin trait?

michaelweilsalesforce:
BTW @tovbinm, couldn't make the Count Min Sketch work. Algebird has however a working CMSTopK Monoid, but we can't use it : we want topK based on minSupport.

core/src/main/scala/com/salesforce/op/stages/impl/feature/OpOneHotVectorizer.scala
```diff
@@ -69,15 +78,30 @@ abstract class OpOneHotVectorizer[T <: FeatureType]
...
+
+    val n = dataset.count()
+    val percentFilter = uniqueCounts.map(_.estimatedSize / n < $(maxPercentageCardinality))
     val rdd: RDD[Seq[Map[String, Int]]] = convertToSeqOfMaps(dataset)
+    val filteredRDD = rdd.map(_.zip(percentFilter).filter(_._2).map(_._1))
 
     val countOccurrences: Seq[Map[String, Int]] = {
-      if (rdd.isEmpty) Seq.empty[Map[String, Int]]
-      else rdd.reduce((a, b) => a.zip(b).map { case (m1, m2) => m1 + m2 })
+      if (filteredRDD.isEmpty) Seq.empty[Map[String, Int]]
```
tovbinm: instead of `isEmpty` and `reduce` use `fold`

tovbinm: also, I think this should be Seq.fill(inN.length)(Map.empty[String, Int]])` no?

core/src/main/scala/com/salesforce/op/stages/impl/feature/Transmogrifier.scala
```diff
@@ -73,6 +78,7 @@ private[op] trait TransmogrifierDefaults {
   val TrackInvalid: Boolean = false
   val TrackTextLen: Boolean = false
   val MinDocFrequency: Int = 0
+  val MaxPercentCardinality = 1.0
```
tovbinm: let's keep using the same const `val MaxPercentCardinality = OpOneHotVectorizer. MaxPercentCardinality`

core/src/main/scala/com/salesforce/op/stages/impl/feature/MultiPickListMapVectorizer.scala
```diff
@@ -63,9 +68,18 @@ class MultiPickListMapVectorizer[T <: OPMap[Set[String]]]
         k -> cats.map(_ -> 1L).groupBy(_._1).map { case (c, a) => c -> a.map(_._2).sum }
       }
     }
+    implicit val classTag: ClassTag[T#Value] = ReflectionUtils.classTagForWeakTypeTag[T#Value]
+    implicit val spark = dataset.sparkSession
+    implicit val kryo = new KryoSerializer(spark.sparkContext.getConf)
+    val uniqueCounts = countMapUniques(dataset, size = inN.length, bits = $(bits))
+    val n = dataset.count()
```
leahmcguire: can you calculate the total size when you get the unique counts and save a pass through the data?

core/src/main/scala/com/salesforce/op/stages/impl/feature/MultiPickListMapVectorizer.scala
```diff
@@ -63,9 +68,18 @@ class MultiPickListMapVectorizer[T <: OPMap[Set[String]]]
         k -> cats.map(_ -> 1L).groupBy(_._1).map { case (c, a) => c -> a.map(_._2).sum }
       }
     }
+    implicit val classTag: ClassTag[T#Value] = ReflectionUtils.classTagForWeakTypeTag[T#Value]
```
leahmcguire: can we put all this under a check that the max cardinality is < 1 so we dont do the extra calculation if we dont want to throw things out?

core/src/main/scala/com/salesforce/op/stages/impl/feature/OpOneHotVectorizer.scala
```diff
@@ -69,15 +77,26 @@ abstract class OpOneHotVectorizer[T <: FeatureType]
     val shouldCleanText = $(cleanText)
     val shouldTrackNulls = $(trackNulls)
 
+    implicit val classTag: ClassTag[T#Value] = ReflectionUtils.classTagForWeakTypeTag[T#Value]
```
leahmcguire: same here lets put this under a check that max card < 1

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextMapPivotVectorizer.scala
```diff
@@ -52,16 +55,28 @@ class TextMapPivotVectorizer[T <: OPMap[String]]
...
+    with MapStringPivotHelper with CleanTextMapFun with MinSupportParam with TrackNullsParam
+    with MaxPercentageCardinalityParams with CountUniqueMapFun {
+
 
   def fitFn(dataset: Dataset[Seq[T#Value]]): SequenceModel[T, OPVector] = {
     val shouldCleanKeys = $(cleanKeys)
     val shouldCleanValues = $(cleanText)
 
     def convertToMapOfMaps(mapIn: Map[String, String]): MapMap = mapIn.map { case (k, v) => k -> Map(v -> 1L) }
+    implicit val classTag: ClassTag[T#Value] = ReflectionUtils.classTagForWeakTypeTag[T#Value]
```
leahmcguire: again lets check before calculating

core/src/test/scala/com/salesforce/op/stages/impl/feature/OpSetVectorizerTest.scala
```diff
@@ -111,10 +114,11 @@ class OpSetVectorizerTest extends FlatSpec with TestSparkContext with AttributeA
...
     val result = transformed.collect(vector)
     val expected = Array(
-      Vectors.sparse(10, Array(0, 2, 5), Array(1.0, 1.0, 1.0)),
+      Vectors.sparse(10, Array(0, 1, 5), Array(1.0, 1.0, 1.0)),
+      Vectors.sparse(10, Array(0, 1, 5), Array(1.0, 1.0, 1.0)),
       Vectors.sparse(10, Array(0, 6, 7), Array(1.0, 1.0, 1.0)),
-      Vectors.sparse(10, Array(1, 5, 6), Array(1.0, 1.0, 1.0)),
-      Vectors.sparse(10, Array(0, 1, 7), Array(1.0, 1.0, 1.0))
+      Vectors.sparse(10, Array(2, 5, 6), Array(1.0, 1.0, 1.0)),
+      Vectors.sparse(10, Array(0, 2, 7), Array(1.0, 1.0, 1.0))
```
leahmcguire: why are the expected results changing for this test?

PR #243, opened 2019-03-15
fix local scoring with multipicklist features

kinfaikan:
**Related issues**
NA

Local scoring fails when there are multipicklist features because `Seq` can't be casted to `MWrappedArray`. Solution is to cast multipicklist value to `Seq`.

NA

**Additional context**
NA


local/src/test/scala/com/salesforce/op/local/OpWorkflowRunnerLocalTest.scala
```diff
@@ -117,4 +120,47 @@ class OpWorkflowRunnerLocalTest extends FlatSpec with PassengerSparkFixtureTest
     } score shouldBe expected
   }
 
+  it should "handle multi picklist features without throwing an exception" in {
```
tovbinm: the tests seems an overkill for such a simple check ;) let's think if we can make is simpler and not so labor intensive.

features/src/main/scala/com/salesforce/op/features/types/FeatureTypeSparkConverter.scala
```diff
@@ -196,7 +196,7 @@ case object FeatureTypeSparkConverter {
 
       // Sets
       case wt if wt <:< weakTypeOf[t.MultiPickList] => (value: Any) =>
-        if (value == null) FeatureTypeDefaults.MultiPickList.value else value.asInstanceOf[MWrappedArray[String]].toSet
+        if (value == null) FeatureTypeDefaults.MultiPickList.value else value.asInstanceOf[Seq[String]].toSet
```
tovbinm: you would also have to make the same fix for `MultiPickListMap`

local/src/test/scala/com/salesforce/op/local/OpWorkflowRunnerLocalTest.scala
tovbinm: Let's simply test it in `FeatureTypeSparkConverterTest`

Issue #245, opened 2019-03-15
Labels may not be dropped in MultiClassClassification


gerashegalov:
**Describe the bug**
Dropping the long tail of low cardinality labels is currently performed DataCutter.prepare. There are multiple call paths and we see it skipped in production. This results in performance degradation or even pipeline failures violating various constraints in MLlib such as 
- 100 label limit in Classifier
- "dense matrix too large"

**To Reproduce**
internal pipeline, failing to reproduce in unit tests yet

**Expected behavior**
an easier way to reason that DataCutter is not bypassed.

**Logs or screenshots**

**Additional context**

tovbinm:
@gerashegalov can we please also consider using HLL instead of Map when computing cardinality? @mweilsalesforce already started a related implementation on pivot with HLL #241 which you might be interested to look at.

gerashegalov:
I suggested approximate accounts on various occasions. Spark implements them on DF https://databricks.com/blog/2016/05/19/approximate-algorithms-in-apache-spark-hyperloglog-and-quantiles.html

michaelweilsalesforce:
@tovbinm @gerashegalov Indeed I implemented a HLL Aggregator for my vectorizer.

Trying Spark's sql function `approx_count_distinct` instead.

Issue I'm having is my input is a `Dataset[Seq[T#Value]]`. If I want to count each element I'll need to "explode" the Data. After doing that, `approx_count_distinct` can't be used because we are using kryo Encoder.

This is for the Vectorizer. For the MapVectorizer, I'll have to "explode" the `Dataset[Seq[Map[String, T]]]` which does not seem to be a good strategy.

michaelweilsalesforce:
For the issue of MultiClass labels, `approx_count_distinct` can be used though.

PR #246, opened 2019-03-15
Ensure data is `prepare`d even if there is no DAG

gerashegalov:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/245

WIP investigating

**Additional context**


core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
```diff
@@ -225,6 +225,16 @@ private[op] trait OpValidator[M <: Model[_], E <: Estimator[_]] extends Serializ
     datasetsByClass.map(_.toDF().rdd)
   }
 
+  protected def prepareWithoutDag(
+    splitter: Option[Splitter],
+    training: DataFrame,
+    validation: DataFrame
+  ): (DataFrame, DataFrame) = {
+    val trainingValidation = Vector(training, validation).flatMap(df =>
```
tovbinm: `val Array(prepTraining, prepValidation) = Array(training, validation)...`

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
```diff
@@ -241,18 +251,14 @@ private[op] trait OpValidator[M <: Model[_], E <: Estimator[_]] extends Serializ
...
-      .withColumn(ModelSelectorNames.idColName, monotonically_increasing_id())
-
-    val selectTest = newTest.select(label, features)
-      .withColumn(ModelSelectorNames.idColName, monotonically_increasing_id())
 
-    val (balancedTrain, balancedTest) = splitter.map(s => (
-      s.prepare(selectTrain).train,
-      s.prepare(selectTest).train)
-    ).getOrElse((selectTrain, selectTest))
+    val balancedTrainTest = Vector(newTrain, newTest)
```
tovbinm: same here

core/src/test/scala/com/salesforce/op/stages/impl/classification/MultiClassificationModelSelectorTest.scala
```diff
@@ -177,6 +178,38 @@ class MultiClassificationModelSelectorTest extends FlatSpec with TestSparkContex
...
+          Option(dataCutter),
+          numFolds = 4,
+          validationMetric = Evaluators.MultiClassification.precision(),
+          seed = 10L,
+          modelTypesToUse = Seq(MTT.OpRandomForestClassifier)
+        )
+        .setInput(label, features)
+
+    testEstimator.fit(labelsBeyondMaxIndexed)
+    assert(dataCutter.getLabelsToKeep.size === dataCutter.getMaxLabelCategories)
```
tovbinm:
```
withClue(<some useful message in case of error>) {
   dataCutter.getLabelsToKeep.size shouldBe dataCutter.getMaxLabelCategories
}
```

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -98,7 +98,7 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
     val dataUse = data.filter(r => keep.contains(r.getDouble(0)))
     val summary = DataCutterSummary(labelsKept = getLabelsToKeep, labelsDropped = getLabelsToDrop)
 
-    ModelData(dataUse, Some(summary))
+    ModelData(dataUse.persist(), Some(summary))
```
tovbinm: Are you sure dont persist somewhere upstream?

my 5c: my problem with persisting inside methods is that users of the methods wont know that they have to unpersist. So I prefer having methods where the input is persisted and unpersisted at the end of the same method. @leahmcguire what are your thoughts on this?

leahmcguire: It is not quite enough for some cases, however it is probably better to control from the workflow. In theory spark may have gotten better at knowing when to forget but I haven't tested it in a while.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
wsuchy: @tovbinm Are there any guidelines for coding style that should be used when contributing to Transmogrify describing details like this one?

tovbinm: Nope. We currently rely on scalastyle. I was considering to add a scalafmt to auto format the code, but I struggled with creating a good style. We might use this tool to autogen a config for us? https://github.com/tanishiking/scalaunfmt

leahmcguire:
@gerashegalov should we close this PR since we changed the way the Splitters work?

gerashegalov:
@leahmcguire yes, and it was not quite addressing the prod issue yet. I will open another PR once we have the fix confirmed

PR #249, opened 2019-03-22
MLeap for local model scoring

tovbinm:
**Related issues**
[Aardpfark](https://github.com/CODAIT/aardpfark) is no longer supported.

Switching to [MLeap](https://github.com/combust/mleap) runtime instead.

N/A

### DIFF
--- a/local/README.md
+++ b/local/README.md
@@ -1,26 +1,21 @@
 # TransmogrifAI Local
 
 This module enables local scoring with TransmogrifAI models without the need for a Spark session.
-Instead it uses [Aardpfark](https://github.com/CODAIT/aardpfark) and [Hadrian](https://github.com/opendatagroup/hadrian)
-runtime for JVM. It delivers unprecedented portability and performance of TransmogrifAI models
-allowing the serving of scores from any JVM process.
+Instead it uses [MLeap](https://github.com/combust/mleap) runtime on JVM. It delivers unprecedented portability and performance of TransmogrifAI models allowing the serving of scores from any JVM process.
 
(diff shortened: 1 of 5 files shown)


tovbinm:
@alexandrnikitin ^

local/README.md
```diff
@@ -44,6 +39,6 @@ val scoreFn = new OpWorkflowRunnerLocal(workflow).score(opParams)
 Below is an example of measured scoring performance on 6m records with 10 fields and 12 transformations applied.
 Executed on MacBook Pro i7 3.5Ghz in a single thread.
 ```
-Scored 6,000,000 records in 239s
-Average time per record: 0.0399215ms
+Scored 6,000,000 records in 710s
```
gerashegalov: interesting that there is 3x throughput difference

tovbinm: I had to warm up the JVM. Updated the numbers.

tovbinm: It's slightly faster.

local/src/main/scala/com/salesforce/op/local/OpWorkflowModelLocal.scala
```diff
@@ -57,29 +58,27 @@ trait OpWorkflowModelLocal extends Serializable {
    */
   implicit class RichOpWorkflowModel(model: OpWorkflowModel) {
 
-    private implicit val formats = DefaultFormats
-
     /**
-     * Internal PFA model representation
+     * Internal OP model representation
```
leahmcguire: MLeap?

PR #250, opened 2019-03-25
Add minimum rows for scoring set in RawFeatureFilter + rewrite tests to use data generators

Jauntbox:
**Related issues**
n/a

Adds `minRowsForScoringSet` to RawFeatureFilter so that features are not flagged for removal when the scoring set is extremely small and not representative. The threshold is currently set to be the same as the minimum training set size.

n/a

**Additional context**
This change broke many of the existing tests in `RawFeatureFilterTest` since they relied on fixed local datasets that fell well below the new threshold on scoring set size. All of the dataframe cleaning tests have been rewritten to use randomly generated data from the testkit generators so the data we perform the tests on is more robust.


core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -371,6 +373,10 @@ object RawFeatureFilter {
     bins
   }
 
+  // If there are not enough rows in the scoring set, we should not perform comparisons between the training and
+  // scoring sets since they will not be reliable. Currently, this is set to the same as the minimum training size.
+  val minRowsForScoringSet = 500
```
leahmcguire: lets not make this hard coded - it should be a parameter that the user can override

core/src/test/scala/com/salesforce/op/OpWorkflowTest.scala
```diff
@@ -241,14 +241,16 @@ class OpWorkflowTest extends FlatSpec with PassengerSparkFixtureTest {
...
-      .withRawFeatureFilter(Option(dataReader), Option(simpleReader),
-        maxFillRatioDiff = 1.0) // only height and the female key of maps should meet this criteria
+      .withRawFeatureFilter(Option(dataReader), None, maxFillRatioDiff = 1.0)
     val data = wf.computeDataUpTo(weight)
 
+    // Since there are < 500 rows in the scoring set, only the training set checks are applied here, and the only
+    // removal reasons should be null indicator - label correlations
     data.schema.fields.map(_.name).toSet shouldEqual
-      Set("key", "height", "survived", "stringMap", "numericMap", "booleanMap")
+      Set("booleanMap", "description", "height", "stringMap", "age", "key", "survived", "numericMap")
```
leahmcguire: once things are parameterized lets set them so that the tests remain the same

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
Jauntbox: Good point, fixed.

core/src/test/scala/com/salesforce/op/OpWorkflowTest.scala
Jauntbox: fixed

core/src/test/scala/com/salesforce/op/filters/RawFeatureFilterTest.scala
```diff
@@ -245,6 +269,559 @@ class RawFeatureFilterTest extends FlatSpec with PassengerSparkFixtureTest with
...
+    filteredRawData.featuresToDrop shouldBe empty
+    filteredRawData.mapKeysToDrop shouldBe empty
+    filteredRawData.cleanedData.schema.fields should contain theSameElementsAs
+      trainReader.generateDataFrame(features).schema.fields
+
+    assertFeatureDistributionEquality(filteredRawData, total = features.length * 2)
+  }
+
+  /**
+   * This test generates three numeric generators with the same underlying distribution, but different fill rates.
```
tovbinm: great test docs!!

core/src/test/scala/com/salesforce/op/OpWorkflowTest.scala
tovbinm: I was just about to suggest the same ;)

core/src/test/scala/com/salesforce/op/filters/RawFeatureFilterTest.scala
```diff
@@ -30,19 +30,37 @@
...
 import org.junit.runner.RunWith
 import org.scalatest.{Assertion, FlatSpec}
 import org.scalatest.junit.JUnitRunner
 
+import scala.reflect.runtime.universe.TypeTag
+
 @RunWith(classOf[JUnitRunner])
 class RawFeatureFilterTest extends FlatSpec with PassengerSparkFixtureTest with FiltersTestData {
 
+  // loggingLevel(Level.INFO)
```
tovbinm: please remove this line `// loggingLevel(Level.INFO)`

core/src/test/scala/com/salesforce/op/filters/RawFeatureFilterTest.scala
```diff
@@ -30,19 +30,37 @@
...
+
+  // Our randomly generated data will generate feature names and corresponding map keys in this universe
+  val featureUniverse = Set("myF1", "myF2", "myF3")
+  val mapKeyUniverse = Set("f1", "f2", "f3")
+  // Number of rows to use in randomly generated data sets
+  val numRows = 1000
```
tovbinm: can we drop this to 500?

PR #251, opened 2019-03-26
Fix so not preparing data twice when calling model selector fit method

leahmcguire:
**Related issues**
The prepare method on the spliters had unclear use - it was used both to measure the properties of the data and to resample data for cross validation / training split. This meant that the fit method of the model selectors had a bug where it would rebalance data before cross validation as well as during cross validation leading to potential label leakage. This was not generally a problem as within a workflow we called the findBestEstimator method rather than fit - but it would lead to bad results for anyone using the model selectors as a stand alone spark stage.

made the base splitter class have two methods which are called independently. one to measure the data and decide how to clean or rebalance it and one to actually do the cleaning or rebalancing

make prepare only estimate once - however this could lead to unexpected results

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -113,7 +113,7 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
   protected[op] def findBestEstimator(data: Dataset[_], dag: StagesDAG, persistEveryKStages: Int = 0)
     (implicit spark: SparkSession): Unit = {
 
-    splitter.map(_.examine(data.select(labelColName).toDF()))
+    splitter.foreach(_.preValidationPrepare(data.select(labelColName).toDF()))
     val theBestEstimator = validator.validate(modelInfo = modelsUse, dataset = data,
       label = in1.name, features = in2.name, dag = Option(dag), splitter = splitter,
       stratifyCondition = validator.isClassification
@@ -147,7 +147,7 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
       }
     require(!datasetWithID.isEmpty, "Dataset cannot be empty")
 
-    val splitterSummary = splitter.flatMap(_.examine(datasetWithID))
+    val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithID))
     val BestEstimator(name, estimator, summary) = bestEstimator.getOrElse {
       setInputSchema(dataset.schema).transformSchema(dataset.schema)
       val best = validator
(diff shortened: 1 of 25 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
```diff
@@ -74,7 +67,17 @@ abstract class Splitter(val uid: String) extends SplitterParams {
...
+
+
+  /**
+   * Function to use examine the data set to set parameters for preparation
+   * eg - do data balancing or dropping based on the labels
+   *
+   * @param data
+   * @return Parameters set in examining data
+   */
+  def examine(data: Dataset[Row]): Option[SplitterSummary]
```
gerashegalov: this function is designed to have a side-effects for prepare whereas examine sounds read-only. Can we call it something like `setupPrepare` or `prePrepare`?

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
```diff
@@ -113,6 +113,7 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
   protected[op] def findBestEstimator(data: Dataset[_], dag: StagesDAG, persistEveryKStages: Int = 0)
     (implicit spark: SparkSession): Unit = {
 
+    splitter.map(_.examine(data.select(labelColName).toDF()))
```
gerashegalov: since we don't use the result of transformation let us do `splitter.foreach(_.examine(...))`

tovbinm:
@leahmcguire please update PR title

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
leahmcguire: what about assesForPrepare?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataBalancer.scala
```diff
@@ -116,26 +116,39 @@ class DataBalancer(uid: String = UID[DataBalancer]) extends Splitter(uid = uid)
     }
   }
 
+
   /**
-   * Split into a training set and a test set and balance the training set
+   * Function to use examine the data set to set parameters for preparation
```
gerashegalov: Search/replace `examine`, sorry for the post-rename work.

michaelweilsalesforce:
Don't realize the bug, but it's been a while. I remember something like the first call of Splitter would "examine", but doesn't seem to be the case

leahmcguire:
@mweilsalesforce yes that it what I remember as well but it is not the case now - it is hard to unravel the history since this got moved to the public repo about the time this code went in.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
gerashegalov: same read-only connotation, I thin your current is `preValidationPrepare` is good

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataBalancer.scala
```diff
@@ -116,26 +116,39 @@ class DataBalancer(uid: String = UID[DataBalancer]) extends Splitter(uid = uid)
...
+
+  /**
+   * Rebalance the training data within the validation step
+   *
+   * @param data to prepare for model training. first column must be the label as a double
+   * @return balanced training set and a test set
+   */
+  def validationPrepare(data: Dataset[Row]): Dataset[Row] = {
+
+    if (summary.isEmpty) throw new RuntimeException("Cannot call prepare until examine has been called")
```
tovbinm: `examine` was renamed to `preValidationPrepare`? so please update the error message.

PR #252, opened 2019-03-26
Correctly handle error scenario and propagate error message up

tovbinm:
**Related issues**
Error message was not propagated to the user following up on #237 change

Propagate the error and return Try correctly

NA

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
@@ -156,32 +156,24 @@ class OpWorkflowModelReader(val workflow: OpWorkflow) extends MLReader[OpWorkflo
     }
   }
 
-  /**
-   * RawFeatureDistributions is now contained in and written / read through RawFeatureFilterResults.
-   * All setters of RawFeatureDistributions are now deprecated.
-   * This resolve function is to allow backwards compatibility where RawFeatureDistributions was a saved field
-   */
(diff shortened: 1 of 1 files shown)


clin-projects:
Thanks @tovbinm ! LGTM

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -156,32 +156,24 @@ class OpWorkflowModelReader(val workflow: OpWorkflow) extends MLReader[OpWorkflo
...
+       * RawFeatureDistributions is now contained in and written / read through RawFeatureFilterResults.
+       * All setters of RawFeatureDistributions are now deprecated.
+       * This resolve function is to allow backwards compatibility where RawFeatureDistributions was a saved field
+       */
+      val rawFeatureDistributionsEntryName = "rawFeatureDistributions"
+      if ((json \ rawFeatureDistributionsEntryName) != JNothing) {
+        val distString = (json \ rawFeatureDistributionsEntryName).extract[String]
+        FeatureDistribution.fromJson(distString).map(d => RawFeatureFilterResults(rawFeatureDistributions = d))
+      } else {
+        Success(RawFeatureFilterResults())
```
gerashegalov: Do we have a test for this issue?

clin-projects: We don't have a test for loading an old model with `rawFeatureDistributions` defined

I'll save a model with previous TMOG version in `src/test/resources/` and create a test

PR #254, opened 2019-03-28
Factoring out spark.gradle in helloworld

### LINKED ISSUE #87: Factor out all common spark/hadoop properties
**Problem**
We currently largely overlapping spark.gradle files especially in terms of spark properties. 
```
$ git ls-files | grep spark.gradle
gradle/spark.gradle
helloworld/gradle/spark.gradle
templates/simple/spark.gradle
```

**Solution**
Provide a way to have a single spark.gradle or at least a single spark-transmogriphai.conf file that with common properties that is passed via `--properties-file` to Spark.
 
**Alternatives**
- common properties file
- refactored spark.gradle 

**Additional context**
DRY

py-ranoid:
**Related issues**
Could resolve #87 

- `helloworld/gradle/spark.gradle` and `gradle/spark.gradle` are duplicates. 
- Hence I deleted `helloworld/gradle/spark.gradle`
- In `helloworld/build.gradle`, I replaced `'spark.gradle'` with `'${rootProject.projectDir}/../gradle/spark.gradle'`

Another alternative was to just use `'../gradle/spark.gradle'` but using `$rootProject.projectDir` replaces it with an absolute path.


py-ranoid:
Yes. Tried running the Titanic example after deleting `helloworld/gradle/spark.gradle`

helloworld/build.gradle
```diff
@@ -25,7 +25,7 @@ apply plugin: 'scalaStyle'
 apply plugin: 'com.commercehub.gradle.plugin.avro'
 
 apply from: 'gradle/tests.gradle'
-apply from: 'gradle/spark.gradle'
+apply from: "${rootProject.projectDir}/../gradle/spark.gradle"
```
gerashegalov: we can consult git for the local repo path to eliminate the need for `..`.
```
def gitRootOS = new ByteArrayOutputStream()
gitRootOS.withStream { os ->
    exec {
        commandLine 'git', 'rev-parse', '--show-toplevel'
        standardOutput = os
    }
}
def gitRoot = gitRootOS.toString().trim()
apply from: "${gitRoot}/gradle/spark.gradle"
```

py-ranoid: @gerashegalov We could try that. However, it could fail if one downloaded a zip of the repo or if we published it as a package down the line. Should we add this as well perhaps ?
```
if ( "${gitRoot}".compareTo("fatal: not a git repository (or any of the parent directories): .git") == 1 ) {
    gitRoot= "${rootProject.projectDir}/.."
    } 
```

gerashegalov: this is a good and valid point. Short term we can use your proposal (probably better use [ExecResult.getExitValue](https://docs.gradle.org/current/javadoc/org/gradle/process/ExecResult.html#getExitValue--) to catch all kinds of git invocation problems. 

Longer term, this hello world project can be configured by the same parent project.

py-ranoid: @gerashegalov So do you suggest checking for gitRoot in this PR?

gerashegalov: @py-ranoid yes, can we try what you do here https://github.com/salesforce/TransmogrifAI/pull/254#discussion_r270288085 but in a more general fashion. Instead of just checking the singl error message "fatal: not a git..." can we check for non-zero ExecResult.getExitValue which should capture all kinds of problems when invoking the git command?

wsuchy:
I think that `helloworld` should be either kicked out of this repo (we could create another one like `TransmogrifAI-hellowrold` or at least should be moved to the parent directory. Right now it is de facto a separate project having TMOG as a dependency.

tovbinm:
@wsuchy sounds reasonable. we have an sbt version out there already - https://github.com/salesforce/transmogrifai-helloworld-sbt

Perhaps we should combine it and just have two build supported in there - gradle and sbt.

py-ranoid:
@wsuchy @tovbinm We could create another repository for TransmogrifAI examples or add to the sbt repository. For example, [PyTorch](https://github.com/pytorch/pytorch) has a separate repo for [examples](https://github.com/pytorch/examples).

gerashegalov:
There are plenty of examples (no pun intended) where some example applications of a project are embedded in the same repo or a dedicated repo. Should we keep it here it should be a source dependency compiling against the master.

tovbinm:
It does not look this is going anywhere. Closing for now.

PR #255, opened 2019-03-28
Part 1 of 2 - Making concrete math transformers 

leahmcguire:
**Related issues**
Want to use concrete math transformer classes so they are easier to generate outside of the shortcuts

Move from lamda definitions of transformers to defined classes

**Additional context**
Next PR will add more math transformers


core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -0,0 +1,222 @@
...
+ * @tparam I      input feature type
+ * @tparam N      value type
+ */
+class ScalarAddTransformer[I <: OPNumeric[_], N]
+(
+  val scalar: N,
+  uid: String = UID[ScalarAddTransformer[_, _]]
+)(
+  implicit override val tti: TypeTag[I],
+  n: Numeric[N]
```
tovbinm: `val n: Numeric[N]` (here and in other transformers)

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -0,0 +1,222 @@
...
+ * Real(x)    + Real.empty = Real(x)
+ * Real(x)    + Real(y)    = Real(x + y)
+ */
+class AddTransformer[I1 <: OPNumeric[_], I2 <: OPNumeric[_]]
+(
+  uid: String = UID[AddTransformer[_, _]]
+)(
+  implicit override val tti1: TypeTag[I1],
+  override val tti2: TypeTag[I2]
+) extends BinaryTransformer[I1, I2, Real](operationName = "addition", uid = uid){
```
tovbinm: the operation name was `"plus"` is there a reason to change it (same with other transformers). I would rather keep it the same as before.

core/src/test/scala/com/salesforce/op/dsl/RichNumericFeatureTest.scala
```diff
@@ -134,15 +134,55 @@ class RichNumericFeatureTest extends FlatSpec with FeatureTestBase with RichNume
     // TODO: add vectorize() test
   }
 
-  Spec[RichRealNNFeature] should "have tests" in {
-    // TODO: add tests
+  Spec[RichRealNNFeature] should "perform math functions correctly" in {
```
tovbinm: thanks for the tests!

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -0,0 +1,222 @@
...
+class ScalarDivideTransformer[I <: OPNumeric[_], N]
+(
+  val scalar: N,
+  uid: String = UID[ScalarDivideTransformer[_, _]]
+)(
+  implicit override val tti: TypeTag[I],
+  n: Numeric[N]
+) extends UnaryTransformer[I, Real](operationName = "scalarDivide", uid = uid){
+  override def transformFn: I => Real = (i: I) => i.toDouble.map(_ / n.toDouble(scalar)).filter(Number.isValid).toReal
+}
```
tovbinm: you might need an endline here

tovbinm:
@leahmcguire can you also please add a test for each of the transformers using the transformer test spec?

core/src/test/scala/com/salesforce/op/dsl/RichNumericFeatureTest.scala
Jauntbox: Is there an easy way to use our `OpTransformerSpec` here, or would we have to make a separate test class for each transformer we're trying to test?

PR #256, opened 2019-03-29
fixed error message to match function names in DataCutter and DataBalancer

leahmcguire:


core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataBalancer.scala
```diff
@@ -144,7 +144,9 @@ class DataBalancer(uid: String = UID[DataBalancer]) extends Splitter(uid = uid)
    */
   def validationPrepare(data: Dataset[Row]): Dataset[Row] = {
 
-    if (summary.isEmpty) throw new RuntimeException("Cannot call prepare until examine has been called")
+    if (summary.isEmpty) throw new RuntimeException(
```
gerashegalov: nits:
we could put this in a method `checkPreconditions()` in the base class and call it here instead of repeating it in DC  and DB.

Also consider using `require` for a more compact syntax. 
```
require(summary.isDefined, "Cannot call validationPrepare until preValidationPrepare has been called")
```

PR #257, opened 2019-04-01
Math transformers part 2 of 2

leahmcguire:
**Related issues**
Made transformers and shortcuts for common math operations


core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -221,3 +221,174 @@ class ScalarDivideTransformer[I <: OPNumeric[_], N]
...
+ * @param uid      uid for instance
+ * @param tti      type tag for input
+ * @tparam I       input feature type
+ */
+class RoundTransformer[I <: OPNumeric[_]]
+(
+  uid: String = UID[FloorTransformer[_]]
+)(
+  implicit override val tti: TypeTag[I]
+) extends UnaryTransformer[I, Integral](operationName = "floor", uid = uid){
```
tovbinm: `operationName = "round"`

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -221,3 +221,174 @@ class ScalarDivideTransformer[I <: OPNumeric[_], N]
...
+    i.toDouble.map(logN).filter(Number.isValid).toReal
+  }
+}
+
+
+
+/**
+ * Power transformer
+ *
+ * @param base     base for log value
```
tovbinm: invalid doc

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -221,3 +221,174 @@ class ScalarDivideTransformer[I <: OPNumeric[_], N]
...
+ * @param uid      uid for instance
+ * @param tti      type tag for input
+ * @param n        value converter
+ * @tparam I       input feature type
+ * @tparam N       value type
+ */
+class PowerTransformer[I <: OPNumeric[_], N]
+(
+  val power: N,
```
tovbinm: let's add validation that `power` is within appropriate limits

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -221,3 +221,174 @@ class ScalarDivideTransformer[I <: OPNumeric[_], N]
...
+ * @param base     base for log value
+ * @param uid      uid for instance
+ * @param tti      type tag for input
+ * @param n        value converter
+ * @tparam I       input feature type
+ * @tparam N       value type
+ */
+class LogTransformer[I <: OPNumeric[_], N]
+(
+  val base: N,
```
tovbinm: lets add some validation on that `base` is within appropriate limits

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
tovbinm: I think `power` can be just `Double`

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
tovbinm: I think `base` can be just `Double`

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -221,3 +221,174 @@ class ScalarDivideTransformer[I <: OPNumeric[_], N]
...
+ * @param tti      type tag for input
+ * @tparam I       input feature type
+ */
+class RoundDigitsTransformer[I <: OPNumeric[_]]
+(
+  val digits: Int,
+  uid: String = UID[PowerTransformer[_, _]]
+)(
+  implicit override val tti: TypeTag[I]
+) extends UnaryTransformer[I, Real](operationName = "round", uid = uid){
```
tovbinm: `operationName = "roundDigits"` (so it will be different from round transformer

core/src/test/scala/com/salesforce/op/stages/impl/feature/AbsoluteValueTransformerTest.scala
```diff
@@ -0,0 +1,58 @@
...
+  val transformer: AbsoluteValueTransformer[Real] = new AbsoluteValueTransformer[Real]().setInput(f1)
+  override val expectedResult: Seq[Real] = Seq(Real(1.0), Real(4.0), Real.empty, Real(5.0),
+    Real(5.5), Real(0.1), Real(2.0), Real(0.0))
+
+  it should "have a working shortcut" in {
+    val f2 = f1.abs()
+    f2.originStage.isInstanceOf[AbsoluteValueTransformer[_]] shouldBe true
+  }
+}
+
```
tovbinm: please remove redundant trailing endlines in all test files

core/src/test/scala/com/salesforce/op/stages/impl/feature/AbsoluteValueTransformerTest.scala
```diff
@@ -0,0 +1,58 @@
...
+import com.salesforce.op.test.{OpTransformerSpec, TestFeatureBuilder}
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class AbsoluteValueTransformerTest extends OpTransformerSpec[Real, AbsoluteValueTransformer[Real]] {
+  val sample = Seq(Real(-1.0), Real(-4.0), Real.empty, Real(5.0), Real(-5.5), Real(0.1), Real(2.0), Real(0.0))
+  val (inputData, f1) = TestFeatureBuilder(sample)
+  val transformer: AbsoluteValueTransformer[Real] = new AbsoluteValueTransformer[Real]().setInput(f1)
+  override val expectedResult: Seq[Real] = Seq(Real(1.0), Real(4.0), Real.empty, Real(5.0),
```
tovbinm: nit: no need to `override` modifier, simply `val expectedResult: Seq[Real]`

core/src/test/scala/com/salesforce/op/stages/impl/feature/ExpTransformerTest.scala
```diff
@@ -0,0 +1,58 @@
...
+import com.salesforce.op.test.{OpTransformerSpec, TestFeatureBuilder}
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class ExpTransformerTest extends OpTransformerSpec[Real, ExpTransformer[Real]] {
+  val sample = Seq(Real(-1.3), Real(-4.9), Real.empty, Real(5.1), Real(-5.1), Real(0.1), Real(2.5), Real(0.4))
+  val (inputData, f1) = TestFeatureBuilder(sample)
+  val transformer: ExpTransformer[Real] = new ExpTransformer[Real]().setInput(f1)
+  override val expectedResult: Seq[Real] = Seq(Real(math.exp(-1.3)), Real(math.exp(-4.9)), Real.empty,
```
tovbinm: `Seq(..).map(math.exp).map(_.toReal)`

core/src/test/scala/com/salesforce/op/stages/impl/feature/PowerTransformerTest.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class PowerTransformerTest extends OpTransformerSpec[Real, PowerTransformer[Real, Double]] {
+  val sample = Seq(Real(-1.3), Real(-4.9), Real.empty, Real(5.1), Real(-5.1), Real(0.1), Real(2.5), Real(0.4))
+  val (inputData, f1) = TestFeatureBuilder(sample)
+  val transformer: PowerTransformer[Real, Double] = new PowerTransformer[Real, Double](3.0).setInput(f1)
+  override val expectedResult: Seq[Real] = Seq(Real(math.pow(-1.3, 3)), Real(math.pow(-4.9, 3)), Real.empty,
+    Real(math.pow(5.1, 3)), Real(math.pow(-5.1, 3)), Real(math.pow(0.1, 3)), Real(math.pow(2.5, 3)),
```
tovbinm: same here `Seq(...).map(math.pow).map(_.toReal)`

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
leahmcguire: and what are those limits?

tovbinm: for example - do we want to allow NaN? or do we want to allow negative values?

leahmcguire: negatives are valid...

core/src/main/scala/com/salesforce/op/stages/impl/feature/MathTransformers.scala
```diff
@@ -221,3 +221,173 @@ class ScalarDivideTransformer[I <: OPNumeric[_], N]
...
+    i.toDouble.map(logN).filter(Number.isValid).toReal
+  }
+}
+
+
+
+/**
+ * Power transformer
+ *
+ * @param power     base for log value
```
tovbinm: invalid doc

tovbinm:
@leahmcguire one thought I have after I merged this is that for features of more concrete numeric types, i.e. `RealNN`, `Currency` etc, all these shortcuts would "downgrade" them into their base type of `Real`, which might not  be desired. Perhaps we should make these transformers to be more specific so that the output type would match the input type, i.e:
```scala
class ExpTransformer[I <: Real]
(
  uid: String = UID[ExpTransformer[_]]
)(
  implicit override val tti: TypeTag[I]
) extends UnaryTransformer[I, I](operationName = "exp", uid = uid){
  private val convert = FeatureTypeSparkConverter[I]()
  override def transformFn: I => I = (i: I) => convert.fromSpark(i.toDouble.map(math.exp).filter(Number.isValid).orNull)
}
```

leahmcguire:
Well I'm not sure that the exp (or log or power) of currency is still currency. I'm sure percent is not still a percent. The RealNN is the only one that it may not be reasonable to downgrade and the difference in treatment between the Real and RealNN is just the null check and fill - which would have no effect if there are no nulls...

PR #258, opened 2019-04-02
Outputting Raw Feature Filter information: Part 2

clin-projects:
**Changes**
- Added serializing of RFF config and metrics in `ModelInsights`
  - RFF metrics serialized in FeatureInsights
  - RFF config is serialized alongside workflow stages, per discussion with @leahmcguire
- Added test for loading in old workflow (before RFF results was implemented)

This is related to earlier PR: https://github.com/salesforce/TransmogrifAI/pull/237


core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -406,6 +407,21 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
...
+          modelInsights.stageInfo("rawFeatureFilter") match {
+            case configInfo: Map[String, Map[String, Any]] => configInfo.getOrElse("params", Map.empty[String, Any])
+            case _ => Map.empty[String, Any]
+          }
+        }
+
+        (getRawFeatureFilterConfig(insights), getRawFeatureFilterConfig(deser)) match {
+          case (paramsMapI, paramsMapD) =>
+            paramsMapI.keys shouldEqual paramsMapD.keys
+            paramsMapI.keys.foreach { `key` => paramsMapI(key) shouldEqual paramsMapD(key) }
```
clin-projects: @leahmcguire @Jauntbox 

Trying to serialize / deserialize RFF config with json4s as part of `stageInfo` in `ModelInsights`

But currently running into issue with trying to serialize `Double.PositiveInfinity`.

At this line, I'm getting error:

> `org.scalatest.exceptions.TestFailedException: Infinity did not equal "Infinity"`

Any thoughts?

leahmcguire: If you look at the way the stage parameters are saved they should be converted to strings
https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/ModelInsights.scala#L679

tovbinm: `SpecialDoubleSerializer` is your friend

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -622,10 +626,11 @@ case object ModelInsights {
           val ftype = allFeatures.find(_.name == fname)
             .map(_.typeName)
             .getOrElse("")
+          val metrics = rawFeatureFilterResults.rawFeatureFilterMetrics.filter(_.name == fname)
           val distributions = rawFeatureFilterResults.rawFeatureDistributions.filter(_.name == fname)
           val exclusionReasons = rawFeatureFilterResults.exclusionReasons.filter(_.name == fname)
```
leahmcguire: Do we really want to flatten all these out? would it make sense to leave them nested under rawFeatureFilterResults?

tovbinm: This is WIP. I synced with @clin-projects today and he is planning to serialize the whole `RawFeatureFilterResults` instance.

clin-projects: There are at least two options here for how we pass the information in RawFeatureFilterResults (config, metrics, distributions, exclusion reasons) out of ModelInsights:

**Option 1. (Current implementation)**:
- Bundle raw feature filter configuration with `stageInfo`- Bundle feature-specific results (metrics, distributions, exclusionReasons) with `FeatureInsights`


**Option 2. (Alternative)**: ModelInsights does nothing with RFF results and outputs as separate object; but would have to re-think what FeatureInsights is doing. E.g., shouldn't be outputting FeatureDistribution (which is part of RFF results), etc.

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
clin-projects: I've updated to now convert to strings

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilterResults.scala
```diff
@@ -88,7 +93,33 @@ case class RawFeatureFilterConfig
...
-)
+) {
+
+  /**
+   * Converts case class constructor to a Map; values converted to String
+   *
+   * @return Map[String, String]
+   */
+  def toStringMap(): Map[String, String] = {
+    (Map[String, String]() /: this.getClass.getDeclaredFields) { (key, value) =>
```
tovbinm: oh please, let's not build our own to Map / from Map serializations. we already have a ton of json stuff. just use json4s to serialize the classes. if you need then json converted to map - it's also easily possible.

core/src/main/scala/com/salesforce/op/ModelInsights.scala
tovbinm: I think the best approach is Option 2, i.e.
```scala
case class ModelInsights
(
  label: LabelSummary,
  features: Seq[FeatureInsights],
  selectedModelInfo: Option[ModelSelectorSummary],
  trainingParams: OpParams,
  rawFeatureFilterResults: Option[RawFeatureFilterResults],
  stageInfo: Map[String, Any]
)
```

clin-projects: Changed to Option 2

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -346,9 +346,15 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest {
             i.featureName shouldEqual o.featureName
             i.featureType shouldEqual o.featureType
             i.derivedFeatures.zip(o.derivedFeatures).foreach{ case (ii, io) => ii.corr shouldEqual io.corr }
-            RawFeatureFilterResultsComparison.compareSeqDistributions(i.distributions, o.distributions)
-            RawFeatureFilterResultsComparison.compareSeqExclusionReasons(i.exclusionReasons, o.exclusionReasons)
         }
+        RawFeatureFilterResultsComparison.compareSeqDistributions(
```
clin-projects: **Note**: if there is some Double.NaN present here, while it may get correctly serialized and deserialized, this test will erroneously fail because it will not consider them equal

We can try to create some cases to correctly return True for NaN in this test, but I don't think this is a practical issue

tovbinm: NaN value equality checks should have been handled by this implicit - https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala#L57

did it not work for your test?

clin-projects: Ah, didn't see those implicits before. Just tested it and not sure why they aren't being picked up by the equality test.

A test with
```scala
Some(Double.NaN) shouldEqual Some(Double.NaN)
```
will fail with error

`org.scalatest.exceptions.TestFailedException: Some(NaN) did not equal Some(NaN)`

But the logic inside the equality check is correct: a direct test of 

```scala
areEqual(Some(Double.NaN), Some(Double.NaN))
```
returns true

core/src/main/scala/com/salesforce/op/ModelInsights.scala
leahmcguire: Ahh, too fast to respond. I'm not sure I agree that option 2 is the best format. Maybe we should all hop on a hangout and discuss?

tovbinm: How about a quick call on Monday?

clin-projects:
@tovbinm @leahmcguire Ready for review

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilterResults.scala
```diff
@@ -86,10 +93,37 @@ case class RawFeatureFilterConfig
...
+
+  /**
+   * Summarize RawFeatureFilterConfig in format of stageInfo; this info will be passed alongside stage info in
+   * ModelInsights
+   *
+   * @return Map[String, Map[String, Any] ]
+   */
+  def toStageInfo(config: RawFeatureFilterConfig): Map[String, Map[String, Any]] = {
+    val stageName = "rawFeatureFilter"
+    val uid = "rawFeatureFilter"
```
tovbinm: This seems a bit of a hack to me, since it’s not a valid UID

tovbinm: Neither it’s a valid stage

tovbinm: And I still don’t understand why is it kept in stage info? :)

PR #260, opened 2019-04-03
Macro paradise plugin, macros for lambdas, and simple changes

wsuchy:
In order to serialize lambdas we need to have them in a separate object. Let's add some sugar around lambda expression that will do the job for us.
More description coming soon.


wsuchy:
Pinging @tovbinm

tovbinm:
@wsuchy are you planning to rework this to reflect the recent changes in the master branch?

PR #261, opened 2019-04-03
Correct conversion of Spark model stages into MLeap local models

tovbinm:
**Related issues**
1. Previously the conversion of Spark model stages into MLeap local models was done without providing all the necessary metadata which resulted in exceptions. For example when converting Spark `StringIndexerModel` into MLeap model it was expecting `ml_attr` metadata to be present in transformed Dataframe.
2. Reflecting `apply` method on MLeap models did not work correctly, since many models had more than one `apply` method present.

1. Provide transformed Dataframe with all the necessary metadata to allow MLeap local models creation.
2. Get rid of reflection for `apply` method on MLeap models and explicitly convert MLeap models into scoring methods.
3. Added tests

N/A


local/src/main/scala/com/salesforce/op/local/MLeapModelConverter.scala
```diff
@@ -0,0 +1,92 @@
...
+    case m: StopWordsRemoverModel => x => m.apply(x(0).asInstanceOf[Seq[String]])
+    case m: StringIndexerModel => x => m.apply(x(0))
+    case m: StringMapModel => x => m.apply(x(0).toString)
+    case m: TokenizerModel => x => m.apply(x(0).toString)
+    case m: VectorAssemblerModel => x => m.apply(x(0).asInstanceOf[Seq[Any]])
+    case m: VectorIndexerModel => x => m.apply(x(0).asInstanceOf[Vector])
+    case m: VectorSlicerModel => x => m.apply(x(0).asInstanceOf[Vector])
+    case m: WordLengthFilterModel => x => m.apply(x(0).asInstanceOf[Seq[String]])
+    case m: WordToVectorModel => x => m.apply(x(0).asInstanceOf[Seq[String]])
+    case m => throw new RuntimeException(s"Unsupported MLeap model: ${m.getClass.getName}")
```
leahmcguire: so every wrapped spark stage has to be in this list? we should add that the the docs on wrapping...

leahmcguire:
So now we have to spool up spark to use this sparkless scoring?? Could you get the necessary info another way? eg. serialize the dataframe schema along with the model?

tovbinm:
@leahmcguire we used to do before when loading spark stages. Now I just explicitly exposed an ability to users to control spark session lifecycle.

local/src/main/scala/com/salesforce/op/local/MLeapModelConverter.scala
tovbinm: I currently added all the stages from [features package](https://github.com/combust/mleap/tree/master/mleap-core/src/main/scala/ml/combust/mleap/core/feature). We can also add models from [classification](https://github.com/combust/mleap/tree/master/mleap-core/src/main/scala/ml/combust/mleap/core/classification), [regression](https://github.com/combust/mleap/tree/master/mleap-core/src/main/scala/ml/combust/mleap/core/regression) and [recommendation](https://github.com/combust/mleap/tree/master/mleap-core/src/main/scala/ml/combust/mleap/core/recommendation) packages, but we already have the first two of them covered as our own `OpTransformer` stages, so I did not see much of a point adding them.

tovbinm: so to your question - for right now I think we have everything covered, except recommenders, which I am planning to add once we are ready.

leahmcguire: Can you add a todo with the classification and regression models? I dont know that this will be much use without them...

tovbinm: Adding those is very easy. the thing is we already have classification and regression models as OpTransformers so MLeap won’t be used to run them.

leahmcguire: good point :-)

PR #263, opened 2019-04-03
DataCutter-related fixes for multiclass

### LINKED ISSUE #245: Labels may not be dropped in MultiClassClassification

gerashegalov:
**Related issues**
Fixes #245

- label sort by count is nondeterministic for labels with the same count
- data cutter label trimming in validation is not called prior to bestEstimator
- column  metadata is not updated after trimming
- in `DataCutter.estimate` whereas `filter` before `sort` is good for performance it may not drop labels non-deterministically unlike filtering over a stable order. 
  
Remove non-determinism

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -156,7 +156,6 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
     }
 
     val preparedData = splitter.map(_.validationPrepare(datasetWithID)).getOrElse(datasetWithID)
-
     val bestModel = estimator.fit(preparedData).asInstanceOf[M]
     val bestEst = bestModel.parent
     log.info(s"Selected model : ${bestEst.getClass.getSimpleName}")
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -202,9 +202,9 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
     val labelColIdx = data.columns.indexOf(labelColName)
     val needColSwapForValidation = labelColIdx > 0
     val df = if (needColSwapForValidation) {
-      data
-    } else {
       data.select(labelColName, data.columns.drop(labelColIdx): _*)
+    } else {
+      data
     }
     df.show()
 
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -48,7 +48,7 @@ import com.salesforce.op.utils.stages.FitStagesUtil._
 import org.apache.spark.ml.param._
 import org.apache.spark.ml.{Estimator, Model, PredictionModel}
 import org.apache.spark.sql.functions._
-import org.apache.spark.sql.{Dataset, SparkSession}
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
 
 import scala.reflect.runtime.universe._
 
@@ -109,11 +109,10 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
    * @param spark               Spark Session
    * @return Updated Model Selector with best model along with best paramMap
    */
-  protected[op] def findBestEstimator(data: Dataset[_], dag: StagesDAG, persistEveryKStages: Int = 0)
+  protected[op] def findBestEstimator(data: DataFrame, dag: StagesDAG, persistEveryKStages: Int = 0)
     (implicit spark: SparkSession): Unit = {
-
-    splitter.foreach(_.preValidationPrepare(data.select(labelColName).toDF()))
-    val theBestEstimator = validator.validate(modelInfo = modelsUse, dataset = data,
+    val (_, dataSet) = prepareForValidation(data)
+    val theBestEstimator = validator.validate(modelInfo = modelsUse, dataset = dataSet,
       label = in1.name, features = in2.name, dag = Option(dag), splitter = splitter,
       stratifyCondition = validator.isClassification
     )
@@ -146,11 +145,7 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
       }
     require(!datasetWithIDPre.isEmpty, "Dataset cannot be empty")
 
-    val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithIDPre))
-    val datasetWithID = splitterSummary.collect { case dc: DataCutterSummary =>
-      dc.preparedDF.getOrElse(datasetWithIDPre)
-    }.getOrElse(datasetWithIDPre)
-
+    val (splitterSummary, datasetWithID: DataFrame) = prepareForValidation(datasetWithIDPre)
     val BestEstimator(name, estimator, summary) = bestEstimator.getOrElse {
       setInputSchema(dataset.schema).transformSchema(dataset.schema)
       val best = validator
@@ -200,6 +195,32 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
       .setEvaluators(evaluators)
   }
 
+  private def prepareForValidation(data: DataFrame) = {
+    println("GERA DEBUG: " + data.schema)
+
+    // ensure label is the first column
+    val labelColIdx = data.columns.indexOf(labelColName)
+    val needColSwapForValidation = labelColIdx > 0
+    val df = if (needColSwapForValidation) {
+      data
+    } else {
+      data.select(labelColName, data.columns.drop(labelColIdx): _*)
+    }
+    df.show()
+
+    val splitterSummary = splitter.flatMap(_.preValidationPrepare(df))
+    val dataFrame = splitterSummary.collect { case dc: DataCutterSummary =>
+      dc.preparedDF.getOrElse(df)
+    }.getOrElse(df)
+    (splitterSummary,
+      if (needColSwapForValidation) {
+        // restore column order
+        dataFrame.select(data.columns.head, data.columns.tail: _*) // TODO figure out a better way
+      } else {
+        dataFrame
+      }
+    )
+  }
 }
 
 /**
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -196,30 +196,12 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
   }
 
   private def prepareForValidation(data: DataFrame) = {
-    println("GERA DEBUG: " + data.schema)
+    val splitterSummary = splitter.flatMap(_.preValidationPrepare(data, Some(labelColName)))
+    val resultDF = splitterSummary.collect { case dc: DataCutterSummary =>
+      dc.preparedDF.getOrElse(data)
+    }.getOrElse(data)
 
-    // ensure label is the first column
-    val labelColIdx = data.columns.indexOf(labelColName)
-    val needColSwapForValidation = labelColIdx > 0
-    val df = if (needColSwapForValidation) {
-      data.select(labelColName, data.columns.drop(labelColIdx): _*)
-    } else {
-      data
-    }
-    df.show()
-
-    val splitterSummary = splitter.flatMap(_.preValidationPrepare(df))
-    val dataFrame = splitterSummary.collect { case dc: DataCutterSummary =>
-      dc.preparedDF.getOrElse(df)
-    }.getOrElse(df)
-    (splitterSummary,
-      if (needColSwapForValidation) {
-        // restore column order
-        dataFrame.select(data.columns.head, data.columns.tail: _*) // TODO figure out a better way
-      } else {
-        dataFrame
-      }
-    )
+    splitterSummary -> resultDF
   }
 }
 
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -148,9 +148,6 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
 
     val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithIDPre))
     val datasetWithID = splitter.collect {
-      // TODO validate in best estimator below cannot defer label trimming to blanket
-      //  validationPrepare, redesign to accommodate for special case
-      //
       case dc: DataCutter => dc.validationPrepare(datasetWithIDPre)
     }.getOrElse(datasetWithIDPre)
     val BestEstimator(name, estimator, summary) = bestEstimator.getOrElse {
@@ -161,7 +158,11 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
       best
     }
 
-    val preparedData = splitter.map(_.validationPrepare(datasetWithID)).getOrElse(datasetWithID)
+    val preparedData = splitter.collect {
+      case _: DataCutter => datasetWithID
+      case sp => sp.validationPrepare(datasetWithID)
+    }.getOrElse(datasetWithID)
+
     val bestModel = estimator.fit(preparedData).asInstanceOf[M]
     val bestEst = bestModel.parent
     log.info(s"Selected model : ${bestEst.getClass.getSimpleName}")
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -37,6 +37,7 @@ import com.salesforce.op.readers.DataFrameFieldNames
 import com.salesforce.op.stages._
 import com.salesforce.op.stages.base.binary.OpTransformer2
 import com.salesforce.op.stages.impl.CheckIsResponseValues
+import com.salesforce.op.stages.impl.selector.ModelSelectorNames.ModelType
 import com.salesforce.op.stages.impl.tuning._
 import com.salesforce.op.stages.sparkwrappers.generic.SparkWrapperParams
 import com.salesforce.op.stages.sparkwrappers.specific.{OpPredictorWrapperModel, SparkModelConverter}
@@ -44,14 +45,12 @@ import com.salesforce.op.utils.spark.RichDataset._
 import com.salesforce.op.utils.spark.RichMetadata._
 import com.salesforce.op.utils.spark.RichParamMap._
 import com.salesforce.op.utils.stages.FitStagesUtil._
-import com.salesforce.op.stages.impl.selector.ModelSelectorNames.{EstimatorType, ModelType}
 import org.apache.spark.ml.param._
-import org.apache.spark.ml.{Estimator, Model, PipelineStage, PredictionModel}
+import org.apache.spark.ml.{Estimator, Model, PredictionModel}
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.{Dataset, SparkSession}
 
 import scala.reflect.runtime.universe._
-import scala.util.Try
 
 
 /**
@@ -138,16 +137,22 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
 
     implicit val spark = dataset.sparkSession
 
-    val datasetWithID =
+    val datasetWithIDPre =
       if (dataset.columns.contains(DataFrameFieldNames.KeyFieldName)) {
         dataset.select(in1.name, in2.name, DataFrameFieldNames.KeyFieldName)
       } else {
         dataset.select(in1.name, in2.name)
           .withColumn(ModelSelectorNames.idColName, monotonically_increasing_id())
       }
-    require(!datasetWithID.isEmpty, "Dataset cannot be empty")
+    require(!datasetWithIDPre.isEmpty, "Dataset cannot be empty")
 
-    val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithID))
+    val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithIDPre))
+    val datasetWithID = splitter.collect {
+      // TODO validate in best estimator below cannot defer label trimming to blanket
+      //  validationPrepare, redesign to accommodate for special case
+      //
+      case dc: DataCutter => dc.validationPrepare(datasetWithIDPre)
+    }.getOrElse(datasetWithIDPre)
     val BestEstimator(name, estimator, summary) = bestEstimator.getOrElse {
       setInputSchema(dataset.schema).transformSchema(dataset.schema)
       val best = validator
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
@@ -147,9 +147,10 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
     require(!datasetWithIDPre.isEmpty, "Dataset cannot be empty")
 
     val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithIDPre))
-    val datasetWithID = splitter.collect {
-      case dc: DataCutter => dc.validationPrepare(datasetWithIDPre)
+    val datasetWithID = splitterSummary.collect { case dc: DataCutterSummary =>
+      dc.preparedDF.getOrElse(datasetWithIDPre)
     }.getOrElse(datasetWithIDPre)
+
     val BestEstimator(name, estimator, summary) = bestEstimator.getOrElse {
       setInputSchema(dataset.schema).transformSchema(dataset.schema)
       val best = validator
@@ -158,10 +159,7 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
       best
     }
 
-    val preparedData = splitter.collect {
-      case _: DataCutter => datasetWithID
-      case sp => sp.validationPrepare(datasetWithID)
-    }.getOrElse(datasetWithID)
+    val preparedData = splitter.map(_.validationPrepare(datasetWithID)).getOrElse(datasetWithID)
 
     val bestModel = estimator.fit(preparedData).asInstanceOf[M]
     val bestEst = bestModel.parent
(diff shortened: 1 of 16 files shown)


tovbinm:
@gerashegalov can you elaborate more on the proposed fixes? thanks.

gerashegalov:
@tovbinm there are multiple issues.

One is the sort stability when there are multiple top labels with identical cardinality. We can solve this by adding the label as a secondary sort key. Similar issue with filtering labels for the minFraction prior to sorting. And #251 caused the data cutter omission to begin with

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -121,25 +166,30 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
       .filter(r => (r.getLong(1) / totalValues) >= minLabelFract)
-      .sort(col(colCount).desc)
       .take(maxLabels)
       .map(_.getDouble(0))
 
     val labelSet = labelsKeep.toSet
-    val labelsDropped = labelCounts.filter(r => !labelSet.contains(r.getDouble(0))).collect().map(_.getDouble(0)).toSet
+    val labelsDropped = labelCounts
+      .filter(r => !labelSet.contains(r.getDouble(0)))
+      .take(100)
```
leahmcguire: maybe instead of just keeping the first hundred dropped labels we should just keep a count of how many we dropped

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
```diff
@@ -138,16 +137,19 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
...
           .withColumn(ModelSelectorNames.idColName, monotonically_increasing_id())
       }
-    require(!datasetWithID.isEmpty, "Dataset cannot be empty")
+    require(!datasetWithIDPre.isEmpty, "Dataset cannot be empty")
 
-    val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithID))
+    val splitterSummary = splitter.flatMap(_.preValidationPrepare(datasetWithIDPre))
+    val datasetWithID = splitter.collect {
+      case dc: DataCutter => dc.validationPrepare(datasetWithIDPre)
+    }.getOrElse(datasetWithIDPre)
```
leahmcguire: if you want to avoid the match on case for the data prepare we could modify the return type of  preValidationPrepare to optionally return a dataset. then the data cutter preValidationPrepare could cut the data and the validationPrepare could just return the dataset unchanged...

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: Good idea. We could also do both: number of dropped labels and the topK of them for some low K, so we can log it `dropped label1, label3, label10 ...`

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
gerashegalov: yes. hacked something together to see if it's generally a good approach

tovbinm:
@gerashegalov please let us know when you want the PR reviewed.

gerashegalov:
@tovbinm it's ready

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
```diff
@@ -156,7 +158,11 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
       best
     }
 
-    val preparedData = splitter.map(_.validationPrepare(datasetWithID)).getOrElse(datasetWithID)
+    val preparedData = splitter.collect {
```
tovbinm: I find this part quite confusing. We apply `validationPrepare` only if it's a `DataCutter`, and then only if it's not `DataCutter` we apply `validationPrepare`. So what I get from this is that when it's a `DataCutter` we want to apply `validationPrepare` early, but not for any other splitter types. Perhaps it's worth adding a few lines of docs explaining why.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -86,16 +89,37 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
   }
 
+
+  private def getLabelsFromMetadata(data: DataFrame): Array[String] =
+    Try {
+      val labelSF = data.schema.head
+      val labelColMetadata = labelSF.metadata
+      log.info(s"Raw label column metadata: $labelColMetadata")
+      labelColMetadata
+        .getMetadata("ml_attr")
```
tovbinm: object `org.apache.spark.ml.attribute.AttributeKeys` has these consts defined. please use those if possible.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -86,16 +89,37 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
 
     import data.sparkSession.implicits._
 
-    val labels = data.map(r => r.getDouble(0) -> 1L)
-    val labelCounts = labels.groupBy(labels.columns(0)).sum(labels.columns(1)).persist()
-    val (resKeep, resDrop) = estimate(labelCounts)
+    val labels = data.map(r => r.getDouble(0))
+    val labelCounts = labels.groupBy(labels.columns(0)).count().persist()
```
tovbinm: `.count().persist()` seems weird to me. what's the return type of `count()`?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -86,16 +89,37 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+    val res = estimate(labelCounts)
     labelCounts.unpersist()
-    setLabels(resKeep, resDrop)
+    setLabels(res.labelsKept.toSet, res.labelsDropped.toSet, res.labelsDroppedTotal)
```
tovbinm: `.distinct` perhaps instead of `.toSet` so the order would be kept?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -104,11 +128,35 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+      dataPrep
+    } else {
+      log.info(s"Dropping rows with columns not in $labelSet")
+      val labelColName = dataPrep.columns(0)
+
+      val na = NominalAttribute.defaultAttr.withName(labelColName)
+      val metadataNA = if (labelMetaArr.isEmpty) {
+        na.withNumValues(labelSet.size)
+      } else {
+        // trim the metadata
```
tovbinm: what does `trim the metadata` and `trim the dataframe` means? please add more explanations  ;)

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -117,29 +165,48 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
    * @param labelCounts
    * @return Set of labels to keep & to drop
    */
-  private[op] def estimate(labelCounts: DataFrame): (Set[Double], Set[Double]) = {
+  private[op] def estimate(labelCounts: DataFrame): DataCutterSummary = {
+    val numDroppedToRecord = 10
```
tovbinm: why magic number 10? dont you want this configurable?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: considered it but it's a package-private class in `org.apache.spark.ml.attribute`

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: accidental leftover, did not remove enough stuff after getting rid of an unnecessary transformation `r.getDouble(0) -> 1L)`

gerashegalov: UPDATE: sorry, I got confused it belongs here for the same reason as before since we produce a df with `(label, count)` per [scaladoc](https://spark.apache.org/docs/2.2.0/api/scala/index.html#org.apache.spark.sql.RelationalGroupedDataset@count():org.apache.spark.sql.DataFrame)

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: good idea, toSet is the current behavior

gerashegalov:
not ready for review yet, checkpointing WIP,  sorting out test issues.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -82,33 +85,95 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+        case dcs: DataCutterSummary if dcs.preparedDF.isDefined => dcs
+      }
+      .getOrElse {
+        val labelMetaArr = getLabelsFromMetadata(data)
+        val labelSet = getLabelsToKeep.toSet
+
+        // discount unseen label
+        val integralLabelSet = 0.until(labelMetaArr.length - 1)
+          .map(_.toDouble)
+          .toSet
```
leahmcguire: I dont think you can convert this to a set - it is a sequence where the name is matched to the value by the index

gerashegalov: you are right Leah. I am also just getting rid of it because it was my old way to prevent multiple executions of trimming, which we now handle more explicitly

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -82,33 +85,87 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
-    import data.sparkSession.implicits._
-
-    val labels = data.map(r => r.getDouble(0) -> 1L)
-    val labelCounts = labels.groupBy(labels.columns(0)).sum(labels.columns(1)).persist()
-    val (resKeep, resDrop) = estimate(labelCounts)
-    labelCounts.unpersist()
-    setLabels(resKeep, resDrop)
+  override def preValidationPrepare(data: Dataset[Row], labelColNameOpt: Option[String]): Option[SplitterSummary] = {
+    val labelColName = labelColNameOpt.getOrElse(data.columns(0))
+    val labelColIdx = data.columns.indexOf(labelColName)
```
leahmcguire: I wouldnt assume that the first column is the label - that is the problem you had in the first place. Make a spark parameter on the base class where you set the name of the label column

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -82,33 +85,87 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+
+    if (Try(getLabelsToKeep).toOption.isEmpty) {
```
leahmcguire: this is a spark parameter just use isSet

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -82,33 +85,87 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+        val dataPrep = {
+          log.info(s"Dropping rows with columns not in $labelSet")
+
+          // Update metadata that spark.ml Classifier is using tp determine the number of classes
+          //
+          val na = NominalAttribute.defaultAttr.withName(labelColName)
+          val metadataNA = if (labelMetaArr.isEmpty) {
+            na
+          } else {
+            na.withValues(labelMetaArr.take(labelSet.size))
```
leahmcguire: I think part of the problem may be the assumption that the first labels are the ones that you keep (they should be but there are probably edge cases where it is not true). `labelMetaArr.zipWithIndex.filter{(case (_, i))labelSet.contains(i)}`

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -82,33 +85,87 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+            .withColumn(labelColName, data
+              .col(labelColName)
+              .as("_", metadataNA.toMetadata))
+        }
+
+        DataCutterSummary(
+          labelsKept = getLabelsToKeep,
+          labelsDropped = getLabelsToDrop,
+          labelsDroppedTotal = getLabelsDroppedTotal,
+          preparedDF = Option(dataPrep)
```
leahmcguire: We should not be putting the preparedDataFrame into the summary

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: you == current version? it's now a method parameter (which I could make a DC param).

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: agreed, more robust and elegant

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
leahmcguire: I thought the rearranging of the dataframe was to solve this ordering assumption...

gerashegalov:
marking as ready for review since all unit testing are passing

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -75,40 +78,91 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+
+    summary = Option(DataCutterSummary(
+      labelsKept = getLabelsToKeep,
+      labelsDropped = getLabelsToDrop,
+      labelsDroppedTotal = getLabelsDroppedTotal
+    ))
+
+    if (isSet(cacheValidatedDFForTesting) && $(cacheValidatedDFForTesting)) {
+      cachedDataFrameForTesting = Option(dataPrep)
+    }
```
leahmcguire: should we remove this before we merge?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -75,40 +78,91 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+    log.info(s"Raw label column metadata: $labelColMetadata")
 
-    dataUse
+    Try {
+      labelColMetadata
+        .getMetadata("ml_attr")
+        .getStringArray("vals")
+    }
+    .recover { case nonFatal =>
+      log.warn("Recovering non-fatal exception using num_vals", nonFatal)
```
leahmcguire: I dont understand these exceptions

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: it would be good to keep the functionality to prevent regressions in the future way. I suppose it would be more elegant to have a TestDataCutter extending DataCutter with the sole purpose of intercepting the result dataframe.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: I wanted to avoid too elaborate of the logic for checking whether these attributes are set and set correctly. The most common case is NoSuchKey when the attributes are not there.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataBalancer.scala
```diff
@@ -181,6 +178,22 @@ class DataBalancer(uid: String = UID[DataBalancer]) extends Splitter(uid = uid)
     balanced.persist()
   }
 
+  private def splitNegativePositive(data: Dataset[Row]): Seq[DataFrame] = {
+    val labelColOpt = if (isSet(labelColumnName)) {
+      Some(data($(labelColumnName)))
```
tovbinm: please use `Option` instead of `Some` (the worst thing is to have `Some(null)`

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -75,40 +78,91 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
-    import data.sparkSession.implicits._
+    log.info(s"Dropping rows with columns not in $labelSet")
 
-    val labels = data.map(r => r.getDouble(0) -> 1L)
-    val labelCounts = labels.groupBy(labels.columns(0)).sum(labels.columns(1)).persist()
-    val (resKeep, resDrop) = estimate(labelCounts)
-    labelCounts.unpersist()
-    setLabels(resKeep, resDrop)
+    // Update metadata that spark.ml Classifier is using tp determine the number of classes
+    //
```
tovbinm: remove extra line

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -75,40 +78,91 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+        }
+
+      na.withValues(newLabelMetaArr)
+    }
 
-    summary = Option(DataCutterSummary(labelsKept = getLabelsToKeep, labelsDropped = getLabelsToDrop))
-    summary
+    // filter low cardinality labels out of the dataframe to reduce the volume and  to keep
+    // it in sync with the new metadata.
+    //
```
tovbinm: remove extra line as well

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -75,40 +78,91 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+    val dataPrep = data
+      .filter(r => labelSet.contains(r.getDouble(labelColIdx)))
+      .withColumn(labelColName, data(labelColName).as("_", metadataNA.toMetadata))
```
tovbinm: what is this magical `"_"` underscore column? ;)

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
tovbinm: you can reach it by creating an object in the same namespace. please do so, since you are accessing the same const values numerous times in this code.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
tovbinm: it's not really a user friendly message either. please provide some more valuable information.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -174,18 +242,39 @@ private[impl] trait DataCutterParams extends Params {
...
   private[op] def getLabelsToDrop: Array[Double] = $(labelsToDrop)
 
+  private[op] final val labelsDroppedTotal = new LongParam(this, "labelsDroppedTotal",
+    "the number of labels dropped")
+
+  private[op] def getLabelsDroppedTotal: Long = $(labelsDroppedTotal)
+
+  final val maxLabelsDroppedForDiagnostics = new IntParam(this, "maxLabelsDroppedForDiagnostics",
+    "maximum number of dropped label categories to retain for logging",
+    ParamValidators.inRange(lowerBound = 0, upperBound = 100, lowerInclusive = true, upperInclusive = true)
```
tovbinm: is there a reason why do we set 100 as an upper limit?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
```diff
@@ -82,12 +79,23 @@ abstract class Splitter(val uid: String) extends SplitterParams {
...
   protected def checkPreconditions(): Unit =
     require(summary.nonEmpty, "Cannot call validationPrepare until preValidationPrepare has been called")
 
+  /**
+    * Add a splitter parameter to name the label column
+    * @param label
+    * @return
+    */
+  def withLabelColumnName(label: String): Splitter = {
+    if (!isSet(labelColumnName)) {
```
tovbinm: so there is not way to override the label column once it's set? why limit it?

core/src/test/scala/com/salesforce/op/stages/impl/classification/MultiClassificationModelSelectorTest.scala
```diff
@@ -311,9 +327,90 @@ class MultiClassificationModelSelectorTest extends FlatSpec with TestSparkContex
...
     fitted.modelStageIn.parent.extractParamMap().toSeq
-      .collect{ case p: ParamPair[_] if p.param.name == "impurity" => p.value }.head shouldBe myParam
+      .collect { case p: ParamPair[_] if p.param.name == "impurity" => p.value }.head shouldBe myParam
 
     val meta = ModelSelectorSummary.fromMetadata(fitted.getMetadata().getSummaryMetadata())
     meta.validationResults.head shouldBe myMetadata
   }
+
+  it should "trim low-cardinality labels during cross validation" in {
+    import org.scalatest.prop.TableDrivenPropertyChecks._
```
tovbinm: you can simply mixin `org.scalatest.prop.PropertyChecks` into your test instead of local import

core/src/test/scala/com/salesforce/op/stages/impl/classification/MultiClassificationModelSelectorTest.scala
```diff
@@ -311,9 +327,90 @@ class MultiClassificationModelSelectorTest extends FlatSpec with TestSparkContex
...
+          .transform(bigNoneIndexedData)
+          .select(labelColName, "features")
+          .persist()
+
+        val countDistinct = bigData.select(labelColName).distinct().count()
+        log.info(s"bigdata uniqs $countDistinct")
+        val bigDataWithMeta = if (metaDataMode == "withNumVals") {
+          val na = NominalAttribute.defaultAttr
+            .withName(labelColName) // no vals, no num_vals
+          bigData.drop("_").withColumn(labelColName, col(labelColName).as("_",
```
tovbinm: or here is the magical `"_"` column again

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
leahmcguire: ok but lets make it private to OP if we just want it for testing

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
leahmcguire: This is the number of names of dropped labels that you keep? can we change the name to reflect that? maybe maxNamesForDroppedLabels

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: sure , coding conventions from past employers are sneaking in

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: magic of avoiding using labelColName 3 times on the same line :)

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: too pragmatic to my taste but I don't mind

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: @tovbinm it's a reasonable limit to log. Int.MaxValue is unreasonable

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
gerashegalov: I find it difficult to reason about the logic without making this param immutable once set.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
tovbinm: Look artificial to me and should perhaps be relative to data size. let's extend it to 1000 at least.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
tovbinm: I think this will bite us in the future.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
gerashegalov: anything proportional to data size will have the effect of creating huge un-GC'able  collections in degenerate cases, thus a hard cap on top of it is also necessary. Less of a concern if we just pipe it to logging instead of storing it in a param.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
gerashegalov: No doubt, but I think it will bite us unpredictably if we make it mutable. I can add a warning log instead of silently refusing to modify it.

tovbinm:
🎆 🍾

PR #264, opened 2019-04-03
Possibility to Return top K positives and top K negatives for LOCO

michaelweilsalesforce:

New feature used only for Binary Classification (probability of predicting a positive label) and Regression. 

The map's contents are different regarding the value of the topKStrategy param :
If PositiveNegative, returns at most 2 x topK elements : the topK most positive and the topK most negative derived features based on the LOCO insight.
If Abs, returns at most topK elements : the topK derived features having the highest absolute value of LOCO score.


core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -33,22 +33,29 @@ package com.salesforce.op.stages.impl.insights
...
+ *
+ * The map's contents are different regarding the value of the topKStrategy param (only for Binary Classification
+ * and Regression) :
+ * - If PositiveNegative, returns at most 2 * topK elements : the topK most positive and the topK most negative
+ * derived features based on the LOCO insight.
+ * - If Abs, returns at most topK elements : the topK derived features having highest absolute value of LOCO score.
+ * @param model model instance that you wish to explain
+ * @param uid   uid for instance
  */
 @Experimental
```
tovbinm: should we remove the `@Experimental` flag already?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -61,10 +68,22 @@ class RecordInsightsLOCO[T <: Model[T]]
...
   def getTopK: Int = $(topK)
+
   setDefault(topK -> 20)
 
+  final val topKStrategy = new Param[String](parent = this, name = "topKStrategy",
+    doc = "Whether returning topK based on absolute value or topK positives and negatives. Only for Binary " +
+      "Classification and Regression."
+  )
+
+  def setTopKStrategy(strat: TopKStrategy): this.type = set(topKStrategy, strat.entryName)
```
tovbinm: please name fully - `strategy: TopKStrategy`

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -74,9 +93,28 @@ class RecordInsightsLOCO[T <: Model[T]]
...
   private lazy val featureInfo = OpVectorMetadata(getInputSchema()(in1.name)).getColumnHistory().map(_.toJson(false))
 
+  private lazy val vectorDummy = Array.fill(featureInfo.length)(0.0).toOPVector
+
+  private[insights] lazy val problemType = modelApply(labelDummy, vectorDummy).score.length match {
+    case 0 => ProblemType.Unknown
+    case 1 => ProblemType.Regression
+    case 2 => ProblemType.BinaryClassification
+    case n if (n > 2) => {
+      log.info("MultiClassification Problem : Top K LOCOs by absolute value")
```
tovbinm: This is not a very user friendly message. Let's log some meaningful message for any problem type with some clear and comprehensible explanation.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -86,26 +124,102 @@ class RecordInsightsLOCO[T <: Model[T]]
...
-      featureArray.update(i, (oldInd, 0))
-      val score = modelApply(labelDummy, OPVector(Vectors.sparse(featureSize, featureArray))).score
-      val diffs = baseScore.zip(score).map{ case (b, s) => b - s }
-      val max = diffs.maxBy(math.abs)
-      maxHeap.enqueue((i, max, diffs))
-      if (i >= k) maxHeap.dequeue()
-      featureArray.update(i, (oldInd, oldVal))
-      i += 1
+    val top = $(topKStrategy) match {
+      case s if s == TopKStrategy.Abs.entryName || problemType == ProblemType.MultiClassification => {
```
tovbinm: add a getter method for strategy `def getTopKStrategy: TopKStrategy` and then use it here instead of string equality: `$(topKStrategy) -> problemType match { case (TopKStrategy.Abs, ProblemType.MultiClassification) => ... }`

tovbinm: please also create a function for each of the cases to avoid this method growing large

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -86,26 +124,102 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+        // Heap that will contain the top K positive LOCO values
+        val positiveMaxHeap = PriorityQueue.empty(MinScore)
+        // Heap that will contain the top K negative LOCO values
+        val negativeMaxHeap = PriorityQueue.empty(MaxScore)
+        // for each element of the feature vector != 0.0
+        // Size of positive heap
+        var positiveCount = 0
+        // Size of negative heap
+        var negativeCount = 0
+        for {i <- 0 until filledSize} {
```
tovbinm: use `while` - it's faster

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -86,26 +124,102 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+          val diffs = computeDiffs(i, oldInd, featureArray, featureSize, baseScore)
+          val max = if (problemType == ProblemType.Regression) diffs(0) else diffs(1)
+
+          if (max > 0.0) { // if positive LOCO then add it to positive heap
+            positiveMaxHeap.enqueue((i, max, diffs))
+            positiveCount += 1
+            if (positiveCount > k) { // remove the lowest element if the heap size goes from 5 to 6
+              positiveMaxHeap.dequeue()
+            }
+          } else if (max < 0.0) { // if negative LOCO then add it to negative heap
```
tovbinm: what about `max == 0.0`? it's not handled.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -86,26 +124,102 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+        val negativeMaxHeap = PriorityQueue.empty(MaxScore)
+        // for each element of the feature vector != 0.0
+        // Size of positive heap
+        var positiveCount = 0
+        // Size of negative heap
+        var negativeCount = 0
+        for {i <- 0 until filledSize} {
+          val (oldInd, oldVal) = featureArray(i)
+          val diffs = computeDiffs(i, oldInd, featureArray, featureSize, baseScore)
+          val max = if (problemType == ProblemType.Regression) diffs(0) else diffs(1)
```
tovbinm: this seems very hacky and error prone - `if (problemType == ProblemType.Regression) diffs(0) else diffs(1)`, perhaps add a helper function?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -121,59 +164,40 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext {
...
+
+
+    val transformer = new RecordInsightsLOCO(sparkModel).setTopKStrategy(TopKStrategy.PositiveNegative)
+      .setInput(featureVector)
+
+    val insights = transformer.transform(testDataMeta)
+    val parsed = insights.collect(name, transformer.getOutput())
+      .map { case (n, i) => n -> RecordInsightsParser.parseInsights(i) }
+    parsed.foreach { case (_, in) =>
+      in.head._1.columnName == "1_1_1_1" || in.last._1.columnName == "3_3_3_3" shouldBe true
```
tovbinm: what does this even mean `in.head._1.columnName == "1_1_1_1" || in.last._1.columnName == "3_3_3_3"`?
please add a `withClue` around it to allow easier debugging in case of failures.

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -121,59 +164,40 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext {
...
-    parsed.foreach { case (_, in) => math.abs(in.head._2(0)._2 + in.head._2(1)._2) < 0.00001 shouldBe true }
+    parsed.foreach { case in => math.abs(in.head._2(0)._2 + in.head._2(1)._2) < 0.00001 shouldBe true }
+  }
+
+  it should "return the most predictive features when using top K Positives + top K negatives strat" in {
+    val (testData, name, labelNoRes, featureVector) = TestFeatureBuilder("name", "label", "features", data)
+    val label = labelNoRes.copy(isResponse = true)
+    val testDataMeta = addMetaData(testData, "features", 5)
+    val sparkModel = new OpLogisticRegression().setInput(label, featureVector).fit(testData)
+
```
tovbinm: remove redundant lines

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -74,9 +93,28 @@ class RecordInsightsLOCO[T <: Model[T]]
 
   private lazy val featureInfo = OpVectorMetadata(getInputSchema()(in1.name)).getColumnHistory().map(_.toJson(false))
 
+  private lazy val vectorDummy = Array.fill(featureInfo.length)(0.0).toOPVector
+
+  private[insights] lazy val problemType = modelApply(labelDummy, vectorDummy).score.length match {
+    case 0 => ProblemType.Unknown
+    case 1 => ProblemType.Regression
+    case 2 => ProblemType.BinaryClassification
+    case n if (n > 2) => {
```
leahmcguire: I was thinking about the multiclass issue -  I think it makes sense to return the difference in score for the winning category so maybe instead of having this find the problem type it should return the index to examine (also move it down to the base score function line 117:

val baseResult = modelApply(labelDummy, features)
val baseScore =  baseResult.score
val indexToExamine = baseScore.length match {
    case 0 => throw new RuntimeExeption("model does not produce scores for insights")
    case 1 => 0
    case 2 => 1
    case n if (n > 2) => baseResult.prediction

and then always use this index for pulling out the diff for the heaps

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
leahmcguire: no - we want to keep it so we can change things :-) For instance the way of creating this class is still terrible

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -86,26 +124,102 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+          i += 1
+        }
+
+        val topAbs = absoluteMaxHeap.dequeueAll
+
+
+        topAbs.sortBy { case (_, v, _) => -math.abs(v) }
+
+      }
+      case s if s == TopKStrategy.PositiveNegative.entryName => {
```
leahmcguire: what if we always kept the positive and negative heap and then just check the strategy when we return the insights?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
leahmcguire: this is where you would use the indexToExamine variable I suggested above

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
michaelweilsalesforce: It would be slower when we want to return the top by abs value. Is there an easy way to go from 2 Priority Queues to 1?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
michaelweilsalesforce: No. LOCOs of 0.0 are not interesting

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
leahmcguire: It would be slightly slower but not by much if we assume that (or enforce that) k is small. To go to one q - just get all the answers and sort them. Again you are right that it is slower but if we limit k to ~100 it is a pretty small difference.

leahmcguire: Or you can get the first of each que and then basically keep popping off the last q that you took from for comparison - then it is not much slower you just have a sightly larger memory footprint

michaelweilsalesforce: Another issue @leahmcguire. For Multiclassification, we return the highest absolute LOCO value, and for the other strategy we look at the winning category (baseScore). Getting the top Positive and negative for winning class doesn't guarantee the highest absolute value : what if this absolute value was from another class?

leahmcguire: Yeah - I think that the fact that we were just taking the highest absolute value was a mistake. It allows for things like baseScore = (0.1, 0.3, 0.6) change in score = (0.3, 0.1, 0.6) to be the difference we report, which seems wrong. Always choosing the change in the winning value will mean that the change we report always has to do with the score that one, even if it is not the largest change in score. So yes we would change strategy for multiclass in absolute value as well - but I think it is slightly more principled...WDYT?

michaelweilsalesforce: Now the question is should we apply the same logic for Binary Classification. 
So far the spikes were focused on the positive class (label 1.0) because we thought it was the the most interesting info to report.

However, recent spikes witness a strong relationship  between LOCO scores and prediction scores (cc @crupley) : Rows with the "highest scores" had on average its top absolute LOCOs being mostly positive. For "lowest scores", those locos are mostly negative.
One can also argue that `LOCO for class 0 = - LOCO for class 1`, hence getting top Positives and Negatives is sufficient. 
Maybe something we might need to add to the output is the base score.

leahmcguire: Hmm, that is interesting.  But wouldn't it be confusing to interpret for Binary if we choose the winning value because they are symetric? the scores will always some to 1 so a positive contribution to the prediction of 1 will be the same negative contribution to the prediction of 0. If we switch based on the prediction they will have to keep in mind the winning prediction to interpret the effect - this is something you need to do in multiclass but seems like an extra complication in binary.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -74,38 +96,104 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+    // for each element of the feature vector != 0.0
+    // Size of positive heap
+    var positiveCount = 0
+    // Size of negative heap
+    var negativeCount = 0
+    var i = 0
+    while (i < filledSize) {
+      val (oldInd, oldVal) = featureArray(i)
+      val diffs = computeDiffs(i, oldInd, featureArray, featureSize, baseScore)
+      val max = diffs(indexToExamine)
```
leahmcguire: maybe rename to diffToExamine

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -74,38 +96,104 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+          positiveMaxHeap.dequeue()
+        }
+      } else if (max < 0.0) { // if negative LOCO then add it to negative heap
+        negativeMaxHeap.enqueue((i, max, diffs))
+        negativeCount += 1
+        if (negativeCount > k) { // remove the highest element if the heap size goes from 5 to 6
+          negativeMaxHeap.dequeue()
+        } // Not keeping LOCOs with value 0
+      }
+      featureArray.update(i, (oldInd, oldVal))
```
leahmcguire: would it make sense to move the update of the array back tot the old version into the compute diffs function?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -74,38 +96,104 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+    }
+    val topPositive = positiveMaxHeap.dequeueAll
+    val topNegative = negativeMaxHeap.dequeueAll
+    (topPositive ++ topNegative)
+  }
 
+  override def transformFn: OPVector => TextMap = (features) => {
+    val baseResult = modelApply(labelDummy, features)
+    val baseScore = baseResult.score
+      modelApply(labelDummy, features).prediction
```
leahmcguire: think line 148 is leftover from a refactor

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -85,27 +153,47 @@ class RecordInsightsLOCO[T <: Model[T]]
...
-      featureArray.update(i, (oldInd, oldVal))
-      i += 1
+    val indexToExamine = baseScore.length match {
+      case 0 => throw new RuntimeException("model does not produce scores for insights")
+      case 1 => 0
+      case 2 => 1
+      case n if (n > 2) => baseResult.prediction.toInt
+    }
+    val topPosNeg = returnTopPosNeg(filledSize, featureArray, featureSize, baseScore, k, indexToExamine)
+    val top = getTopKStrategy match {
```
tovbinm:
```
getTopKStrategy match {
   case TopKStrategy.Abs =>
   case TopKStrategy.PositiveNegative =>  
}
```

shenzgang:
hello!How does the multi-category evaluator scale to compute roc,pr, and FPR and TPR for each category？

PR #265, opened 2019-04-03
Added string transformers for substring search and valid email

leahmcguire:


README.md
```diff
@@ -16,6 +16,7 @@ Use TransmogrifAI if you need a machine learning library to:
 To understand the motivation behind TransmogrifAI check out these:
 - [Open Sourcing TransmogrifAI: Automated Machine Learning for Structured Data](https://engineering.salesforce.com/open-sourcing-transmogrifai-4e5d0e098da2), a blog post by [@snabar](https://github.com/snabar)
 - [Meet TransmogrifAI, Open Source AutoML That Powers Einstein Predictions](https://www.youtube.com/watch?v=93vsqjfGPCw&feature=youtu.be&t=2800), a talk by [@tovbinm](https://github.com/tovbinm)
+- [Low Touch Machine Learning](https://www.youtube.com/watch?v=PKTvo9X9Sjg), a talk by [@leahmcguire](https://github.com/leahmcguire)
```
tovbinm: 👍

core/src/main/scala/com/salesforce/op/dsl/RichTextFeature.scala
```diff
@@ -98,8 +98,12 @@ trait RichTextFeature {
      * @param nGramSize the size of the n-gram to be used to compute the string distance
      * @return ngrammed feature
      */
-    def toNGramSimilarity(that: FeatureLike[T], nGramSize: Int = NGramSimilarity.nGramSize): FeatureLike[RealNN] =
-      f.transformWith(new TextNGramSimilarity[T](nGramSize), that)
+    def toNGramSimilarity(
+      that: FeatureLike[T],
+      nGramSize: Int = NGramSimilarity.nGramSize,
+      toLowerCase: Boolean = TextTokenizer.ToLowercase
```
tovbinm: please update the doc on the shortcut

core/src/main/scala/com/salesforce/op/stages/impl/feature/ValidEmailTransformer.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.salesforce.op.stages.base.unary.UnaryTransformer
+
+/**
+ * Checks if an email is valid
+ * @param uid           uid for instance
+ */
+class ValidEmailTransformer(uid: String = UID[ValidEmailTransformer]) extends
+  UnaryTransformer[Email, Binary](operationName = "isValidEmail", uid = uid) {
+  override def transformFn: Email => Binary = (in: Email) => {
+    if (in.isEmpty) None
```
tovbinm: use `Binary.empty` - this would avoid necessary allocation, ie.  `if (in.isEmpty) Binary.empty else Some(in.prefix.nonEmpty && in.domain.nonEmpty).toBinary`

core/src/test/scala/com/salesforce/op/stages/impl/feature/ValidEmailTransformerTest.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+import com.salesforce.op.features.types._
+import com.salesforce.op.test.{OpTransformerSpec, TestFeatureBuilder}
+
+@RunWith(classOf[JUnitRunner])
+class ValidEmailTransformerTest extends OpTransformerSpec[Binary, ValidEmailTransformer] {
+
+  val sample = Seq(Email("abc"), Email("a@b"), Email("a@"), Email("@blah"), Email.empty, Email("real@stuff"))
```
tovbinm: `real@stuff` does not seem to be a valid domain to me. how robust do we what this validation code to be? perhaps it's worth adding a proper email domain? (long awaited btw ;) - https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/features/types/Text.scala#L85

leahmcguire: The issue is that you can actually have an email domain like that. The only way we could really validate is if we used a service that checked if the inbox exists....

PR #268, opened 2019-04-06
Changes to allow passing reader as optional param to be able to  gene…

yufeldman:
…rateDataFrame based on the outside reader

versus reader set on the workflowmodel, as latter one can be cached

**Related issues**

**Additional context**

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
@@ -219,8 +219,14 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
    *
    * @return Dataframe with all the features generated + persisted
    */
-  protected def generateRawData()(implicit spark: SparkSession): DataFrame = {
-    (reader, rawFeatureFilter) match {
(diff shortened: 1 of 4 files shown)


leahmcguire:
You can always change the reader on a workflow  or workflow model using `setReader`.  It just seems confusing to add another way to do this

yufeldman:
> You can always change the reader on a workflow or workflow model using `setReader`. It just seems confusing to add another way to do this

workflow is cached, if I set a reader on it, it won't work with concurrent requests. This is my amateur way of fixing caching issues - I will be discussing it, so it may change completely. This PR was more of a reference point - at least at first

amateurhuman:
@yufeldman you've been invited to the org, once you accept it you can refresh the CLA check by going to https://cla.salesforce.com/status/salesforce/TransmogrifAI/pull/268

leahmcguire:
Are you working on streaming? we have a separate flow for streaming https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/OpWorkflowRunner.scala#L232

tovbinm:
@leahmcguire we just discussed with @yufeldman how to allow what they need. I opened a PR to do that #270

PR #270, opened 2019-04-08
Allow copying model instances

tovbinm:
**Related issues**
Currently once the model is loaded it's impossible to reuse the same instance across multiple threads.

Added `model.copy()` to allow deep copy of model instance so each thread would get it's own instance.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
@@ -32,12 +32,13 @@ package com.salesforce.op
 
 import com.salesforce.op.evaluators.{EvaluationMetrics, OpEvaluatorBase}
 import com.salesforce.op.features.types.FeatureType
-import com.salesforce.op.features.{FeatureLike, OPFeature}
+import com.salesforce.op.features.{Feature, FeatureLike, OPFeature}
 import com.salesforce.op.readers.DataFrameFieldNames._
 import com.salesforce.op.stages.{OPStage, OpPipelineStage, OpTransformer}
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
```diff
@@ -419,6 +420,31 @@ class OpWorkflowModel(val uid: String = UID[OpWorkflowModel], val trainingParams
...
+   */
+  def copy(): OpWorkflowModel = {
+    def copyFeatures(features: Array[OPFeature]): Array[OPFeature] = features.collect { case f: Feature[_] =>
+      implicit val tt = f.wtt
+      f.copy()
+    }
+    val copy =
+      new OpWorkflowModel(uid = uid, trainingParams = trainingParams.copy())
+        .setFeatures(copyFeatures(resultFeatures))
+        .setRawFeatures(copyFeatures(rawFeatures))
```
leahmcguire: you shouldnt set the raw features - they should be derived from the features

tovbinm: good point. what if somebody modifies it separately? I dont want to mess up any internal state.

leahmcguire: they shouldnt be able to - it is private to op. setting the raw features should never be done directly. In fact the method has no usages maybe we shoudl delete it

tovbinm: Ha ha ha!! Removed it.

PR #274, opened 2019-04-09
Workflow independent model loading

tovbinm:
**Related issues**
Following up on #269 - a sequence of PRs on allowing model loading without workflows...

**1.**
 
This PR finally allows loading models without recreating the workflow by simply calling: 
```scala
val model = OpWorkflowModel.load(path)
```
Importantly the old syntax (i.e. `workflow.loadModel(path)`), still works as before.

With this change we are going to be serializing all transformers and feature generator instances, then reconstructing them when loading the model without requiring the original workflow.

1. Introduce a new reader/writer interface with a default reflection based implementation
2. Update stage reader & writer to use it
3. Add custom serializer for feature generator stages
3. Update model reader & writer to use new stage read / write technique

Of course updated all the tests with the changes.

**Migration Notes**
All lamdba transformers and extract functions cannot neither be inline nor have external dependencies on variables. Functions are now required to be specified in objects (Scala 2.11) and as concrete classes for Scala 2.12 and up. Example:
```scala
// the following syntax wont work anymore and error at runtime:
val description = FeatureBuilder.Text[Passenger].extract(_.getDescription.toText).asPredictor
val lowerCase = textFeature.map[Text](t => t.value.map(_.toLowerCase).toText)

// for Scala 2.11 replace with follows:
val description = FeatureBuilder.Text[Passenger].extract(Lambdas.descriptionExtract).asPredictor
val lowerCase = textFeature.map[Text](Lambdas.toLowercase)
object Lambdas {
    def descriptionExtract: Passenger => Text = _.getDescription.toText
    def toLowercase: Text => Text = t => t.value.map(_.toLowerCase).toText
}
// for Scala 2.12 and up create concrete function classes:
class DescriptionExtract extends Function1[Passenger, Text] { def apply(p: Passenger): Text = p.getDescription.toText }
val description = FeatureBuilder.Text[Passenger].extract(new DescriptionExtract).asPredictor
```

**2.**
Additionally one can now provide a custom reader/writer for each transformer stage if necessary as follows:
```scala
// implement a custom reader/writer for your stage
class MyStageReaderWriter extends OpPipelineStageReaderWriter[MyStage] {
    def read(stageClass: Class[MyStage], json: JValue): Try[MyStage] = ???
    def write(stage: MyStage): Try[JValue] = ???
}

// add annotation with ReaderWriter to your transformer with the customer reader/writer implementation class
@ReaderWriter(classOf[MyStageReaderWriter])
class MyStage(...) extends UnaryTransformer[Text, Text] { ... } 
```
[Here](https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala#L114) is a full example for our own TextTokenizer stage.


tovbinm:
Can anyone please review this PR?

features/src/test/scala/com/salesforce/op/stages/base/binary/BinaryTransformerTest.scala
```diff
@@ -42,10 +42,14 @@ class BinaryTransformerTest extends OpTransformerSpec[Real, BinaryTransformer[Re
   val sample = Seq(Real(1.0) -> RealNN(0.0), Real(2.0) -> RealNN(2.0), Real.empty -> RealNN(1.0))
   val (inputData, f1, f2) = TestFeatureBuilder(sample)
 
-  val transformer = new BinaryLambdaTransformer[Real, RealNN, Real](operationName = "bmi",
-    transformFn = (i1, i2) => new Real(for { v1 <- i1.value; v2 <- i2.value } yield v1 / (v2 * v2))
+  val transformer = new BinaryLambdaTransformer[Real, RealNN, Real](
```
Jauntbox: General question for our new transformers. I thought we could still use lambdas, they just won't serialize when we save the model so the only way to load them is if you also have the workflow. Or is that not the case?

If we still can use lambdas in our transformers as long as we don't try to save/load models, then are you rewriting all the tests to follow this pattern just so people don't see a version that will suddenly break when they try and load things?

tovbinm:
Kudos to @wsuchy @Jauntbox for all the help!

tovbinm:
@Jauntbox 

1. Correct.
2. I think we should deprecate `workflow.loadModel(path)` call eventually just to keep things simple.
3. Yes, we will update them before the next release.
4. Some of the test changes were not necessary, though I did it anyways.

PR #276, opened 2019-04-10
Proper UID for Raw Feature Filter

tovbinm:
**Related issues**
RFF has an invalid UID

Proper UID for Raw Feature Filter

### DIFF
--- a/core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
+++ b/core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
@@ -592,6 +592,9 @@ object RawFeatureFilter {
   // scoring sets since they will not be reliable. Currently, this is set to the same as the minimum training size.
   val minScoringRowsDefault = 500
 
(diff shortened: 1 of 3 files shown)


clin-projects:
LGTM!

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -592,6 +592,9 @@ object RawFeatureFilter {
   // scoring sets since they will not be reliable. Currently, this is set to the same as the minimum training size.
   val minScoringRowsDefault = 500
 
+  val stageName = classOf[RawFeatureFilter[_]].getSimpleName
+
+  val uid = s"${stageName}_100000000000"
```
leahmcguire: why not use the UID method?

tovbinm: I guess it's better to have it be the same (!?) not important really, since there is only one.

PR #280, opened 2019-04-11
Fixed onSetInput so is always called with new input

leahmcguire:
**Related issues**
Metadata for transformers was not being updated when raw feature filter removed features

Moved onSetInput call into the base (internal) set input call

**Additional context**

### DIFF
--- a/features/src/main/scala/com/salesforce/op/stages/OpPipelineStages.scala
+++ b/features/src/main/scala/com/salesforce/op/stages/OpPipelineStages.scala
@@ -84,11 +84,6 @@ trait OpPipelineStageBase extends OpPipelineStageParams with MLWritable {
   }
 
   /**
-   * Function to be called on setInput
-   */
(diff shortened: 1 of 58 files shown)


tovbinm:
@leahmcguire there are some build failures you might want to look into.

features/src/main/scala/com/salesforce/op/stages/OpPipelineStages.scala
```diff
@@ -165,6 +159,11 @@ trait OpPipelineStageBase extends OpPipelineStageParams with MLWritable {
 
   final override def write: MLWriter = new OpPipelineStageWriter(this)
 
+
+  /**
+   * Function to be called on setInput
+   */
+  override protected def onSetInput(): Unit = {}
```
leahmcguire: @tovbinm any idea why I had to override the method here to make all the tests using the pipeline stage spec pass?

PR #282, opened 2019-04-12
Skip writing empty stages & skip loading stages without uid-s

tovbinm:
**Related issues**
Follows up on #281.
Every stage must have a UID, if so some reason uid field is missing from the model json currently we fail miserably.

As precaution: 1) skip saving empty stages 2) skip reading stages without uid-s


tovbinm:
@wsuchy every Spark `org.apache.spark.ml.PipelineStage` extends `Params`, which extends `Identifiable` which required `val uid: String` -  https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/util/Identifiable.scala#L39

PR #283, opened 2019-04-15
Improved test by extending OpTransformerSpec

wsuchy:
I've kept original tests as well.

### DIFF
--- a/core/src/test/scala/com/salesforce/op/stages/impl/feature/ToOccurTransformerTest.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/feature/ToOccurTransformerTest.scala
@@ -31,17 +31,22 @@
 package com.salesforce.op.stages.impl.feature
 
 import com.salesforce.op.features.types._
-import com.salesforce.op.test.{TestFeatureBuilder, TestSparkContext}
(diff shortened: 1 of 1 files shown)


core/src/test/scala/com/salesforce/op/stages/impl/feature/ToOccurTransformerTest.scala
```diff
@@ -57,9 +62,9 @@ class ToOccurTransformerTest extends FlatSpec with TestSparkContext {
   }
 
   lazy val (ds, leadId, numEmails, opptyId, doNotContact, numFormSubmits) =
-    TestFeatureBuilder("leadId", "numEmails", "opptyId", "doNotContact", "numFormSubmits", testData)
+    TestFeatureBuilder("leadId", "numEmails", "opptyId", "doNotContact", "numFormSubmits", extendedTestData)
 
-  Spec[ToOccurTransformer[_]] should "convert features to doolean using shortcuts" in {
+  Spec[ToOccurTransformer[_]] should "convert features to boolean using shortcuts" in {
```
tovbinm: nit: you can do `it should` here (since the test name is already set in the base class)

Issue #278, opened 2019-04-11
Improve test coverage for transformers & estimators


tovbinm:
**Problem**
Some of our transformers & estimators are not thoroughly tested or not tested at all.

**Solution**
Use `OpTransformerSpec` and `OpEstimatorSpec` base test specs to provide tests for all existing  transformers & estimators.

crupley:
After a quick survey of `com.salesforce.op.stages.impl.feature`, found the following:

## These classes don't appear to have associated tests:

* FilterMap
* OpIndexToString (just a wrapper for Spark IndexToString, though)
* OpLDA (has test, OpLdaTest, but names don't match)
* OpOneHotVectorizer
* OpScalarStandardScaler (OpStandardScalerTest exists but names don't match)
* RealNNVectorizer
* TextMapPivotVectorizer (tested in both OPMapVectorizerTest and TextMapVectorizerTest but names don't match)
* Transmogrifier (TransmogrifyTest exists but names don't match)

## These tests exist but do not extend `OpTransformerSpec` or `OpEstimatorSpec`

* Base64VectorizerTest
* DateMapVectorizerTest
* DateTimeVectorizerTest
* DateVectorizerTest
* EmailParserTest
* EmailVectorizerTest
* FillMissingWithMeanTest
* GeolocationVectorizerTest
* HashingTFTest
* IntegralVectorizerTest
* IsotonicRegressionCalibratorTest
* LinearScalerTest
* MultiPickListMapVectorizerTest
* NGramSimilarityTest
* NGramTest
* NumericBucketizerTest
* NumericVectorizerTest
* OPCollectionHashingVectorizerTest
* OPCollectionTransformerTest
* OpCountVectorizerTest
* OpIndexToStringNoFilterTest
* OpLdaTest
* OPMapVectorizerTest
* OpSetVectorizerTest
* OpStandardScalerTest
* OpStringIndexerNoFilterTest
* OpStringIndexerTest
* OpWord2VecTest
* PercentileCalibratorTest
* PhoneNumberParserTest
* RealVectorizerTest
* ScalerMetadataTest
* ScalerTest
* SmartTextMapVectorizerTest
* TextMapVectorizerTest
* TextTokenizerTest
* TextTransmogrifyTest
* TextVectorizerTest
* ToOccurTransformerTest
* TransmogrifyTest
* UniqueCountTest
* URLVectorizerTest

Sammyalhashe:
[...]

PR #284, opened 2019-04-15
Update names of transformer tests and files to match class names

crupley:
**Related issues**

https://github.com/salesforce/TransmogrifAI/issues/278

Update names of some test classes and files to match name of the class they are testing

None

**Additional context**
None


core/src/test/scala/com/salesforce/op/stages/impl/feature/OpLDATest.scala
```diff
@@ -78,7 +78,7 @@ class OpLdaTest extends FlatSpec with TestSparkContext {
     .toSeq
     .map(_.getAs[Vector](0))
 
-  Spec[OpLdaTest] should "convert document term vectors into topic vectors" in {
+  Spec[OpLDATest] should "convert document term vectors into topic vectors" in {
```
tovbinm: Spec[OpLDA]

PR #285, opened 2019-04-16
Created new tests for OpCollection*

wsuchy:
Added new tests for `OpCollection*` that extends `OpTransformerSpec`. 
I've kept original tests as they are more detailed.

### DIFF
--- a/core/src/test/scala/com/salesforce/op/stages/impl/feature/OPMapTransformerTest.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/feature/OPMapTransformerTest.scala
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2017, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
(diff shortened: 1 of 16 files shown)


core/src/test/scala/com/salesforce/op/stages/impl/feature/OPMapTransformerTest.scala
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * @author ksuchanek
+ * @since 214
+ */
+@RunWith(classOf[JUnitRunner])
+class OPMapTransformerTest extends OpTransformerSpec[IntegralMap, TransformerType] {
+
+  import OPMapTransformerTest._
+
+  lazy val (dataEmailMap, top) = TestFeatureBuilder("name",
```
tovbinm: `lazy val (inputData, top) = ...`

core/src/test/scala/com/salesforce/op/stages/impl/feature/OPMapTransformerTest.scala
```diff
@@ -0,0 +1,89 @@
...
+    IntegralMap(Map("p1" -> 10L, "p2" -> 11L))
+  )
+}
+
+object OPMapTransformerTest {
+  type TransformerType = OPMapTransformer[Email, Integral, EmailMap, IntegralMap]
+
+  class BaseTransformer extends UnaryTransformer[Email, Integral](
+    operationName = "testUnary",
+    uid = "1234"
```
tovbinm: `uid = UID[BaseTransformer]`

wsuchy:
@tovbinm  I pushed changes regarding your comments. The companion objects are necessary because of the underlying lambdas and this is the easiest way of making serializing/deserializing working.

tovbinm:
@wsuchy I dont see any lambdas. these are just regular transformers. simply define them in the same file as the test. I.e. this should work:
```scala
class OPListTransformerTest extends OpTransformerSpec[TextList, BaseTransformer] {
  // ... 
  val transformer new = BaseTransformer()
  // ... 
}

class BaseTransformer extends UnaryTransformer[Text, Text](
  operationName = "testUnary",
  uid = UID[BaseTransformer]
) {
  override def transformFn: (Text => Text) = (input: Text) => input.value.map(_.toLowerCase()).toText
}

```

PR #293, opened 2019-04-17
Threshold metrics calculation fix when unseen labels are present

Jauntbox:
**Related issues**
Issue exposed by fixes in #263 

Simple one-line fix to threshold metrics calculation. When the true label score is calculated, it will either come from the list of labels the model was trained on (which may be pruned by DataCutter) or be 0 (eg. if it corresponds to a label the model was not trained on).

n/a

**Additional context**
Merge #263 first since this PR was branched off of it.
Unit test resides in MultiClassificationModelSelectorTest


core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -75,40 +78,91 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
-    val dataUse = dataPrep.filter(r => keep.contains(r.getDouble(0)))
+  def getLabelsFromMetadata(data: DataFrame): Array[String] = {
+    val labelSF = data.schema.head
+    val labelColMetadata = labelSF.metadata
+    log.info(s"Raw label column metadata: $labelColMetadata")
 
-    dataUse
+    Try {
+      labelColMetadata
+        .getMetadata("ml_attr")
```
wsuchy: Maybe we could keep those magic values as constants?

gerashegalov: @wsuchy This PR is very confusing because it's based and contains commits from an older version of #263. @Jauntbox's change is actually a small commit.  @Jauntbox please open a PR stacked on mine

Jauntbox: @gerashegalov I don't know how to make this PR against your branch since it's from a branch on your fork. This should be up to date w/ your branch now. I was hoping we could merge your branch first and then this PR should reduce down to ~10 lines when I merge it with master.

gerashegalov: You can fetch branches from other forks by `git remote add` other remotes first, branch off of it, and  push to another remote, and and create a cross-fork PR. Here is a demo: 
```
git remote add mtovbin-github git@github.com:tovbinm/TransmogrifAI.git
git fetch mtovbin-github 
git checkout mt/xgboost 
git checkout gera/crossfork-mt-xgboost
git checkout -b gera/crossfork-mt-xgboost
vim README.md 
git commit -a -m test 
git push gera-github gera/crossfork-mt-xgboost 
```
 https://github.com/tovbinm/TransmogrifAI/pull/1/files

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -188,7 +188,8 @@ private[op] class OpMultiClassificationEvaluator
     def computeMetrics(scoresAndLabels: (Array[Double], Double)): MetricsMap = {
       val scores: Array[Double] = scoresAndLabels._1
       val label: Label = scoresAndLabels._2.toInt
-      val trueClassScore: Double = scores(label)
+      // The label may be unseen during model training, so treat scores for unseen classes as all being zero
+      val trueClassScore: Double = if (scores.isDefinedAt(label)) scores(label) else 0.0
```
gerashegalov: how about `scores.lift(label).getOrElse(0.0)`

Jauntbox: Ahh, I was trying to remember how to do that!

PR #295, opened 2019-04-18
Added base spec for testing Spark wrapping transformers

tovbinm:
**Related issues**
Spark wrapping transformers were missing a test spec

1. Added `SwTransformerSpec` for testing Spark wrapping transformers
2. Factored out some common functionality between `OpTransformerSpec` and `SwTransformerSpec` into private `TransformerSpecCommon`
3. Updated `OpIndexToStringTest` to use it

### DIFF
--- a/features/src/main/scala/com/salesforce/op/test/OpTransformerSpec.scala
+++ b/features/src/main/scala/com/salesforce/op/test/OpTransformerSpec.scala
@@ -94,7 +94,7 @@ private[test] trait TransformerSpecCommon[O <: FeatureType, TransformerType <: O
   self: OpPipelineStageSpec[O, TransformerType] =>
 
   /**
-   * Transfotmer instance to be tested
+   * Transformer instance to be tested
    */
   val transformer: TransformerType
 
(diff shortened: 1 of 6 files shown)


features/src/main/scala/com/salesforce/op/test/OpTransformerSpec.scala
```diff
@@ -49,12 +49,52 @@ import scala.reflect.runtime.universe._
...
+ *
+ * @tparam O               output feature type
+ * @tparam TransformerType type of the transformer being tested
+ */
+private[test] trait TransformerSpecCommon[O <: FeatureType, TransformerType <: OpPipelineStage[O] with Transformer] {
+  self: OpPipelineStageSpec[O, TransformerType] =>
 
   /**
-   * [[OpTransformer]] instance to be tested
+   * Transfotmer instance to be tested
```
crupley:
```suggestion
   * Transformer instance to be tested
```

tovbinm:
> LGTM, I am just curious if checking the underlying type of the wrapper is really necessary.

@wsuchy maybe not as it should be checked by compiler. But I dont think there is anything wrong to check in a test as well.

PR #296, opened 2019-04-18
improved test SmartTextMapVectorizerTest

wsuchy:
Improved test by extending OpEstimatorSpec trait


core/src/test/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizerTest.scala
```diff
@@ -71,6 +73,26 @@ class SmartTextMapVectorizerTest extends FlatSpec with TestSparkContext with Att
     )
   )
 
+  /**
+   * Estimator instance to be tested
+   */
+  override val estimator: SmartTextMapVectorizer[TextMap] = new SmartTextMapVectorizer[TextMap]()
+    .setMaxCardinality(2).setNumFeatures(4).setMinSupport(1).setTopK(2).setPrependFeatureName(true)
```
tovbinm: let's perhaps test the stage with all the default settings, i.e. just `val estimator: SmartTextMapVectorizer[TextMap] = new SmartTextMapVectorizer[TextMap]()`

tovbinm: @wsuchy can you please address my comment so we can merge this?

PR #299, opened 2019-04-19
Spec-ifying ngram similarity tests

snabar:
Converting NGramSimilarityTest to use OpTransformerSpec


core/src/test/scala/com/salesforce/op/stages/impl/feature/TextNGramSimilarityTest.scala
```diff
@@ -32,31 +32,15 @@ package com.salesforce.op.stages.impl.feature
...
-      (Seq[String](), Seq("Red", "Yellow")),
-      (Seq[String](), Seq[String]()),
-      (Seq[String](""), Seq[String]("asdf")),
-      (Seq[String](""), Seq[String]("")),
-      (Seq[String]("", ""), Seq[String]("", ""))
-    ).map(v => v._1.toMultiPickList -> v._2.toMultiPickList)
-  )
-
-  val(dsText, f1Text, f2Text) = TestFeatureBuilder(
+class TextNGramSimilarityTest extends OpTransformerSpec[RealNN, NGramSimilarity[Text]]{
```
tovbinm: why not use the correct type here? 
```scala
class TextNGramSimilarityTest extends OpTransformerSpec[RealNN, TextNGramSimilarity]
```

and the same for the other test

PR #308, opened 2019-05-08
Aggregated LOCOs of SmartTextVectorizer outputs

michaelweilsalesforce:
**Related issues**
LOCOs of derived text features of smartTextVectorizer are hard to interpret (no indicator value). 

Aggregate all these LOCOs (except null indicator column) from the same text feature by the mean.

Further improvement PR : add other ways to aggregate (max, min, ...). This will become a param.


core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -105,6 +110,15 @@ class RecordInsightsLOCO[T <: Model[T]]
     diffs
   }
 
+  private def sumArray(left: Array[Double], right: Array[Double]): Array[Double] = {
```
gerashegalov: maybe consider something like `a.zipAll(b, 0, 0).map { case (l, r) => l + r }`

michaelweilsalesforce: Thanks

tovbinm:
@mweilsalesforce is it ready for review?

michaelweilsalesforce:
> @mweilsalesforce is it ready for review?
Yes, let's have a first round of review/feedback.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -93,8 +94,12 @@ class RecordInsightsLOCO[T <: Model[T]]
     case m => toOPUnchecked(m).transformFn
   }
   private val labelDummy = RealNN(0.0)
+  private lazy val vectorMetadata = OpVectorMetadata(getInputSchema()(in1.name))
+  private lazy val histories = vectorMetadata.getColumnHistory()
+  private lazy val featureInfo = histories.map(_.toJson(false))
 
-  private lazy val featureInfo = OpVectorMetadata(getInputSchema()(in1.name)).getColumnHistory().map(_.toJson(false))
+  private val textClassName = SmartTextVectorizer.getClass.getSimpleName.split("\\$").last
```
tovbinm: `val textClassName = classOf[SmartTextVectorizer[_]].getSimpleName

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -270,23 +274,99 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext {
...
 
     // Compare the ratio of importances between "important" and "other" features in both paradigms
-    assert(math.abs(avgRecordInsightRatio - featureImportanceRatio)*2 /
+    assert(math.abs(avgRecordInsightRatio - featureImportanceRatio) * 2 /
       (avgRecordInsightRatio + featureImportanceRatio) < 0.8,
       "The ratio of feature strengths between important and other features should be similar to the ratio of " +
         "feature importances from Spark's RandomForest")
   }
 
+  it should "aggregate text derived features" in {
```
tovbinm: @michaelweilsalesforce @Jauntbox do these test have to be that verbose & explicit? such large tests seem barely readable to comprehend.

michaelweilsalesforce: Some of the verbose come from the creation of the dataset. Not sure we need to move this part at the beginning of the class.

tovbinm: When lookin on a test - it has to be easy to understand what are the inputs, outcomes and assertions that are verified in the test; everything else have to elsewhere.

michaelweilsalesforce:
Oopsie. I forgot to add the same logic with SmartTextMapVectorizer. Working on it.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -93,8 +95,20 @@ class RecordInsightsLOCO[T <: Model[T]]
     case m => toOPUnchecked(m).transformFn
   }
   private val labelDummy = RealNN(0.0)
+  private lazy val vectorMetadata = OpVectorMetadata(getInputSchema()(in1.name))
+  private lazy val histories = vectorMetadata.getColumnHistory()
+  private lazy val featureInfo = histories.map(_.toJson(false))
+
+  /**
+   * These are the name of the stages we want to perform an aggregation of the LOCO results over ferived features
```
sanmitra: Typo: derived

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -93,8 +95,20 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+   */
+  private val smartTextClassName = classOf[SmartTextVectorizer[_]].getSimpleName
```
sanmitra: Can there be a custom text vectorizer, if yes shouldn't we handle that too ?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -135,29 +151,75 @@ class RecordInsightsLOCO[T <: Model[T]]
...
     val featureSize = featuresSparse.size
 
     val k = $(topK)
+    // Index where to examine the difference in the prediction vector
     val indexToExamine = baseScore.length match {
       case 0 => throw new RuntimeException("model does not produce scores for insights")
       case 1 => 0
       case 2 => 1
-      case n if (n > 2) => baseResult.prediction.toInt
+      case n if n > 2 => baseResult.prediction.toInt
```
tovbinm: I am not following - why prediction value becomes an index?!

tovbinm:
another question:
why in case of `TopKStrategy.PositiveNegative` we dont apply `take(k)`?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -135,65 +162,96 @@ class RecordInsightsLOCO[T <: Model[T]]
...
     }
-    val topPosNeg = returnTopPosNeg(filledSize, featureArray, featureSize, baseScore, k, indexToExamine)
+    val topPosNeg = returnTopPosNeg(featureArray, featureSize, baseScore, k, indexToExamine)
     val top = getTopKStrategy match {
       case TopKStrategy.Abs => topPosNeg.sortBy { case (_, v, _) => -math.abs(v) }.take(k)
       case TopKStrategy.PositiveNegative => topPosNeg.sortBy { case (_, v, _) => -v }
```
tovbinm: why in case of `TopKStrategy.PositiveNegative` we dont apply `take(k)`?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
michaelweilsalesforce: We want to return the LOCO score of the predicted class.

Let's say for a row the LOCOs are `0 -> LOCO_0, 1 -> LOCO_1, 2 -> LOCO_2`, and the model predicts the class `1` on this row. Then we want to return `LOCO_1`

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
michaelweilsalesforce: In `TopKStrategy.PositiveNegative`, we want to return `k` positives and `k` negatives.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
michaelweilsalesforce: Unfortunately custom Text Vectorizer cannot necessary return hashes as output. This aggregation is super limited and doesn't cover all the use cases yet.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
tovbinm: makes sense. so perhaps `take(2*k)`?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
tovbinm: let's add some docs to explain it, cause it looks weird to me.

tovbinm:
@Jauntbox please have a look

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: ok, thanks.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -93,107 +91,179 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+  /**
+   * These are the name of the stages we want to perform an aggregation of the LOCO results over derived features
+   */
+  private val smartTextClassName = classOf[SmartTextVectorizer[_]].getSimpleName
+  private val smartTextMapClassName = classOf[SmartTextMapVectorizer[_]].getSimpleName
+  // Indices of features derived from SmartText(Map)Vectorizer
+  private lazy val textFeatureIndices = histories
+    .filter(_.parentFeatureStages.exists(s => s.contains(smartTextClassName) || s.contains(smartTextMapClassName)))
+    .map(_.index)
+    .distinct.sorted
```
sanmitra: Do we need `distinct` here ?

michaelweilsalesforce: Added by @tovbinm. Not sure it is necessary but it seems safe.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -93,107 +91,179 @@ class RecordInsightsLOCO[T <: Model[T]]
...
-        if (positiveCount > k) { // remove the lowest element if the heap size goes from 5 to 6
-          positiveMaxHeap.dequeue()
+      // Let's check the indicator value and descriptor value
+      // If those values are empty, the field is likely to be a derived text feature (hashing tf output)
+      if (history.indicatorValue.isEmpty && history.descriptorValue.isEmpty) {
+        // Name of the field
+        val rawName = history.parentFeatureStages match {
+          case s if s.exists(_.contains(smartTextClassName)) => history.parentFeatureOrigins.headOption
+          case s if s.exists(_.contains(smartTextMapClassName)) => history.grouping
+          case _ => None
```
sanmitra: If the derived text feature is not from a SmartTextVectorizer then it is added neither to `minMaxHeap` nor `aggregationMap`. Basically its LOCO value is being ignored. I think that's incorrect we should add it to `minMaxHeap`.   @mweilsalesforce What do you think ?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
tovbinm: no need to keep duplicate indices, so added distinct. sorting was added for convenience.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
tovbinm: I would prefer to completely remove dependency of specific stage class and only decide based on the grouping & indicator values. is that even possible? (cc @Jauntbox )

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: Is there a scenario where there can be duplicates ? I don't think so, but yeah it seems safe to add.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
michaelweilsalesforce: I don't know the capabilities of TMOG regard Text engineering, but do you mean @sanmitra that any text derived feature with no indicator/descriptor value won't be able to be interpreted by LOCO (from other stages like OPLDA, OPWord2Vec, ...) ?

sanmitra: yes.

michaelweilsalesforce: Ok. The rule is now : "If a derived text feature has no indicator/descriptor value, then it will be aggregated. Aggregation by `parentFeatureName + parentFeatureStage`


Let me put the changes

tovbinm: @mweilsalesforce are you planning to do it in this PR?

michaelweilsalesforce: Done

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -143,17 +148,18 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+        val rawName = history.parentFeatureType match {
+          case s if s.exists(mapTypes.contains) => history.grouping
+          case s if s.exists(textTypes.contains) => history.parentFeatureOrigins.headOption
+          case s => throw new Error(s"type should be Text or TextMap, here ${s.mkString(",")}")
         }
         // Update the aggregation map
         for {name <- rawName} {
-          val (indices, array) = aggregationMap.getOrElse(name, (Array.empty[Int], Array.empty[Double]))
-          aggregationMap.update(name, (indices :+ i, sumArrays(array, diffToExamine)))
+          val key = name + "_" + history.parentFeatureStages.mkString(",")
```
tovbinm: the key value can be quite large is do concatenate all the parent stages. what's the rationale behind it? @mweilsalesforce

michaelweilsalesforce: It is possible to apply different transformations on a same feature.
After thought, maybe we should aggregate all the derived features even if 2 different transformations were applied

michaelweilsalesforce: Ok fixed

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
Jauntbox: Afaik, we can't currently have duplicate indices here but these are usually small (~few thousand element) arrays so no biggie.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -93,107 +90,183 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+      val history = histories(oldInd)
 
-      if (max > 0.0) { // if positive LOCO then add it to positive heap
-        positiveMaxHeap.enqueue((i, max, diffToExamine))
-        positiveCount += 1
-        if (positiveCount > k) { // remove the lowest element if the heap size goes from 5 to 6
-          positiveMaxHeap.dequeue()
+      // Let's check the indicator value and descriptor value
+      // If those values are empty, the field is likely to be a derived text feature (hashing tf output)
+      if (textFeatureIndices.contains(oldInd) && history.indicatorValue.isEmpty && history.descriptorValue.isEmpty) {
```
Jauntbox: Is this check here to make sure we don't accidentally pick up text features that were determined to be categorical by the smart text vectorizer? If so, we should probably make an easier way to tell what transformations were applied to the text.

michaelweilsalesforce: Not only smartVectorizer.  Any feature with indicator/descriptor values and derived from a text transformation is likely to be easily interpreted once LOCO is done.

PR #313, opened 2019-05-09
Tests performance improvements

tovbinm:
**Related issues**
Our tests became painfully slow

1. Reuse the same spark session across tests
2. Avoid setting checkpoint dirs

NA

** Additional context **
We need to investigate more is this really improving the performance @wsuchy

### DIFF
--- a/utils/src/main/scala/com/salesforce/op/test/TestSparkContext.scala
+++ b/utils/src/main/scala/com/salesforce/op/test/TestSparkContext.scala
@@ -20,10 +20,9 @@
 
 package com.salesforce.op.test
 
-import java.io.File
-
 import com.salesforce.op.utils.kryo.OpKryoRegistrator
 import org.apache.log4j.{Level, Logger}
+import org.apache.spark.serializer.KryoSerializer
 import org.apache.spark.sql.SparkSession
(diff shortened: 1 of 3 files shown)


utils/src/main/scala/com/salesforce/op/test/TestSparkContext.scala
```diff
@@ -60,22 +59,11 @@ trait TestSparkContext extends TempDirectoryTest with TestCommon {
...
 
-  lazy val checkpointDir: String = createDirectory(tempDir.getCanonicalPath, "checkpoints").toString
-
   override def beforeAll: Unit = {
     super[TempDirectoryTest].beforeAll()
-    sc.setCheckpointDir(checkpointDir)
+    val _ = sc // triggers initialization of spark session
   }
 
-  override def afterAll: Unit = {
```
wsuchy: Does it mean that we no longer want to cleanup temp directories created by spark?

tovbinm: It’s being cleared in the base trait

tovbinm:
We yet have to see how much of an improvement this change is.

tovbinm:
No improvement was observed.

Issue #315, opened 2019-05-13
Improved datetime unit transformer shortcuts


crupley:
**Problem**
It would be nice to make it easier to extract certain units from a date or datetime feature such as month, day of week, year, etc.

**Solution**
Add additional methods to `RichDate` and `RichDateTime`

**Alternatives**

**Additional context**

PR #316, opened 2019-05-13
Improved datetime unit transformer shortcuts

crupley:
**Related issues**
Improved datetime unit transformer shortcuts https://github.com/salesforce/TransmogrifAI/issues/315

Enhanced `RichDate` and `RichDateTime` to allow conversion to various time periods using a common class. Each now supports a `toTimePeriod` method which can convert to any of the current TimePeriods:

`TimePeriod.DayOfMonth`
`TimePeriod.DayOfWeek`
`TimePeriod.DayOfYear`
`TimePeriod.HourOfDay`
`TimePeriod.MonthOfYear`
`TimePeriod.WeekOfMonth`
`TimePeriod.WeekOfYear`

There are many other potential values that could be extracted from the datetime, but I stuck to only those defined in the `TimePeriod` enum.

**Additional context**


core/src/main/scala/com/salesforce/op/dsl/RichDateFeature.scala
```diff
@@ -183,3 +185,40 @@ trait RichDateFeature {
...
+  val f: FeatureLike[T]
+
+  /**
+   * Converts a Date or DateTime feature into a selected time period.
+   *
+   * @param period type of [[TimePeriod]] to convert date feature to
+   * @return Integer feature of time period value
+   */
+  def toTimePeriod(period: TimePeriod): FeatureLike[Integral] = {
+    val periodFun: Long => Int = period match {
```
tovbinm: Lets add a concrete transformer class that extends `UnaryTransformer` instead of using `UnaryLambdaTransformer`

core/src/main/scala/com/salesforce/op/dsl/RichDateFeature.scala
```diff
@@ -183,3 +185,40 @@ trait RichDateFeature {
...
+      }
+      case TimePeriod.WeekOfYear => t => new JDateTime(t, DateTimeUtils.DefaultTimeZone).weekOfWeekyear.get
+    }
+    f.transformWith(
+      new UnaryLambdaTransformer[T, Integral](operationName = "dateToTimePeriod",
+        transformFn = _.value.map(t => periodFun(t).toLong).toIntegral)
+    )
+  }
+
+  def toDayOfMonth(): FeatureLike[Integral] = toTimePeriod(TimePeriod.DayOfMonth)
```
tovbinm: Is there much value of having these methods? let's just have `def toTimePeriod(period: TimePeriod)` and that way you wont need this abstract class bu simply add `toTimePeriod` to both `RichDateTimeFeature` and `RichDateFeature`.

core/src/test/scala/com/salesforce/op/dsl/RichDateFeatureTest.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.salesforce.op.utils.date.DateTimeUtils
+import com.salesforce.op.utils.spark.RichDataset._
+import org.apache.spark.ml.Transformer
+import org.joda.time.{DateTime => JDateTime}
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class RichDateFeatureTest extends FlatSpec with TestSparkContext {
```
tovbinm: once you have a concrete transformer, please use `OpTransformerSpec` to test it

core/src/main/scala/com/salesforce/op/dsl/RichDateFeature.scala
crupley: I think so. As a user, I would much prefer to use these shortcuts and likely not use the time period shortcuts at all.

tovbinm: Once introduced, we would have to maintain them ;) and each time a new entry in enum is added we would need to add a new method. that's why I just prefer a single method with an enum argument.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TimePeriodTransformer.scala
```diff
@@ -0,0 +1,41 @@
+package com.salesforce.op.stages.impl.feature
```
tovbinm: please add license header

core/src/main/scala/com/salesforce/op/stages/impl/feature/TimePeriodTransformer.scala
```diff
@@ -0,0 +1,41 @@
...
+ * @tparam I            input feature type
+ */
+class TimePeriodTransformer[I <: Date]
+(
+  val period: TimePeriod,
+  uid: String = UID[TimePeriodTransformer[_]]
+)(
+  implicit override val tti: TypeTag[I]
+) extends UnaryTransformer[I, Integral](operationName = "dateToTimePeriod", uid = uid){
+  val periodFun: Long => Int = period match {
```
tovbinm: just make this a regular function `def periodConversion(time: Long): Int = period match { ... }`

tovbinm:
@crupley fyi, we also have `DateList`, `DateTimeList`, `DateMap` and `DateTimeMap` feature types. are you planning to add similar shortcuts to those in the following PRs?

crupley:
Thanks for pointing that out; yes, I can add that in a new PR

core/src/main/scala/com/salesforce/op/stages/impl/feature/TimePeriodTransformer.scala
```diff
@@ -0,0 +1,70 @@
...
+  implicit override val tti: TypeTag[I]
+) extends UnaryTransformer[I, Integral](operationName = "dateToTimePeriod", uid = uid){
+  def periodFun(t: Long): Int = period match {
+    case TimePeriod.DayOfMonth => new JDateTime(t, DateTimeUtils.DefaultTimeZone).dayOfMonth.get
+    case TimePeriod.DayOfWeek => new JDateTime(t, DateTimeUtils.DefaultTimeZone).dayOfWeek.get
+    case TimePeriod.DayOfYear => new JDateTime(t, DateTimeUtils.DefaultTimeZone).dayOfYear.get
+    case TimePeriod.HourOfDay => new JDateTime(t, DateTimeUtils.DefaultTimeZone).hourOfDay.get
+    case TimePeriod.MonthOfYear => new JDateTime(t, DateTimeUtils.DefaultTimeZone).monthOfYear.get
+    case TimePeriod.WeekOfMonth =>
+      val dt = new JDateTime(t, DateTimeUtils.DefaultTimeZone)
```
gerashegalov: `JDateTime(t, DateTimeUtils.DefaultTimeZone)` is repeated on every case. Let us pull it before the pattern match.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TimePeriodTransformer.scala
```diff
@@ -0,0 +1,70 @@
...
+ * @tparam I            input feature type
+ */
+class TimePeriodTransformer[I <: Date]
+(
+  val period: TimePeriod,
+  uid: String = UID[TimePeriodTransformer[_]]
+)(
+  implicit override val tti: TypeTag[I]
+) extends UnaryTransformer[I, Integral](operationName = "dateToTimePeriod", uid = uid){
+  def periodFun(t: Long): Int = period match {
```
gerashegalov: would it make sense to use OOP  and put 
```
sealed abstract class TimePeriod extends EnumEntry with Serializable {
  def periodFun(t: Long): Int
}

case object Blah extends TimePeriod {
    override def periodFun(t: Long): Int = ???
}
```
and call `period.periodFun` here

tovbinm: put logic into enum? I am not a big fan of this pattern.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TimePeriodTransformer.scala
crupley: Good idea; let me try it

core/src/main/scala/com/salesforce/op/stages/impl/feature/TimePeriodTransformer.scala
crupley: Good idea; let me try it

crupley: Eliminates the case matching or having to pass a trait around with the conversion function in it. What do you think of the implementation, @tovbinm?

features/src/main/scala/com/salesforce/op/stages/impl/feature/TimePeriod.scala
```diff
@@ -29,19 +29,27 @@
...
 object TimePeriod extends Enum[TimePeriod] {
   val values: Seq[TimePeriod] = findValues
-  case object DayOfMonth extends TimePeriod
-  case object DayOfWeek extends TimePeriod
-  case object DayOfYear extends TimePeriod
-  case object HourOfDay extends TimePeriod
-  case object MonthOfYear extends TimePeriod
-  case object WeekOfMonth extends TimePeriod
-  case object WeekOfYear extends TimePeriod
+  case object DayOfMonth extends TimePeriod { def extractFromTime(t: Long): Int = longToDateTime(t).dayOfMonth.get }
```
gerashegalov: I think it's easier to read if the function definition is after a line break.

core/src/test/scala/com/salesforce/op/stages/impl/feature/TimePeriodTransformerTest.scala
```diff
@@ -0,0 +1,83 @@
...
+    TimePeriod.values.foreach(tp => {
+      val expected = tp match {
+        case TimePeriod.DayOfMonth => Array(Integral(14), Integral(12), Integral(8), Integral.empty, Integral(30))
+        case TimePeriod.DayOfWeek => Array(Integral(5), Integral(6), Integral(1), Integral.empty, Integral(2))
+        case TimePeriod.DayOfYear => Array(Integral(73), Integral(316), Integral(67), Integral.empty, Integral(120))
+        case TimePeriod.HourOfDay => Array(Integral(0), Integral(10), Integral(12), Integral.empty, Integral(13))
+        case TimePeriod.MonthOfYear => Array(Integral(3), Integral(11), Integral(3), Integral.empty, Integral(4))
+        case TimePeriod.WeekOfMonth => Array(Integral(2), Integral(1), Integral(1), Integral.empty, Integral(4))
+        case TimePeriod.WeekOfYear => Array(Integral(11), Integral(45), Integral(10), Integral.empty, Integral(18))
+        case _ => throw new Exception(s"Unexpected TimePeriod encountered, $tp")
```
tovbinm: in tests instead of `throw new Exception(s"Unexpected TimePeriod encountered, $tp")` do `fail(s"Unexpected TimePeriod encountered, $tp")

PR #318, opened 2019-05-13
improved PercentileCalibratorTest

wsuchy:


core/src/test/scala/com/salesforce/op/stages/impl/feature/PercentileCalibratorTest.scala
```diff
@@ -49,10 +49,25 @@ import org.scalatest.FlatSpec
 import scala.util.Random
 
 @RunWith(classOf[JUnitRunner])
-class PercentileCalibratorTest extends FlatSpec with TestSparkContext {
+class PercentileCalibratorTest extends OpEstimatorSpec[RealNN, UnaryModel[RealNN, RealNN], PercentileCalibrator] {
+
   import spark.implicits._
 
-  Spec[PercentileCalibrator] should "return a minimum calibrated score of 0 and max of 99 when buckets is 100" in {
+  val testData = Seq(10, 100, 1000).map(_.toRealNN)
```
tovbinm: why not use a bit of a larger sample?

wsuchy: It is already being tested by 1000 elements sample later in that test, the major purpose of those values and extending of `OpEstimatorSpec` trait is for checking de/serializing into json, etc.

tovbinm: Base spec is supposed to test everything including transformations, serialization etc. Further tests usually test for additional behavior which is not covered in the base test. It makes sense to test as much as you can in the base test and avoid additional ones.

PR #320, opened 2019-05-15
 Serialize blacklisted map keys with the model + updated access on workflow/model members

tovbinm:
**Related issues**
We currently do not serialize backlisted map keys computed in RFF with the model.

1. Serialize blacklisted map keys with the model
2. Bonus: updated access on workflow/model members

NA


core/src/main/scala/com/salesforce/op/OpWorkflowCore.scala
```diff
@@ -161,6 +176,13 @@ private[op] trait OpWorkflowCore {
    */
   final def getStages(): Array[OPStage] = stages
 
+  /**
+   * Get the raw features generated by the workflow
+   *
+   * @return raw features for workflow
+   */
+  final def getRawFeatures(): Array[OPFeature] = rawFeatures
```
Jauntbox: woo! 👍

core/src/test/scala/com/salesforce/op/OpWorkflowModelReaderWriterTest.scala
```diff
@@ -97,13 +97,12 @@ class OpWorkflowModelReaderWriterTest
...
   def makeDummyModel(wf: OpWorkflow): OpWorkflowModel = {
-    val model = new OpWorkflowModel(wf.uid, wf.parameters)
-      .setStages(wf.stages)
-      .setFeatures(wf.resultFeatures)
-      .setParameters(wf.parameters)
+    new OpWorkflowModel(wf.uid, wf.getParameters())
+      .setStages(wf.getStages())
+      .setFeatures(wf.getResultFeatures())
+      .setParameters(wf.getParameters())
       .setRawFeatureFilterResults(rawFeatureFilterResults)
```
Jauntbox: Do we have a `getRawFeatureFilterResults()`?

PR #325, opened 2019-05-26
Add FeatureBuilder.fromSchema

takezoe:

This pull request adds `FeatureBuilder.fromSchema` that allows to create features from schema.

`FeatureBuilder` already has a method to create features from `DataFrame`, but actually, only schema is necessary to create features.


features/src/main/scala/com/salesforce/op/features/FeatureBuilder.scala
```diff
@@ -215,6 +216,7 @@ object FeatureBuilder {
     }
     responseFeature -> features
   }
+  def fromDataFrame[ResponseType <: FeatureType : WeakTypeTag](data: DataFrame, response: String, nonNullable: Set[String] = Set.empty): (Feature[ResponseType], Array[Feature[_ <: FeatureType]]) = fromSchema(data.schema, response, nonNullable)
```
tovbinm: Please add docs

takezoe: Added

Issue #184, opened 2018-11-22
Upgrade Apache Spark to 2.4


seratch:
As you may already know, Apache Spark 2.4.0 was released on Nov 3.

https://spark.apache.org/releases/spark-release-2-4-0.html

Are you already planning to upgrade Spark version? I am interested in working on the upgrade task.

tovbinm:
0.5.0 release is going to be with Spark 2.3.2.  We can start working on adding Spark 2.4.0 once we are done with the release.

seratch:
@tovbinm I understand it. Once Spark 2.4.1 is out, I will start working on it.

tovbinm:
@seratch since it might take some time to implement - let's start rolling adding spark 2.4.x support.

 Ideally it would be great to have a cross version build for Spark versions 2.3.x/2.4.x and Scala 2.11/2.12. For example: `com.salesforce.transmogrifai:transmogrifai-core_<scala-version>:<version>_<spark-version>`, i.e.:
 `com.salesforce.transmogrifai:transmogrifai-core_2.11:0.6.0_2.3.2`
 `com.salesforce.transmogrifai:transmogrifai-core_2.11:0.6.0_2.4.0`
 `com.salesforce.transmogrifai:transmogrifai-core_2.12:0.6.0_2.4.0`

WDYT?

seratch:
@tovbinm Sounds nice. It's also possible to have the `spark-version` in the artifact like `com.salesforce.transmogrifai:transmogrifai-core_spark-<spark-version>_<scala-version>:<version>`, i.e.:`transmogrifai-core_spark-2.3.2_2.11:0.6.0`

The idea is inspired by Scala.js bin versions. Scala.js libraries have `sjs0.6`in the middle of artifact names. https://search.maven.org/search?q=_sjs0.6_

tovbinm:
This plugin seems promising - https://github.com/ADTRAN/gradle-scala-multiversion-plugin (though after I tried it, scoverage plugin broke)

tovbinm:
This plugin seems promising - https://github.com/ADTRAN/gradle-scala-multiversion-plugin (though after I tried it, scoverage plugin broke):

And a few more options that I found:
1. https://github.com/prokod/gradle-crossbuild-scala
2. https://github.com/linkedin/scanns/tree/master/buildSrc/src/main/groovy/com/linkedin/nn/build/plugins

deanwampler:
[...]

PR #327, opened 2019-05-30
Update to Spark 2.4.3 + XGBoost 0.90 + MLeap 0.14

tovbinm:
**Related issues**
We would like to use XGBoost 0.90, but it requires Spark 2.4.x and also #184

- Update to Spark 2.4.3 and XGBoost 0.90
- Update XGBoost classifier and regressor params
- Update predictor models to use now public method `PredictionModel.predict` (since 2.4 - https://github.com/apache/spark/pull/19381)
- Update MLeap to 0.14
- Update Spark testing base
- Get rid of `spark-avro` dependency (since avro support is included with spark)
- Bump avro version to 1.8.2
- And a few more required changes

NA


wsuchy:
LGTM

features/src/main/scala/com/salesforce/op/utils/spark/RichDataset.scala
```diff
@@ -72,7 +71,7 @@ object RichDataset {
     val schemaStr = spark.sparkContext.textFile(schemaPath(path)).collect().mkString
     val schema = DataType.fromJson(schemaStr).asInstanceOf[StructType]
     val origNames = schema.fields.map(_.metadata.getString(OriginalNameMetaKey))
-    val data = spark.read.avro(dataPath(path)).toDF(origNames: _*)
+    val data = spark.read.format("avro").load(dataPath(path)).toDF(origNames: _*)
```
gerashegalov: nit: Can we leverage https://github.com/apache/spark/blob/master/external/avro/src/main/scala/org/apache/spark/sql/avro/AvroFileFormat.scala#L129 instead of hardcoding "avro" multiple times

tovbinm: Seems private. I will see what can be done.

koertkuipers:
https://issues.apache.org/jira/browse/SPARK-26133
OneHotEncoderEstimator got renamed to OneHotEncoder

leahmcguire:
LGTM

PR #329, opened 2019-06-05
Replace joda by java.time 

### LINKED ISSUE #287: Replace Joda with Java 8 native Date Time
**Problem**
Since Java 8 was introduced there is no need to use Joda as it has been replaced the native Date-Time API.

**Solution**
Ideally greping and replacing the text should work (mostly)

**Additional context**
Need to check if de/serializing will still work.

gerashegalov:
**Related issues**
Fixes #287 

WIP


tovbinm:
any updates? @gerashegalov

gerashegalov:
the prototyped approach may not be the best way. Maybe we'll first create a shim API for dates to be used in the internal pipeline.

PR #334, opened 2019-06-17
Added missing test for java conversions

wsuchy:


wsuchy:
My assumption was that this test should cover only the basic functionality (what it does by now). By looking at the coverage you can clearly see that all lines are now covered. I don't think that exploding the scope of it makes much sense as there are also different things to focus on.

tovbinm:
well, the code that was introduced in #333 now clearly has a problem which I wanted you discover through the test:
```scala
import collection.JavaConverters._
val jm = new java.util.HashMap[String, java.lang.Boolean]()
jm.put("nullValue", null)
jm.toBinaryMap // NullPointerException
```

tovbinm:
I propose to add a test for null values and modify the code from #333 with following:
```scala
def toBinaryMap: BinaryMap = new BinaryMap(
    Option(v).map(_.asScala.collect { case (k, v) if v != null => v.booleanValue() } .toMap).getOrElse(Map.empty)
)
```

wsuchy:
I am aware of that and assumed that you are fine with this by merging my PR.

tovbinm:
Clearly we need to fix it as we don't want to have any NPE flying around.

wsuchy:
Hmmm I wouldn't say that `null`s must be implicitly omitted. Why not convert them to false in Boolean and 0 in Integer?
Or maybe leave execptions as a suggestion for the user that there is a problem with the data.a

tovbinm:
Replacing with default values would be even worse. Let's have then explicit handling for null values as follows, e.g. for RealMap:
```scala
def toRealMap: RealMap = new RealMap(
    Option(v).map(_.asScala.map {
      case (k, null) => (k, null.asInstanceOf[Double])
      case (k, x) => (k, x.doubleValue())
    }.toMap).getOrElse(Map.empty)
)
```

PR #337, opened 2019-06-19
Forecast evaluator + SMAPE metric

wsuchy:
This is only a POC, as I would like to check if it gets (eventually) merged. 
So far I've implemented only one metric (sMAPE) but ideally I would like to put all forecast related metrics there (including some already present like MAE, etc). 

As we don't have any forecasting model yet I've used linear regression.


core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
```diff
@@ -176,6 +176,22 @@ object RegressionEvalMetrics extends Enum[RegressionEvalMetric] {
...
+sealed abstract class ForecastEvalMetric
+(
+  val sparkEntryName: String,
+  val humanFriendlyName: String
+) extends EvalMetric
+
+
+object ForecastEvalMetrics extends Enum[RegressionEvalMetric] {
+  val values: Seq[RegressionEvalMetric] = findValues
+  case object sMAPE extends RegressionEvalMetric("smape", "symmetric Mean Absolute Percentage Error")
```
tovbinm: lets rename to `SMAPE` and lowercase the human friendly name to `symmetric mean absolute percentage error` (as we do with all other metrics)

core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
```diff
@@ -176,6 +176,22 @@ object RegressionEvalMetrics extends Enum[RegressionEvalMetric] {
...
+ * Forecast Metrics
+ */
+sealed abstract class ForecastEvalMetric
+(
+  val sparkEntryName: String,
+  val humanFriendlyName: String
+) extends EvalMetric
+
+
+object ForecastEvalMetrics extends Enum[RegressionEvalMetric] {
```
tovbinm: `extends Enum[ForecastEvalMetric]`

core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
```diff
@@ -176,6 +176,22 @@ object RegressionEvalMetrics extends Enum[RegressionEvalMetric] {
...
+  val values: Seq[RegressionEvalMetric] = findValues
+  case object sMAPE extends RegressionEvalMetric("smape", "symmetric Mean Absolute Percentage Error")
+}
+
```
tovbinm: 1. please also add ForecastEvalMetrics to `EvalMetric.withNameInsensitive` function see Line 109
2. as well as add a new entry of `Forecast` to `OpEvaluatorNames`

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -0,0 +1,111 @@
...
+
+import com.salesforce.op.UID
+import com.salesforce.op.utils.spark.RichEvaluator._
+import org.apache.spark.ml.evaluation.RegressionEvaluator
+import org.apache.spark.sql.Dataset
+import org.slf4j.LoggerFactory
+
+/**
+ *
+ * Instance to evaluate Regression metrics
```
tovbinm: docs are incorrect

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -0,0 +1,111 @@
...
+ * Default evaluation returns Root Mean Squared Error
+ *
+ * @param name           name of default metric
+ * @param isLargerBetter is metric better if larger
+ * @param uid            uid for instance
+ */
+
+private[op] class OpForecastEvaluator
+(
+  override val name: EvalMetric = OpEvaluatorNames.Regression,
```
tovbinm: `override val name: EvalMetric = OpEvaluatorNames.Forecast`

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -0,0 +1,111 @@
...
+  @transient private lazy val log = LoggerFactory.getLogger(this.getClass)
+
+  def getDefaultMetric: ForecastMetrics => Double = _.sMAPE
+
+  override def evaluateAll(data: Dataset[_]): ForecastMetrics = {
+    val dataUse = makeDataToUse(data, getLabelCol)
+
+    val smape: Double = getSMAPE(dataUse, getLabelCol, getPredictionValueCol)
+    val metrics = ForecastMetrics(
+      sMAPE = smape
```
tovbinm: reformat into one liner

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  protected def getSMAPE(data: Dataset[_], labelCol: String, predictionValueCol: String): Double = {
+    data.select(labelCol, predictionValueCol).rdd
+      .map(r => ReduceSMAPE(r.getAs[Double](0), r.getAs[Double](1)))
+      .reduce(_ + _).metric
+  }
+}
+
+// https://www.m4.unic.ac.cy/wp-content/uploads/2018/03/M4-Competitors-Guide.pdf
+case class ReduceSMAPE(nominator: Double, denominator: Double, cnt: Long) {
```
tovbinm: instead of defining `+` you can use algebird macro for case classes, i.e.
```scala

implicit val smapeSG = caseclass.semigroup[SMAPEValue]
case class SMAPEValue(nominator: Double, denominator: Double, cnt: Long)
```

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -0,0 +1,111 @@
...
+    if (denominator == 0.0) {
+      Double.PositiveInfinity
+    } else {
+      (nominator / denominator) / cnt
+    }
+  }
+}
+
+object ReduceSMAPE {
+  def apply(y: Double, y_hat: Double): ReduceSMAPE = {
```
tovbinm: 1. please make this the only public ctor
2. avoid `_` in names, instead use camelCase

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -0,0 +1,111 @@
...
+    ReduceSMAPE(2 * Math.abs(y - y_hat), Math.abs(y) + Math.abs(y_hat), 1L)
+  }
+}
+
+/**
+ * Metrics of Regression Problem
```
tovbinm: invalid docs

PR #341, opened 2019-06-24
Aggregate spark metrics during run time instead of post processing by default

gerashegalov:
Compute by job metrics (sum of stage metrics) by default because it's O(1) non-GC'able memory

### DIFF
--- a/utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
+++ b/utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
@@ -33,6 +33,8 @@ package com.salesforce.op.utils.spark
 import com.salesforce.op.utils.date.DateTimeUtils
 import com.salesforce.op.utils.json.JsonLike
 import com.salesforce.op.utils.version.VersionInfo
+import com.twitter.algebird.Operators._
+import com.twitter.algebird.macros.caseclass
 import org.apache.spark.scheduler._
 import org.joda.time.Duration
 import org.joda.time.format.PeriodFormatterBuilder
@@ -102,7 +104,8 @@ class OpSparkListener
     if (collectStageMetrics) {
       stageMetrics += StageMetrics(si)
     } else {
-      cumulativeStageMetrics.plus(StageMetrics(si))
+      implicit val stageSG = caseclass.semigroup[StageMetrics]
+      cumulativeStageMetrics = cumulativeStageMetrics + StageMetrics(si)
     }
     if (logStageMetrics) {
       log.info("{},STAGE:{},MEMORY_SPILLED_BYTES:{},GC_TIME_MS:{},STAGE_TIME_MS:{}",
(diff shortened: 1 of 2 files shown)


utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -197,7 +202,46 @@ case class StageMetrics private
   shuffleWriteTime: Long,
   shuffleBytesWritten: Long,
   shuffleRecordsWritten: Long
-) extends JsonLike
+) extends JsonLike {
+  def plus(sm: StageMetrics): StageMetrics = {
```
tovbinm: instead of defining + you can use algebird macro for case classes, i.e.

```scala

implicit val stageSG = caseclass.semigroup[StageMetrics]
val sumStages = stage1 + stage2
```

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -252,4 +296,38 @@ object StageMetrics {
       shuffleRecordsWritten = tm.shuffleWriteMetrics.recordsWritten
     )
   }
+  def zero(): StageMetrics = StageMetrics(0,
+    0,
```
tovbinm: please ident it compactly. you can either use several lines or turn off the scalastyle on the line

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -91,14 +94,19 @@ class OpSparkListener
...
 
   override def onStageCompleted(stageCompleted: SparkListenerStageCompleted): Unit = {
     val si = stageCompleted.stageInfo
     val tm = si.taskMetrics
-    if (collectStageMetrics) stageMetrics += StageMetrics(si)
+    if (collectStageMetrics) {
+      stageMetrics += StageMetrics(si)
+    } else {
+      implicit val stageSG = caseclass.semigroup[StageMetrics]
+      cumulativeStageMetrics = cumulativeStageMetrics + StageMetrics(si)
```
gerashegalov: since you go this route you need to make some of the fields a dedicated Semigroup and override `plus` for it:
https://github.com/salesforce/TransmogrifAI/pull/341/files/3c25e7cf6581df8b29832b211acfb671b399d4f7#diff-862ee01af3e20d3ea939a0f9b89e362bR229

gerashegalov: see all fields that aren't `+`-ed

tovbinm: tests would help to reveal the truth ;)

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -197,7 +256,13 @@ case class StageMetrics private
   shuffleWriteTime: Long,
   shuffleBytesWritten: Long,
   shuffleRecordsWritten: Long
-) extends JsonLike
+) extends BaseStageMetrics with MetricJsonLike {
+  val duration: Option[Long] =
```
tovbinm: is `duration` is not serialized?

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -253,3 +317,91 @@ object StageMetrics {
     )
   }
 }
+
+case class CumulativeStageMetrics(
```
tovbinm: nit: ident, e.g.
```
case class CumulativeStageMetrics
(
  ...
)
```

utils/src/test/scala/com/salesforce/op/utils/spark/OpSparkListenerTest.scala
```diff
@@ -32,6 +32,9 @@ package com.salesforce.op.utils.spark
 
 import com.salesforce.op.test.TestSparkContext
 import com.salesforce.op.utils.date.DateTimeUtils
+import com.twitter.algebird.Max
+import com.twitter.algebird.Operators._
+import com.twitter.algebird.macros.caseclass
```
tovbinm: redundant import here

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -129,6 +143,23 @@ class OpSparkListener
 
 }
 
+trait MetricJsonLike extends JsonLike {
```
tovbinm: please add a test for this

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -129,6 +143,23 @@ class OpSparkListener
...
+  override def toJson(pretty: Boolean): String = {
+    JsonUtils.toJsonString(this, pretty = pretty, Seq(SerDes[Max[Long]](
+      classOf[Max[Long]],
+      new StdSerializer[Max[Long]](classOf[Max[Long]]) {
+        override def serialize(
+          value: Max[Long],
+          gen: JsonGenerator, provider: SerializerProvider
+        ): Unit = {
+          gen.writeNumber(value.get)
```
rachelwarren: do you need getOR else there? Can it be null?

utils/src/test/scala/com/salesforce/op/utils/spark/OpSparkListenerTest.scala
```diff
@@ -94,6 +100,71 @@ class OpSparkListenerTest extends FlatSpec with TableDrivenPropertyChecks with T
...
+      shuffleRemoteBlocksFetched = 1,
+      shuffleWriteTime = 1,
+      shuffleBytesWritten = 1,
+      shuffleRecordsWritten = 1
+    )
+
+    val total = Seq(sm0, sm1).foldLeft(CumulativeStageMetrics.zero)(_ + _)
+
+    total.peakExecutionMemory shouldBe Max(1001)
+    total.toJson(true).contains("\"peakExecutionMemory\" : 1001") shouldBe true
```
rachelwarren: add one more test case for non max addition serialization

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
gerashegalov: it was here in StageMetrics, but I forgot to add it to the Cumulative. Adding with tests

utils/src/test/scala/com/salesforce/op/utils/spark/OpSparkListenerTest.scala
gerashegalov: which one?

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
gerashegalov: see https://github.com/salesforce/TransmogrifAI/pull/341/files#diff-f6bf18b1d0ed98372298b89b11dfed41R170

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
gerashegalov: it's not an option. It's a function unwrapping `Long` from `Max[Long]` semigroup

utils/src/test/scala/com/salesforce/op/utils/spark/OpSparkListenerTest.scala
tovbinm: `com.twitter.algebird.macros.caseclass`

PR #342, opened 2019-06-24
Forecast Evaluator - fixes SMAPE, adds MASE and Seasonal Error metrics

wsuchy:
Fixed `SMAPE` and added `seasonal error` and `MASE`. 
Computation is done locally due to the way the seasonal error is being calculated.


core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
```diff
@@ -190,6 +190,8 @@ sealed abstract class ForecastEvalMetric
 object ForecastEvalMetrics extends Enum[ForecastEvalMetric] {
   val values: Seq[ForecastEvalMetric] = findValues
   case object SMAPE extends ForecastEvalMetric("smape", "symmetric mean absolute percentage error")
+  case object MASE extends ForecastEvalMetric("mase", "mean absolute scaled error ")
+  case object seasonalError extends ForecastEvalMetric("seasonalError", "seasonal error")
```
tovbinm: CamelCase the object name to `case object SeasonalError`

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -48,8 +51,10 @@ import com.twitter.algebird.macros.caseclass
  * @param isLargerBetter is metric better if larger
  * @param uid            uid for instance
  */
+
 private[op] class OpForecastEvaluator
 (
+  val seasonalWindow: Int = 1,
```
tovbinm: 1. please update the docs explaining what the `seasonalWindow` param represents, it's accepted range of values etc.
2. add `require` checks for `seasonalWindow` range in the initialization block of `OpForecastEvaluator` class

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -62,40 +67,105 @@ private[op] class OpForecastEvaluator
...
   }
 
-  protected def getSMAPE(data: Dataset[_], labelCol: String, predictionValueCol: String): Double = {
-    data.select(labelCol, predictionValueCol).rdd
-      .map(r => SMAPEValue(r.getAs[Double](0), r.getAs[Double](1)))
-      .reduce(_ + _).value
+
+  protected def getMetrics(data: Dataset[_], labelCol: String, predictionValueCol: String): ForecastMetrics = {
+    val res = data.select(labelCol, predictionValueCol).rdd
+      .map(r => MetricValue(r.getAs[Double](0), r.getAs[Double](1))).coalesce(1)
```
tovbinm: why `coalesce(1)`? it would be very slow to execute.

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -62,40 +67,105 @@ private[op] class OpForecastEvaluator
...
- */
-object SMAPEValue {
-  def apply(y: Double, yHat: Double): SMAPEValue = {
-    SMAPEValue(2 * Math.abs(y - yHat), Math.abs(y) + Math.abs(yHat), 1L)
+// https://www.m4.unic.ac.cy/wp-content/uploads/2018/03/M4-Competitors-Guide.pdf
+object MetricValue {
+  def apply(y: Double, yHat: Double): MetricValue = {
+    val absDiff = Math.abs(y - yHat)
+    val sumAbs = Math.abs(y) + Math.abs(yHat)
+    val smapeSum = if (sumAbs > 0) {
```
tovbinm: one line please to `val smapeSum = if (sumAbs > 0) absDiff / sumAbs else 0.0`

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -62,40 +67,105 @@ private[op] class OpForecastEvaluator
...
   }
-  implicit val smapeSG: Semigroup[SMAPEValue] = caseclass.semigroup[SMAPEValue]
+
+  implicit val smapeSG: Semigroup[MetricValue] = caseclass.semigroup[MetricValue]
 }
 
-case class SMAPEValue private (nominator: Double, denominator: Double, cnt: Long) {
-  def value: Double = if (denominator == 0.0) Double.NaN else (nominator / denominator) / cnt
+case class MetricValue private(absDiff: Double, smapeSum: Double, seasonalAbsDiff: SeasonalAbsDiff, cnt: Long) {
+  def smape: Double = {
```
tovbinm: one line here as well

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -62,40 +67,105 @@ private[op] class OpForecastEvaluator
...
+  def zero(seasonalWindow: Int): SeasonalAbsDiff =
+    SeasonalAbsDiff(seasonalWindow, 0, mutable.ListBuffer[Double](), 0.0)
+
+  implicit val seasonalAbsDiffSG: Semigroup[SeasonalAbsDiff] = new Semigroup[SeasonalAbsDiff] {
+    override def plus(x: SeasonalAbsDiff, y: SeasonalAbsDiff): SeasonalAbsDiff = x.combine(y)
+  }
 }
 
+case class SeasonalAbsDiff private
+(seasonalWindow: Int, lagsLength: Int, lags: mutable.ListBuffer[Double], absDiff: Double) {
```
tovbinm: why do you need to keep `lagsLength`? you can do `lags.length` instead.

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -62,40 +67,105 @@ private[op] class OpForecastEvaluator
...
+  def combine(that: SeasonalAbsDiff): SeasonalAbsDiff = {
+    var newLength = lagsLength + that.lagsLength
+    var l = this.lags
+    var newAbsDiff = this.absDiff + that.absDiff
+    var idx = seasonalWindow - lagsLength
```
tovbinm: can the index be negative here?
why not use foldLeft instead of vars?!

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -62,40 +67,105 @@ private[op] class OpForecastEvaluator
...
+      l = l.tail
+      newLength -= 1
+      idx += 1
+    }
+    SeasonalAbsDiff(seasonalWindow, newLength, l ++ that.lags, newAbsDiff)
+  }
+
+  def seasonalError(dataLength: Long): Double = {
+    val denominator = dataLength - seasonalWindow
+    if (denominator > 0) {
```
tovbinm: one line please

core/src/test/scala/com/salesforce/op/evaluators/OpForecastEvaluatorTest.scala
```diff
@@ -87,23 +89,76 @@ class OpForecastEvaluatorTest extends FlatSpec with TestSparkContext {
...
-    import spark.implicits._
+  it should "evaluate the metrics when data is empty" in {
+    val data = Seq().map(x => (x, Map("prediction" -> x)))
     val df = spark.sparkContext.parallelize(data).toDF("f1", "r1")
     val metrics = new OpForecastEvaluator().setLabelCol("f1").setPredictionCol("r1").evaluateAll(df).toMetadata()
     metrics.getDouble(ForecastEvalMetrics.SMAPE.toString).isNaN shouldBe true
   }
 
+
+  it should "verify that SeasonalAbsDiff works" in {
```
tovbinm: please provide a better test case description

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
wsuchy: @tovbinm that is the whole point of my questions in the description. You have to run the fold from the first result to the last. Without the coalesce it seems to apply fold per partition (or at least this it my guess given the observed behavior and wrong result)

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
wsuchy: In practice that loop will be iterated only once because we use it in `fold` rather than `reduce` context on spark. 
I am interested in index N and index N + M (where m = seasonal window) from the collection that would be created by concatenating already aggregated elements and new elements (in fact there is always only one element in the new collection due to the `fold`)

core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
wsuchy: It seems that if I camel case that object then I will have to camel case also field name in `ForecastMetrics` - is that expected?

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
wsuchy: I've just realized that ListBuffer::tail is O(N) and will have to replace that anyway.

tovbinm: Perhaps use mutable.ArraryBuffer?

core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
tovbinm: Hmm, no. you can show me tomorrow perhaps? 

The following should work just fine
```scala
case object SeasonalError extends ForecastEvalMetric("seasonalError", "seasonal error")
```

tovbinm:
if `SeasonalAbsDiff ` is not  associative, then how are you planning to reduce on it? Spark does not guarantee an order.

wsuchy:
> if `SeasonalAbsDiff ` is not associative, then how are you planning to reduce on it? Spark does not guarantee an order.

@tovbinm  Therefore I am using `fold` (with coealesce(1)) instead of `reduce` which makes me wonder if doing that math on spark makes still sense...

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -73,99 +81,50 @@ private[op] class OpForecastEvaluator
...
-case class MetricValue private(absDiff: Double, smapeSum: Double, seasonalAbsDiff: SeasonalAbsDiff, cnt: Long) {
-  def smape: Double = {
-    if (cnt == 0) {
-      Double.NaN
-    } else {
-      2 * smapeSum / cnt
+    val rows = data.select(labelCol, predictionValueCol).rdd
+      .map(r => (r.getAs[Double](0), r.getAs[Double](1))).take(maxItems)
+
+    val cnt = rows.length
```
tovbinm: is `rows.length` is the same as `rows.count`? let's use count, since it's easier to grep. thanks.

core/src/main/scala/com/salesforce/op/evaluators/OpForecastEvaluator.scala
```diff
@@ -73,99 +81,50 @@ private[op] class OpForecastEvaluator
...
+    val seasonalLimit = cnt - seasonalWindow
+
+    var i = 0
+    var seasonalAbsDiff = 0.0
+    var absDiffSum = 0.0
+    var smapeSum = 0.0
+
+    while (i < cnt) {
+      val r = rows(i)
```
tovbinm: `val (y, yHat) = rows(i)`

PR #345, opened 2019-06-25
Descale feature contribution for Linear Regression & Logistic Regression

TuanNguyen27:
**Problem context**
Spark returns feature contribution on the original scale of the feature, making it hard to compare relative importance between two features of different scales. During training for linear regression, Spark normalizes features and label to zero mean and unit variance, and we want to display feature contribution (which is the model's coefficient) on this normalized scale.  

Note: Spark does the same standardization for logistic regression, but it is unclear if the same problem applies to feature contribution for logistic regression. 

Descale feature contribution by multiplying Spark's returned weight with the respective feature standard deviation, and divide by the label standard deviation. This change only involves ModelInsights, and will not affect scoring.

N/A. The descaling needs access to the best trained model, label summary stats & feature summary stats.


tovbinm:
@TuanNguyen27 can you also please explain what problem are you solving in the PR description?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -647,6 +670,18 @@ case object ModelInsights {
     }
   }
 
+  private[op] def checkLRStandardization(model: Option[Model[_]]): Option[Boolean] = {
+    val stage = model.flatMap {
```
tovbinm:
```scala
for {
   m <- model
   stage <- m.getSparkMlStage()
} yield stage match {
    case s: LogisticRegressionModel | LinearRegressionModel => s.getStandardization
    case _ => false
}
```

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +558,25 @@ case object ModelInsights {
...
           val keptIndex = indexInToIndexKept.get(h.index)
+          val featureStd = getIfExists(h.index, s.featuresStatistics.variance).getOrElse(1.0)
+          val sparkFtrContrib = keptIndex
+            .map(i => contributions.map(_.applyOrElse(i, (_: Int) => 0.0))).getOrElse(Seq.empty)
+          val LRStandardization = checkLRStandardization(model).getOrElse(false)
+          val labelStd = label.distribution.getOrElse(1.0) match {
+            case Continuous(_, _, _, variance) => math.sqrt(variance)
+            // for (binary) logistic regression we only need to multiply by feature standard deviation
+            case Discrete(domain, prob) =>
+              def computeVariance(domain: Seq[String], prob: Seq[Double]): Double = {
```
tovbinm: you dont need the inner function and you can do it faster. simply do:
```scala
case Discrete(domain, prob) =>
   val (weighted, sqweighted) = (domain zip prob).foldLeft((0.0, 0.0)) { case ((weightSum, sqweightSum), (d, p)) =>
         val floatD = d.toDouble
         val weight = floatD * prob
         val sqweight = floatD * weight
         (weightSum+ weight, sqweightSum + sqweight)  
   }
   sqweighted - weighted
```

tovbinm: another question, since `domain` is `Seq[String]` how can we be sure that `d.toDouble` wont throw an error?! should we handle it gracefully?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +558,25 @@ case object ModelInsights {
             case _ => None
           }
           val keptIndex = indexInToIndexKept.get(h.index)
+          val featureStd = getIfExists(h.index, s.featuresStatistics.variance).getOrElse(1.0)
+          val sparkFtrContrib = keptIndex
+            .map(i => contributions.map(_.applyOrElse(i, (_: Int) => 0.0))).getOrElse(Seq.empty)
+          val LRStandardization = checkLRStandardization(model).getOrElse(false)
+          val labelStd = label.distribution.getOrElse(1.0) match {
```
tovbinm: dont return default value of a different type, instead properly handle None

```scala
val labelStd = label.distribution match {
       case Some(Continuous(_, _, _, variance)) =>
       case Some(Discrete(domain, prob)) =>
       case Some(d) => // handle this case - perhaps throw unsupported distribution type error?
       case None => // handle this case
}
```

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +558,25 @@ case object ModelInsights {
...
+                val floatDomain = domain.map(_.toDouble)
+                val sqfloatDomain = floatDomain.map(math.pow(_, 2))
+                val weighted = (floatDomain, prob).zipped map (_ * _)
+                val sqweighted = (sqfloatDomain, prob).zipped map (_ * _)
+                val mean = weighted.sum
+                return sqweighted.sum - mean
+              }
+              computeVariance(domain, prob)
+          }
+          // TODO: throw exception if (labelStd == 0)
```
tovbinm: please do error on it ;)

core/src/main/scala/com/salesforce/op/ModelInsights.scala
TuanNguyen27: `m.getSparkMlStage()` the compiler is not happy about this

core/src/main/scala/com/salesforce/op/ModelInsights.scala
TuanNguyen27: For regression problem where there are too few unique labels (and they get treated as categoricals), `d.toDouble` should work fine. I'm not sure how it will behave on classification. For classification I assume that `domain = Array("0", "1")`, is this correct ?

tovbinm: Why did we decide to define `domain` as `Seq[String]` in the first place? @Jauntbox @leahmcguire

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -562,22 +562,23 @@ case object ModelInsights {
...
+          val labelStd = label.distribution match {
+            case Some(Continuous(_, _, _, variance)) => math.sqrt(variance)
+            case Some(Discrete(domain, prob)) =>
+              val (weighted, sqweighted) = (domain zip prob).foldLeft((0.0, 0.0)) {
+              case ((weightSum, sqweightSum), (d, p)) =>
+                val floatD = d.toDouble
+                val weight = floatD * p
+                val sqweight = floatD * weight
+                (weightSum + weight, sqweightSum + sqweight)
+            }
```
tovbinm: ident this case block

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -562,22 +562,23 @@ case object ModelInsights {
...
+              sqweighted - weighted
+            case Some(d) => throw new Exception("Unsupported distribution type for the label")
```
tovbinm: we use RuntimeException elsewhere please do the same

core/src/main/scala/com/salesforce/op/ModelInsights.scala
TuanNguyen27: the compiler is not happy with `m.getSparkMlStage()`

tovbinm: either change the type of the function argument to `model: Option[Model[_] with SparkWrapperParams])` or add collect:
```scala
for {
   m <- model.collect { case v: SparkWrapperParams => v }
   stage <- m.getSparkMlStage()
} yield stage match {
    case s: LogisticRegressionModel | LinearRegressionModel => s.getStandardization
    case _ => false
}
```

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,7 +558,27 @@ case object ModelInsights {
...
+            case Some(Discrete(domain, prob)) =>
+              val (weighted, sqweighted) = (domain zip prob).foldLeft((0.0, 0.0)) {
+                case ((weightSum, sqweightSum), (d, p)) =>
+                  val floatD = d.toDouble
+                  val weight = floatD * p
+                  val sqweight = floatD * weight
+                  (weightSum + weight, sqweightSum + sqweight)
+              }
+              sqweighted - weighted
+            case Some(d) => throw new RuntimeException("Unsupported distribution type for the label")
```
leahmcguire: Some(_) or get the type off of D and include it in the error message

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,7 +558,27 @@ case object ModelInsights {
...
+            case None => throw new RuntimeException("Label does not exist, please check your data")
+          }
+          if (labelStd == 0) throw new RuntimeException("The standard deviation of the label is zero, " +
+            "so the coefficients and intercepts of the model will be zeros, training is not needed.\"")
```
leahmcguire: In general we have tried not to have the model insights throw exceptions - it is kind of a rude place to fail. Please have it log a warning and put in a graceful empty default value

core/src/main/scala/com/salesforce/op/ModelInsights.scala
leahmcguire: If you have the raw label rather than the indexed label you will get strings - this was designed to support that

TuanNguyen27: is it safe to do `toDouble` in this case ?

tovbinm: Any value which is not numeric with throw an exception, e.g. `"".toDouble`

TuanNguyen27: but will there ever be non numeric string in the label field...?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +560,42 @@ case object ModelInsights {
...
+            case Some(Discrete(domain, prob)) =>
+              val (weighted, sqweighted) = (domain zip prob).foldLeft((0.0, 0.0)) {
+                case ((weightSum, sqweightSum), (d, p)) =>
+                  val floatD = d.toDouble
+                  val weight = floatD * p
+                  val sqweight = floatD * weight
+                  (weightSum + weight, sqweightSum + sqweight)
+              }
+              if (sqweighted == weighted) {
+                log.info("The standard deviation of the label is zero, " +
```
leahmcguire: lets set the log level to warn on these

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +560,42 @@ case object ModelInsights {
...
+          val LRStandardization = checkLRStandardization(model).getOrElse(false)
+          val defaultLabelStd = 1.0
+          val labelStd = label.distribution match {
+            case Some(Continuous(_, _, _, variance)) =>
+              if (variance == 0) {
+                log.warn("The standard deviation of the label is zero, " +
+                  "so the coefficients and intercepts of the model will be zeros, training is not needed.\"")
+                defaultLabelStd
+              }
+              else math.sqrt(variance)
```
leahmcguire: are these formulas (here and on line 582) from the spark regularization? and if so does it work for all model types?

TuanNguyen27: this and line 582 are calculating standard deviation of the label (not related to spark reg), so this depends on whether the label is binary or continuous (i'm not sure about multi-class case), and it shouldn't depend on model type (unless it's multi-class classification).

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +560,49 @@ case object ModelInsights {
             case _ => None
           }
           val keptIndex = indexInToIndexKept.get(h.index)
+          val featureStd = getIfExists(h.index, s.featuresStatistics.variance).getOrElse(1.0)
```
leahmcguire: do you need to sqrt this?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -579,7 +625,10 @@ case object ModelInsights {
                 case _ => Map.empty[String, Double]
               },
               contribution =
-                keptIndex.map(i => contributions.map(_.applyOrElse(i, (_: Int) => 0.0))).getOrElse(Seq.empty),
+                if (LRStandardization && sparkFtrContrib.nonEmpty) {
+                  sparkFtrContrib.updated(0, sparkFtrContrib.head * featureStd / labelStd)
```
leahmcguire: why head? sparkFtrContrib.head is this for binary only? - I dont think as written it will effect binary only

core/src/main/scala/com/salesforce/op/ModelInsights.scala
TuanNguyen27: good catch, let me fix that

core/src/main/scala/com/salesforce/op/ModelInsights.scala
TuanNguyen27: i'm descaling the first coeff in `sparkFtrContrib` for each (derived) feature. This works for binary logistic regression & linear regression, but will break/ won't fully descale multi-nomial logistic regression for multi-class classification.

leahmcguire: shouldnt you scale the entire vector?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
TuanNguyen27: TODO: this part also needs to change if multi-class classification is considered for descaling

TuanNguyen27: Turns out for logistic regression, descaling only requires feature standard deviation, so this part stays the same.

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +560,48 @@ case object ModelInsights {
...
+              if (variance == 0) {
+                log.warn("The standard deviation of the label is zero, " +
+                  "so the coefficients and intercepts of the model will be zeros, training is not needed.\"")
+                defaultLabelStd
+              }
+              else math.sqrt(variance)
+            case Some(Discrete(domain, prob)) =>
+              // mean = sum (x_i * p_i)
+              val mean = (domain zip prob).foldLeft(0.0) {
+              case (weightSum, (d, p)) =>
```
leahmcguire: nit fix indent

leahmcguire: also I would find this a lot easier to read as a single line `case (weightSum, (d, p)) => (weightSum + d.toDouble * p)`

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -652,4 +695,19 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
...
+    // trained on unstandardized data and we can analytically calculate
+    // the scaled version of them by the linear regression formula, the coefficients
+    // of the model trained on standardized data should be within a small distance of the analytical formula.
+
+    // difference between the real coefficient and the analytical formula
+    val absError = math.abs(orginalbigCoeff * math.sqrt(smallFeatureVariance) / labelStd - descaledbigCoeff)
+    val absError2 = math.abs(originalsmallCoeff * math.sqrt(bigFeatureVariance) / labelStd - descaledsmallCoeff)
+    absError < 0.01 shouldBe true
+    absError2 < 0.01 shouldBe true
+  }
```
leahmcguire: please add a test for logistic regression as well

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -647,6 +693,32 @@ case object ModelInsights {
...
+    labelStd: Double): Option[Seq[Double]] = {
+    val stage = model.flatMap {
+      case m: SparkWrapperParams[_] => m.getSparkMlStage()
+      case _ => None
+    }
+    stage.collect {
+      case m: LogisticRegressionModel =>
+        if (m.getStandardization && sparkFtrContrib.nonEmpty) {
+          // scale entire feature contribution vector
+          sparkFtrContrib.map(_ * featureStd)
```
leahmcguire: can you explain the logic to me here? the spark coefficients are on the actual scale of the data. If a feature has a large standard deviation that means it has a large range, if it has a small standard deviation it has a small range. Shouldn't you be dividing by the standard deviation to bring them all into a range that can be compared? 
Is there something I am missing?

TuanNguyen27: It's a bit counter-intuitive, and only works for certain cases. For regression let's take `y = 1 mil * x_1 + 1 * x_2`, where `x_1` is small scale and `x_2` is large scale (millions), and due to the formula `y` will be on the same large scale as `x_2`. 

Spark in this case will return (1 mil, 1) but instead we want roughly both coeff to be between 0 and 1. The small feature have a small range, so 1mil * std(small feature) / std(y -- large) will bring it back to a small scale, because it is multiplied by a small number and divided by a large number. The large feature 1 * std (large feature) / std(y -- large) will stay roughly the same, because it is multiplied by a large number and divided by a large number. 

The same math is harder to reason for the classification case, because y is either 0, 1 and we are regressing on odds. p (probability y = 1) = 1 / 1 + e^ -( 1 mil * x_1 + 1 * x_2) then spark is still returning (1mil, 1) and by descaling, we instead returns (1, 1 mil) ?

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
TuanNguyen27: I just checked again and I couldn't find an example where the feature contributions are on wildly different scales for logistic regression, so i will remove that case from the descaling logic, no test needed for logistic regression case.

core/src/main/scala/com/salesforce/op/ModelInsights.scala
leahmcguire: Can you add a citation as a comment for future reference?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -647,6 +687,34 @@ case object ModelInsights {
...
+//        if (m.getStandardization && sparkFtrContrib.nonEmpty) {
+//          // scale entire feature contribution vector
+//          sparkFtrContrib.map(_ * featureStd)
+//        }
+//        else sparkFtrContrib
+      case m: LinearRegressionModel =>
+        if (m.getStandardization && sparkFtrContrib.nonEmpty) {
+          // need to also divide by labelStd for linear regression
+          // See https://github.com/salesforce/TransmogrifAI/pull/345#discussion_r303634114
+          // Also see https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala#L551-L558
```
leahmcguire: I meant a paper citation. How did you arrive at this formula? @Jauntbox any comments?

TuanNguyen27: I reverse engineered Spark's implementation (link to their implementation is included in the code comment).

TuanNguyen27: Will add a formal paper citation if i can find one

TuanNguyen27: In formal literature, this is referred to as standardized (regression) coefficients, or beta coefficients. See https://en.wikipedia.org/wiki/Standardized_coefficient. The two citations in the Wiki page are not open access, would you still want me to include those? Informal citations are https://www3.nd.edu/~rwilliam/stats1/x92.pdf or https://u.demog.berkeley.edu/~andrew/teaching/standard_coeff.pdf

For standardizing coeffs in logistic regression, see this blog: https://think-lab.github.io/d/205/. Tldr, they suggest using the Agresti approach (what I implemented, but currently commented out) -- which is multiplying the coeff by the respective standard deviation of the feature.

Jauntbox: Yeah, I just double-checked the math. If you do the zero mean and unit std dev rescaling, then both the features and label should scale as x' = (x - mean(x))/std(x).

If you plug the scaled values into the linear regression equation, then the weights in the unscaled space are weight_unscaled(i) = std(label) * weight_scaled(i) / std(i)

So the weights in the scaled space are weight_scaled(i) = weight_unscaled(i) * std(i) / std(label), which is what @TuanNguyen27 has. Is there a specific example you're worried this is incorrect for @leahmcguire?

leahmcguire: Ok please add those citations @TuanNguyen27. And thanks for double checking @TuanNguyen27 and @Jauntbox!

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -557,6 +560,42 @@ case object ModelInsights {
...
+                case (sqweightSum, (d, p)) => sqweightSum + (d.toDouble - mean) * (d.toDouble - mean) * p
+              }
+              if (discreteVariance == 0) {
+                log.warn("The standard deviation of the label is zero, " +
+                  "so the coefficients and intercepts of the model will be zeros, training is not needed.\"")
+                defaultLabelStd
+              }
+              else math.sqrt(discreteVariance)
+            case Some(_) => {
+              log.warn("Performing weight descaling on an unsupported distribution")
```
leahmcguire: Failing to perform weight descaling because distribution is unsupported

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -651,4 +722,40 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
...
+    // difference between the real coefficient and the analytical formula
+    //     return Array(descaledsmallCoeff, originalsmallCoeff, descaledbigCoeff, orginalbigCoeff)
+    val coeffs = getFeatureImp(standardizedLinpred, unstandardizedLinpred, linRegDF._3)
+    val descaledsmallCoeff = coeffs(0)
+    val originalsmallCoeff = coeffs(1)
+    val descaledbigCoeff = coeffs(2)
+    val orginalbigCoeff = coeffs(3)
+    val absError = math.abs(orginalbigCoeff * math.sqrt(smallFeatureVariance) / labelStd - descaledbigCoeff)
+    val absError2 = math.abs(originalsmallCoeff * math.sqrt(bigFeatureVariance) / labelStd - descaledsmallCoeff)
+    absError < 0.01 shouldBe true
```
Jauntbox: Can you make these tolerances a single value for the whole test class?

Also, if you're comparing a floating point calculation like this, it's usually easier to do a relative comparison. You can often use smaller tolerances if you do that too.
For example, if you have a calculation result `res` that you think should be equal to a ground truth value `truth`, then you can compare something like `abs(res - truth) * 2 / (res + truth) < tolerance` which is just the difference over the average value. If you compare the differences directly, then you often have to use different tolerances for different comparisons (like you do in the next test).

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -651,4 +722,40 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
...
+    val descaledsmallCoeff = coeffs(0)
+    val originalsmallCoeff = coeffs(1)
+    val descaledbigCoeff = coeffs(2)
+    val orginalbigCoeff = coeffs(3)
+    // difference between the real coefficient and the analytical formula
+    val absError = math.abs(orginalbigCoeff * math.sqrt(smallFeatureVariance) - descaledbigCoeff)
+    val absError2 = math.abs(originalsmallCoeff * math.sqrt(mediumFeatureVariance) - descaledsmallCoeff)
+    println(descaledsmallCoeff, originalsmallCoeff, descaledbigCoeff, orginalbigCoeff, absError, absError2)
+    absError < 0.2 shouldBe true
+    absError2 < 0.01 shouldBe true
```
Jauntbox: If you use a relative comparison instead of an absolute one, can you use the same tolerance for both?

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
Jauntbox: Oh, I meant making a single variable like `val tol = 0.05` so that we can adjust them all at once if this ends up being flaky in the future.

Jauntbox: Also, can you check what the average difference is? I would guess you can go much lower than your absolute difference tolerance now that you've converted them to relative differences.

PR #346, opened 2019-06-25
Quick Fix Alias Type Names

michaelweilsalesforce:
**Related issues**
When aliasing Type, the alias is sometimes lost in the reader.

Quick fix

### DIFF
--- a/readers/src/main/scala/com/salesforce/op/readers/Reader.scala
+++ b/readers/src/main/scala/com/salesforce/op/readers/Reader.scala
@@ -51,7 +51,7 @@ private[readers] trait ReaderType[T] extends Serializable {
    *
    * @return full input type name
    */
-  final def fullTypeName: String = wtt.tpe.typeSymbol.fullName
+  final def fullTypeName: String = wtt.tpe.toString
 
   /**
    * Short reader input type name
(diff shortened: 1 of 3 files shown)


readers/src/main/scala/com/salesforce/op/readers/Reader.scala
```diff
@@ -66,7 +66,8 @@ private[readers] trait ReaderType[T] extends Serializable {
    * @param opParams contains map of reader type to ReaderParams instances
    * @return ReaderParams instance if it exists
    */
-  final def getReaderParams(opParams: OpParams): Option[ReaderParams] = opParams.readerParams.get(this.typeName)
+  final def getReaderParams(opParams: OpParams): Option[ReaderParams] = opParams.readerParams
+    .get(this.wtt.tpe.toString.split("\\.").last)
```
tovbinm: This seems invalid, since it will only use the type name without package prefix.

michaelweilsalesforce: Yes that's what we want. Example of reader param is `Map{"entity1" -> ...}`

tovbinm: well it does match the check in the `readerInputTypes` then ;) I propose to use the full aliased name then `tpe.toString`

michaelweilsalesforce: That would involve changing all our OpParams...

PR #348, opened 2019-06-28
Adding `key` ctor field in all RawFeatureFilter results

michaelweilsalesforce:
**Related issues**
`FeatureDistribution` contains the field `key`   which is the map key associated with distribution (when the feature is a map). But why not having the same field in `RawFeatureFilterMetrics` and `ExclusionReasons`?

Adding the same field in `RawFeatureFilterMetrics` and `ExclusionReasons`

**Additional context**

### DIFF
--- a/core/src/main/scala/com/salesforce/op/filters/RawFeatureFilterResults.scala
+++ b/core/src/main/scala/com/salesforce/op/filters/RawFeatureFilterResults.scala
@@ -125,6 +125,7 @@ object RawFeatureFilterConfig extends RawFeatureFilterFormats {
  * Contains raw feature metrics computing in Raw Feature Filter
  *
  * @param name                          feature name
+ * @param key          map key associated with distribution (when the feature is a map)
  * @param trainingFillRate              proportion of values that are null in the training distribution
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/salesforce/op/filters/RawFeatureFilterResults.scala
```diff
@@ -125,6 +125,7 @@ object RawFeatureFilterConfig extends RawFeatureFilterFormats {
  * Contains raw feature metrics computing in Raw Feature Filter
  *
  * @param name                          feature name
+ * @param key          map key associated with distribution (when the feature is a map)
```
tovbinm: Ident

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilterResults.scala
```diff
@@ -135,6 +136,7 @@ object RawFeatureFilterConfig extends RawFeatureFilterFormats {
 case class RawFeatureFilterMetrics
 (
   name: String,
+  key: Option[String],
```
tovbinm: Docs please

PR #349, opened 2019-06-28
Aggregate LOCOs of DateToUnitCircleTransformer.

sanmitra:
**Related issues**
Aggregated LOCOs of DateToUnitCircleTransformer outputs.

For each time period we aggregate all the LOCOs from the same date feature by the mean.


core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -161,7 +192,28 @@ class RecordInsightsLOCO[T <: Model[T]]
           val (indices, array) = aggregationMap.getOrElse(key, (Array.empty[Int], Array.empty[Double]))
           aggregationMap.update(key, (indices :+ i, sumArrays(array, diffToExamine)))
         }
-      } else {
+      }
+      // Let's check the descriptor value
+      // If those values exist, the field is likely to be a derived date feature from unit circle transformer.
+      else if (dateFeatureIndices.contains(oldInd) && history.descriptorValue.isDefined) {
+        // Name of the field
+        val rawName: Option[String] = history.parentFeatureType match {
```
michaelweilsalesforce: There is some code duplication. Can we avoid that?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -466,4 +467,156 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext {
...
+
+    implicit val enc: Encoder[(Array[Double], Long)] = ExpressionEncoder()
+    implicit val enc2: Encoder[Seq[Double]] = ExpressionEncoder()
+
+    /**
+     * Compare the aggregation made by RecordInsightsLOCO to one made manually
+     *
+     * @param predicate  predicate used by RecordInsights in order to aggregate
+     */
+    def assertAggregatedWithPredicate(predicate: OpVectorColumnHistory => Boolean): Unit = {
```
michaelweilsalesforce: Same here. Can we reuse the code in the other test to avoid duplication?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: sure

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
sanmitra: sure

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: done.

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
sanmitra: done.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -127,6 +150,29 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+   * Optionally convert columnMetadata's descriptorValue like
+   * "y_DayOfWeek", "x_DayOfWeek" to TimePeriod enum - DayOfWeek.
+   * @return Option[TimePeriod]
+   */
+  private def convertToTimePeriod(descriptorValue: String): Option[TimePeriod] =
+    descriptorValue.split("_").lastOption.flatMap(TimePeriod.withNameInsensitiveOption)
+
+  private def isMapFeature(featureType: String ): Boolean = {
+    val featureTypeTag = FeatureType.featureTypeTag(featureType)
+    featureTypeTag.tpe <:< weakTypeOf[OPMap[_]]
```
tovbinm: this check is very very very slow to perform at runtime. let's just check if `history.grouping` is present or not. ie. 

```scala
private def getRawFeatureName(history: OpVectorColumnHistory): Option[String] = history.grouping match {
    case Some(grouping) => history.parentFeatureOrigins.headOption.map(_ + "_" + grouping)
    case None => history.parentFeatureOrigins.headOption
}
```

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -143,25 +189,28 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+      val isHashTextFeature = textFeatureIndices.contains(oldInd) &&
+        history.indicatorValue.isEmpty && history.descriptorValue.isEmpty
+      // If the descriptor value exists, the field is likely to be a derived date feature from unit circle transformer.
+      val isUnitCircleDateFeature = dateFeatureIndices.contains(oldInd) && history.descriptorValue.isDefined
+
+      if (isHashTextFeature || isUnitCircleDateFeature) {
+        for {name <- getRawFeatureName(history)} {
+          // Update the aggregation map for each (rawFeatureName, timePeriod) in case of date features.
+          val key = if (isUnitCircleDateFeature) {
+            name + "_" + history.descriptorValue.flatMap(convertToTimePeriod).map(_.entryName).getOrElse("")
```
tovbinm: let's replace `isUnitCircleDateFeature` val  and `if / else`  with pattern matching

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -466,4 +427,167 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext {
     assertAggregatedTextMap(textAreaMap, "k1")
 
   }
+  it should "aggregate values for date, datetime, dateMap and dateTimeMap derived features" in {
```
tovbinm: sorry, I am not able to follow these tests anymore.

Can we please make them more readable by either splitting them into smaller test cases or rather create a behavior (http://www.scalatest.org/user_guide/sharing_tests), or at least add clues `withClue` - http://www.scalatest.org/user_guide/using_assertions

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: pattern match on what exactly ? `history` ?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
sanmitra: sure, will refractor.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: btw, isUnitCircleDateFeature is referred multiple times, hence created a val.

tovbinm: yes, you can probably pattern match on history and it should make the code more readable as you would have clearly separated cases. give it a try.

tovbinm: and remember that you can pattern match together with `if` conditions.

sanmitra:
@tovbinm please take a look at the refactoring of test. Thanks.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: done.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -143,26 +179,23 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+        case h if h.indicatorValue.isEmpty && h.descriptorValue.isEmpty && textFeatureIndices.contains(oldInd) =>
+          for {name <- getRawFeatureName(h)} {
+            val (indices, array) = aggregationMap.getOrElse(name, (Array.empty[Int], Array.empty[Double]))
+            aggregationMap.update(name, (indices :+ i, sumArrays(array, diffToExamine)))
+          }
+        // If the descriptor value of a derived date feature exists, then it is likely to be
+        // from unit circle transformer. We aggregate such features for each (rawFeatureName, timePeriod).
+        case h if h.descriptorValue.isDefined && dateFeatureIndices.contains(oldInd) =>
+          for {name <- getRawFeatureName(h)} {
+            val key = name + "_" + h.descriptorValue.flatMap(convertToTimePeriod).map(_.entryName).getOrElse("")
```
tovbinm: so we can end up with key value `name_` here, I dont think this is what we want. instead we need to have simply `name`.

sanmitra: It shouldn't matter as long as the LOCOS of same parent features are merged for dates/texts. But it is much cleaner if its just `name` instead of `name_` in case convertToTimePeriod is empty. Let me change that.

PR #354, opened 2019-07-03
Track mean & standard deviation of text length as a metric for text feature

TuanNguyen27:
**Problem context**
If not treated as a categorical, text features are tokenized and hashed during feature engineering. However, things like IDs, dates, geographical information, etc should be treated differently. Even when hashing is the right approach, the current default hash space of TransmogrifAI is too small to capture all the information in the text. To better detect what is contained in a text field and dynamically determine an appropriate hash space, we want to track the mean and standard deviation of the string length of a text feature. 

Mean and Std of text length is computed inside `RawFeatureFilter` and will be part of `FeatureDistribution`.

N/A. `RawFeatureFilter` is the appropriate place to track this information, because similar calculations (e.g distribution of tokens) also happen here, and the additional information about text length could help inform these other calculations to remove raw features more intelligently.


core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -524,12 +524,14 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
...
     trainingReader: Option[Reader[T]],
     scoringReader: Option[Reader[T]],
-    bins: Int = 100,
+    bins: Int = 500,
     minFillRate: Double = 0.001,
     maxFillDifference: Double = 0.90,
     maxFillRatioDiff: Double = 20.0,
     maxJSDivergence: Double = 0.90,
     maxCorrelation: Double = 0.95,
+    pvalCutoff: Double = 0.05,
```
tovbinm: docs please

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -91,6 +95,23 @@ case class FeatureDistribution
...
+   * Test whether the given distribution is Uniform, for detecting useless text hashes
+   *
+   * @return true means we don't have enough evidence to reject Null hypothesis (current distribution is uniform)
+   *         likely to drop the feature, but will combine with average text length check.
+   *         If the hash space is too small w.r.t, a text feature could still appear uniformly distributed
+   *
+   *         false means rejecting the Null hypothesis
+   *         hashed feature deos not follow uniform distribution, but could still be useless
+   */
+
```
tovbinm: redundant endline

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -91,6 +95,23 @@ case class FeatureDistribution
...
+  def chiSqUnifTest(cutoff: Double): Boolean = {
+    val vectorizedDistr = Vectors.dense(distribution)
+    val goodnessOfFitTestResult = Statistics.chiSqTest(vectorizedDistr)
+    return goodnessOfFitTestResult.pValue >= cutoff
```
tovbinm: not need in return statement

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -100,9 +121,11 @@ case class FeatureDistribution
   def reduce(fd: FeatureDistribution): FeatureDistribution = {
     checkMatch(fd)
     val combinedDist = distribution + fd.distribution
+    val combinedAvgTextLen = (avgTextLen * count + fd.avgTextLen * fd.count)/(count + fd.count)
```
tovbinm: add a space between brackets `) / (`

core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala
```diff
@@ -97,6 +97,8 @@ class RawFeatureFilter[T]
   val maxFillRatioDiff: Double,
   val maxJSDivergence: Double,
   val maxCorrelation: Double,
+  val pvalCutoff: Double,
```
tovbinm: docs here as well

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -91,6 +95,24 @@ case class FeatureDistribution
    */
   def fillRate(): Double = if (count == 0L) 0.0 else (count - nulls) / count.toDouble
 
+  /**
+   * Test whether the given distribution is Uniform, for detecting useless text hashes
```
tovbinm: is this documentation block related to `chiSqUnifTest`?

tovbinm: if so, there is a redundant line between the docs block and the function

TuanNguyen27: fixed!

core/src/main/scala/com/salesforce/op/filters/AllFeatureInformation.scala
```diff
@@ -42,6 +42,7 @@ package com.salesforce.op.filters
  *                               1st level keys correspond to response keys
  *                               2nd level keys correspond to predictor keys with values being
  *                               null-label leakage corr. value
+ * @param avgtextLength          average length of text features
```
tovbinm: hmm, I dont see `avgtextLength` passed into `AllFeatureInformation`. is this a mistake?

TuanNguyen27: Good catch! It's not in there, i'll remove it

core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -59,7 +80,10 @@ case object Summary {
    */
   def apply(preppedFeature: ProcessedSeq): Summary = {
     preppedFeature match {
-      case Left(v) => Summary(v.size, v.size, v.size, 1.0)
+      case Left(v) =>
+        val textLenMoments = MomentsGroup.sum(v.map(x => Moments(x.length.toDouble)))
+        val tokenDistribution = TextStats(v.groupBy(identity).mapValues(_.size))
```
Jauntbox: This looks weird to me, why not just `v.map(x => x -> x.length).toMap`?

Jauntbox: Oh nm, you want the counts of each token - I got it.

leahmcguire: what if there are repeats? but also you should not do a mapValues. So 
`v.groupBy(identity).map( t => t -> t.size )`

core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -39,18 +40,38 @@ import com.twitter.algebird.Monoid
  * @param max   maximum value seen for double, maximum number of tokens in one text for text
  * @param sum   sum of values for double, total number of tokens for text
  * @param count number of doubles for double, number of texts for text
+ * @param maxCardinality maximum number of unique tokens to keep track of, for a given text feature
```
Jauntbox: Can you add docs for `textLength` and `textCard`?

core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -39,18 +40,38 @@ import com.twitter.algebird.Monoid
...
  */
-case class Summary(min: Double, max: Double, sum: Double, count: Double)
+case class Summary(min: Double, max: Double, sum: Double, count: Double,
+                   textLength: Option[Moments] = None,
+                   textCard: Option[TextStats] = None,
+                   maxCardinality: Int = 500)
```
Jauntbox: Can we put this value in the `RawFeatureFilter` companion object and pull it from there to keep the concrete thresholds are in one place?

core/src/main/scala/com/salesforce/op/filters/Summary.scala
TuanNguyen27: I think `groupBy(identity)` takes care of the repeats.

core/src/main/scala/com/salesforce/op/filters/Summary.scala
TuanNguyen27: Noted, will move this to a more sensible place once I confirm that the metrics work correctly.

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -63,6 +63,8 @@ case class FeatureDistribution
   nulls: Long,
   distribution: Array[Double],
   summaryInfo: Array[Double],
+  moments: Option[Moments] = None,
+  cardEstimate: Option[TextStats] = None,
```
leahmcguire: Are these optional for backwards compatibility of model insights?

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -99,10 +101,25 @@ case class FeatureDistribution
...
     // summary info can be empty or min max if hist is empty but should otherwise match so take the longest info
-    val combinedSummary = if (summaryInfo.length > fd.summaryInfo.length) summaryInfo else fd.summaryInfo
-    FeatureDistribution(name, key, count + fd.count, nulls + fd.nulls, combinedDist, combinedSummary, `type`)
+    val combinedSummaryInfo = if (summaryInfo.length > fd.summaryInfo.length) summaryInfo else fd.summaryInfo
+    val combinedMoments = (moments, fd.moments) match {
+      case (Some(x), None) => Some(x)
+      case (Some(x), Some(y)) => Some(x + y)
+      case (None, Some(y)) => Some(y)
+      case (_, _) => None
+    }
```
leahmcguire: do you really need to write all this out when you have operators imported? I think this should be the default behavior with operators.

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -99,10 +101,25 @@ case class FeatureDistribution
...
+    val combinedCard = (cardEstimate, fd.cardEstimate) match {
+      case (Some(x), None) => Some(x)
+      case (Some(x), Some(y)) => Some(testStatsSG.plus(x, y))
+      case (None, Some(y)) => Some(y)
+      case (_, _) => None
+    }
```
leahmcguire: same

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -155,24 +172,29 @@ case class FeatureDistribution
       "count" -> count.toString,
       "nulls" -> nulls.toString,
       "distribution" -> distribution.mkString("[", ",", "]"),
-      "summaryInfo" -> summaryInfo.mkString("[", ",", "]")
+      "summaryInfo" -> summaryInfo.mkString("[", ",", "]"),
+      "cardinality" -> cardEstimate.getOrElse(None).toString(),
+      "moments" -> moments.getOrElse(None).toString()
```
leahmcguire: getOrElse(None) is weird maybe moments.map(_.toString).getOrElse("")

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
TuanNguyen27: I got this error `diverging implicit expansion for type algebra.Monoid[com.salesforce.op.stages.impl.feature.TextStats]
starting with method fromAlgebraMonoid in trait FromAlgebraMonoidImplicit0
` if I simply do `moments + fd.moments` or `value + is not a member of Option[com.salesforce.op.stages.impl.feature.TextStats]` for`cardEstimate + fd.cardEstimate` (`testStatsSG.plus` only works on non-Option `TextStats` objects)

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
TuanNguyen27: correct!

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
leahmcguire: maybe try defining the implicit function like here: https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/filters/RawFeatureFilter.scala#L156

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -225,17 +246,53 @@ object FeatureDistribution {
...
+
     FeatureDistribution(
       name = name,
       key = key,
       count = 1L,
       nulls = nullCount,
       summaryInfo = summaryInfo,
       distribution = distribution,
+      moments = momentsAndCard._1,
+      cardEstimate = momentsAndCard._2,
```
leahmcguire: why put it in a tuple and then untuple?
maybe just 
`moments = m.map(momentsValues)`
`cardEstimate = m.map(cardinalityValues)`

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -152,6 +154,8 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
     lazy val workFlow = new OpWorkflow()
       .setResultFeatures(standardizedModel, unstandardizedModel).setInputDataset(DF)
     lazy val model = workFlow.train()
+    println(model.modelInsights(standardizedModel).features
+      .flatMap(_.distributions).size)
```
leahmcguire: no prints in test

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
TuanNguyen27:
```
implicit val sgTuple2Moments = new Tuple2Semigroup[Moments, Moments]()
implicit val sgTuple2TextStats = new Tuple2Semigroup[TextStats, TextStats]()
```
i tried doing this ^ and got a bunch of errors 

Error:(110, 38) diverging implicit expansion for type com.twitter.algebird.Semigroup[com.salesforce.op.stages.impl.feature.TextStats]
starting with method fromAlgebraMonoid in trait FromAlgebraMonoidImplicit0
    implicit val sgTuple2TextStats = new Tuple2Semigroup[TextStats, TextStats]()
Error:(110, 38) not enough arguments for constructor Tuple2Semigroup: (implicit asemigroup: com.twitter.algebird.Semigroup[com.salesforce.op.stages.impl.feature.TextStats], implicit bsemigroup: com.twitter.algebird.Semigroup[com.salesforce.op.stages.impl.feature.TextStats])com.twitter.algebird.Tuple2Semigroup[com.salesforce.op.stages.impl.feature.TextStats,com.salesforce.op.stages.impl.feature.TextStats].
Unspecified value parameters asemigroup, bsemigroup.
    implicit val sgTuple2TextStats = new Tuple2Semigroup[TextStats, TextStats]()
Error:(112, 50) diverging implicit expansion for type algebra.Monoid[com.salesforce.op.stages.impl.feature.TextStats]
starting with method fromAlgebraMonoid in trait FromAlgebraMonoidImplicit0
      combinedSummaryInfo, moments + fd.moments, cardEstimate + fd.cardEstimate, `type`)
Error:(112, 50) diverging implicit expansion for type algebra.ring.AdditiveMonoid[com.salesforce.op.stages.impl.feature.TextStats]
starting with method fromAlgebraMonoid in trait FromAlgebraMonoidImplicit0
      combinedSummaryInfo, moments + fd.moments, cardEstimate + fd.cardEstimate, `type`)
Error:(112, 50) diverging implicit expansion for type algebra.Monoid[Option[com.salesforce.op.stages.impl.feature.TextStats]]
starting with method fromAlgebraMonoid in trait FromAlgebraMonoidImplicit0
      combinedSummaryInfo, moments + fd.moments, cardEstimate + fd.cardEstimate, `type`)
Error:(112, 63) value + is not a member of Option[com.salesforce.op.stages.impl.feature.TextStats]
      combinedSummaryInfo, moments + fd.moments, cardEstimate + fd.cardEstimate, `type`)

leahmcguire: yes this is not a tuple - it is an option

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
leahmcguire: moments = value.map(momentsValues),
      cardEstimate = value.map(cardinalityValues),

leahmcguire: delete momentsAndCard

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -163,6 +164,18 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
     return Array(descaledsmallCoeff, originalsmallCoeff, descaledbigCoeff, orginalbigCoeff)
   }
 
+  def getFeatureMomentsAndCard(inputModel: FeatureLike[Prediction],
```
Jauntbox: I think it might be easier to read in the tests below if you make this return a `Map[String, FeatureDistribution]` where the key is the feature name - that way we know which distribution goes with which input feature.

You could make it like:
```def getFeatureMomentMap(inputModel: FeatureLike[Prediction], df: DataFrame): Map[String, FeatureDistribution] = {
    val workFlow = new OpWorkflow().setResultFeatures(inputModel).setInputDataset(df)
    val worflowWithRFF = workFlow.withRawFeatureFilter(
      trainingReader = Some(workFlow.getReader()),
      scoringReader = None
    )
    val model = worflowWithRFF.train()

    val insights = model.modelInsights(inputModel)
    insights.features.map(f => f.featureName -> f.distributions.head).toMap
  }```

File: core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala [hunk 18]
```diff
@@ -761,4 +774,29 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
...
   }
+
+  val tol2 = 0.1
+  val MomentsAndCard = getFeatureMomentsAndCard(standardizedLinpred, linRegDF._3)
+  val moments = MomentsAndCard._1
+  val cardinality = MomentsAndCard._2
+  val uniqueValsmallFeature = smallNorm.map(_.toDouble.get).toSet
+  val uniqueValbigFeature = bigNorm.map(_.toDouble.get).toSet
+
+  it should "correctly return moments calculation for numeric features" in {
```
Jauntbox: If you do the above suggestion, then you can also combine these two tests and make it so you're not hardcoding indices in your comparisons (downside is that since you have the feature name, you'd have to get the expected values from the dataframe itself):

```it should "correctly return moments calculation for numeric features" in {
    import spark.implicits._
    val df = linRegDF._3
    val featMap = getFeatureMomentMap(standardizedLinpred, linRegDF._3)

    // Go through each feature and check that the mean, variance, and unique counts match the data
    featMap.foreach(f => {
      f._2.moments.get.count shouldBe 1000

      val (expectedMean, expectedVariance) = df.select(avg(f._1), variance(f._1)).as[(Double, Double)]
        .collect().head
      math.abs((f._2.moments.get.mean - expectedMean) / expectedMean) < meanTol shouldBe true
      math.abs((f._2.moments.get.variance - expectedVariance) / expectedVariance) < varTol shouldBe true

      val actualUniques = df.select(f._1).as[Double].collect().toSet
      f._2.cardEstimate.get.valueCounts.keySet.map(_.toDouble).subsetOf(actualUniques) shouldBe true
    })
  }

File: core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala [hunk 19]
```diff
@@ -761,4 +774,29 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
...
+    moments.size shouldBe 2
+    moments(0).count shouldBe 1000
+    math.abs(moments(1).mean) < tol2 shouldBe true
+    math.abs(moments(1).variance - 10.0) / (moments(1).variance + 10.0)  < tol2 shouldBe true
```
Jauntbox: If you have an analytic expected answer you can just compare to that instead of an avg. of two quantities you expect to be the same (you're missing the factor of 2 for your averages too).

eg. `math.abs((moments(1).variance - expected) / expected) < 1e-12`

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -761,4 +774,29 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
     absError / bigCoeffSum < tol shouldBe true
     absError2 / smallCoeffSum < tol shouldBe true
   }
+
+  val tol2 = 0.1
```
Jauntbox: Can we put all of these inside the test(s) that use them or up with the other fields?

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -761,4 +774,29 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
...
+  val uniqueValsmallFeature = smallNorm.map(_.toDouble.get).toSet
+  val uniqueValbigFeature = bigNorm.map(_.toDouble.get).toSet
+
+  it should "correctly return moments calculation for numeric features" in {
+    moments.size shouldBe 2
+    moments(0).count shouldBe 1000
+    math.abs(moments(1).mean) < tol2 shouldBe true
+    math.abs(moments(1).variance - 10.0) / (moments(1).variance + 10.0)  < tol2 shouldBe true
+    moments(1).count shouldBe 1000
+    math.abs(moments(0).mean - 10000) / (moments(0).mean + 10000) < tol2 shouldBe true
```
Jauntbox: Also, you can compare to the actual mean of the sample rather than the mean of the distribution you're pulling from. That would let you use tighter tolerances and would probably be less subject to flakyness for small samples.

TuanNguyen27: good point, it was pretty flaky and i had to pick a higher threshold...

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
leahmcguire: @tovbinm any idea how to get the implicits to work so this doesnt have to be written out?

Jauntbox: @leahmcguire adding `implicit val opMonoid = optionMonoid[TextStats]` let me do `val combinedMoments = moments + fd.moments` and the tests passed

Jauntbox: Ooo, I think I fixed it!

I was digging into Algebirds monoid construction code and it looked like it was expecting an implicit `Monoid[TextStats]`, even though the helper function only asked for a SemiGroup.

I added a monoid to out `TextStats` and then it worked fine with
```
implicit val testStatsMonoid: Monoid[TextStats] = TextStats.monoid(RawFeatureFilter.MaxCardinality)
implicit val opMonoid = optionMonoid[TextStats]
```

I don't actually know why we made `TextStats` a SemiGroup instead of a monoid in the first place, the only difference between the two is an identity element and we already explicitly defined a zero element for `TextStats`.

tovbinm:
@TuanNguyen27 please remember to clean the commit message prior to merging next time.

koertkuipers:
FYI we are running into some issues with this in spark 3 which has json4s 3.6.6 instead of 3.5.3
i first thought it was maybe a scala 2.12 issue but our scala 2.12 branch (not spark3) works fine.

seems it dislikes the companion object for Moments with its own apply methods which have context bounds.

```com.salesforce.op.filters.RawFeatureFilterTest > RawFeatureFilter should correctly serialize and deserialize raw feature filter results with fixed data FAILED
    org.json4s.package$MappingException: Can't find constructor for Moments
        at org.json4s.reflect.package$.fail(package.scala:95)
        at org.json4s.reflect.ScalaSigReader$.$anonfun$readConstructor$3(ScalaSigReader.scala:26)
        at scala.Option.getOrElse(Option.scala:138)
        at org.json4s.reflect.ScalaSigReader$.readConstructor(ScalaSigReader.scala:26)
        at org.json4s.reflect.Reflector$ClassDescriptorBuilder.ctorParamType(Reflector.scala:93)
        at org.json4s.reflect.Reflector$ClassDescriptorBuilder.$anonfun$createConstructorDescriptors$7(Reflector.scala:177)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:237)
        at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
        at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
        at scala.collection.TraversableLike.map(TraversableLike.scala:237)
        at scala.collection.TraversableLike.map$(TraversableLike.scala:230)
        at scala.collection.AbstractTraversable.map(Traversable.scala:108)
        at org.json4s.reflect.Reflector$ClassDescriptorBuilder.$anonfun$createConstructorDescriptors$3(Reflector.scala:159)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:237)
        at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
        at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
        at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:39)
        at scala.collection.TraversableLike.map(TraversableLike.scala:237)
        at scala.collection.TraversableLike.map$(TraversableLike.scala:230)
        at scala.collection.AbstractTraversable.map(Traversable.scala:108)
        at org.json4s.reflect.Reflector$ClassDescriptorBuilder.createConstructorDescriptors(Reflector.scala:139)
        at org.json4s.reflect.Reflector$ClassDescriptorBuilder.constructorsAndCompanion(Reflector.scala:135)
        at org.json4s.reflect.Reflector$ClassDescriptorBuilder.result(Reflector.scala:204)
        at org.json4s.reflect.Reflector$.createDescriptor(Reflector.scala:53)
        at org.json4s.reflect.Reflector$.$anonfun$describe$1(Reflector.scala:48)
        at org.json4s.reflect.package$Memo.apply(package.scala:36)
        at org.json4s.reflect.Reflector$.describe(Reflector.scala:48)
        at org.json4s.Extraction$.decomposeObject$1(Extraction.scala:119)
        at org.json4s.Extraction$.internalDecomposeWithBuilder(Extraction.scala:231)
        at org.json4s.Extraction$.addField$1(Extraction.scala:111)
        at org.json4s.Extraction$.decomposeObject$1(Extraction.scala:142)
        at org.json4s.Extraction$.internalDecomposeWithBuilder(Extraction.scala:231)
```

leahmcguire:
thanks for the heads up!

@TuanNguyen27 can you please add the FeatureDistribution class and subclasses to the json serialization formats for record insights https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/ModelInsights.scala#L394

Issue #248, opened 2019-03-19
Deep Learning in TransmogrifAI


LakshKD:
**Problem**
I am thinking that can we start incorporating some deep learning models as well in TransmogrifAI. 

**Solution**
I am happy to collaborate, if there is any plan or I can discuss also in order to start some deep learning stuff in TransmogrifAI.

**Additional context**
We can start with some basic models like Multi-Layer Perceptron etc.

LakshKD:
@tovbinm  Any ideas regarding this :-) ?

tovbinm:
@LakshKD yes, we are planning to add some DL support for TransmogrifAI, in particular integration with TensorFlow runtime. Do you have a particular use case you want to solve with TransmogrifAI?

LakshKD:
> @LakshKD yes, we are planning to add some DL support for TransmogrifAI, in particular integration with TensorFlow runtime. Do you have a particular use case you want to solve with TransmogrifAI?

I don't have a use case right now but I want to contribute to TransmogrifAI by starting some Deep Learning stuff. I would love to discuss this tensorflow integration and will be very happy to contribute as well.

tovbinm:
In order to add TensorFlow support we would need roughly the following:
1. research and evaluate which libraries can be used to provide TensorFlow model training on Spark (if not training then at least loading pre-trained models & scoring)
2. investigate how to integrate it with TransmogrifAI's estimator or transformer
3. actually implement it ;)

a1pha:
Hello everyone, 
I was interested in working on this issue/feature request, as a first issue with TransmogrifAI. I have some familiarity with working with Tensorflow and deep learning models. I was wondering if we should start with common models such as MobleNetV2, VGG, ResNet, etc. To allow users to load these models from checkpoints/pre-trained weights for transfer learning. Please let me know what are the next steps I can take to help out, if possible.
Cheers,
Abhi

py-ranoid:
Hi @a1pha 
The models you've suggested are for object recognition (from images) and can't be applied to the current type hierarchy.
![](https://cdn-images-1.medium.com/max/1600/1*vkrbLrOyIinhonrPCGHivA.png)

tovbinm:
For binary data we have `Base64` type.

LakshKD:
> In order to add TensorFlow support we would need roughly the following:
> 
>     1. research and evaluate which libraries can be used to provide TensorFlow model training on Spark (if not training then at least loading pre-trained models & scoring)
[...]

PR #355, opened 2019-07-05
Added TensorFlow module + BERT model

tovbinm:
**Related issues**
TransmogrifAI supports neither training nor inferencing deep learning models #248

- Add additional TensorFlow module
- Add ability to create, load and run TensorFlow models
- Add BERT model
- Compare with JohnSnow performance - https://github.com/JohnSnowLabs/spark-nlp/blob/master/src/main/scala/com/johnsnowlabs/nlp/embeddings/BertEmbeddings.scala. Update: JohnSnow dev team investigation is here - https://github.com/JohnSnowLabs/spark-nlp/issues/570

- *TBD* BERT model inference is **very** slow. Additional investigation is required. And since JohnSnow library works better, perhaps we should wrap around it?

NA


tovbinm:
I think it makes sense to replace BERT with ALBERT - https://ai.googleblog.com/2019/12/albert-lite-bert-for-self-supervised.html 

@kinfaikan ?

tovbinm:
This isn't going anywhere. Closing for now.

Issue #332, opened 2019-06-14
Support Scala 2.12


koertkuipers:
Spark 2.4 has Scala 2.11 and 2.12 builds. Starting with Spark 3.0 Scala 2.12 will become default and Scala 2.11 will be dropped. 

This depends on https://github.com/salesforce/TransmogrifAI/issues/184

This will require:
* Changes to the gradle build support scala 2.11 and 2.12. See kafka for an example how to do this.
* Some hopefully minor code changes for compilation in Scala 2.12
* Updates/upgrades to dependencies. Some are not build for Scala 2.12 at all, other only have newer versions available for Scala 2.12

koertkuipers:
Starting from #184 first issue i ran into is gradle-scalastyle-plugin not found for scala 2.12. resolved by adding this maven repo:
```
        maven {
            name = "ngbinh"
            url = "https://dl.bintray.com/ngbinh/maven"
        }
```

koertkuipers:
To stop scala compiler from exiting with strange messages like `assertion failed: ClassBType.info not yet assigned: Lorg/apache/spark/rdd/RDD;` i removed `-optimize` flag

tovbinm:
Just a heads up - with merged #274 and with Scala 2.12 compile the majority of the tests will fail, e.g. `UnaryLamdaTransformerTest`, due to stage serialization issues.  We need to update all lambda functions with concrete classes as explained in #274

tovbinm:
Thanks for making the effort to upgrade @koertkuipers !!

koertkuipers:
after publishing my own in-house versions of xgboost and mleap it seems everything compiles and tests fine using scala 2.12. seemed a little too easy... wonder if i am not running tests correctly.

this is before https://github.com/salesforce/TransmogrifAI/pull/274
i get a bunch of merge conflicts with that one that i need to take a look at. since i didnt make any significant changes the merge conflicts are likely with spark-2.4 branch.

koertkuipers:
xgboost not having scala version in package name doesnt help
e.g. `xgboost4j-spark-0.90.jar` vs `xgboost4j-spark_2.11-0.90.jar `

koertkuipers:
see also https://github.com/dmlc/xgboost/pull/4574

koertkuipers:
[...]

PR #357, opened 2019-07-10
Convert lambda functions into concrete classes to allow compatibility with Scala 2.11/2.12

tovbinm:
**Related issues**
Follows up on PR #274 and lands the foundation to support Scala 2.12 #332 

Scala 2.12 generates different class names for lambdas than Scala 2.1. Example:
```scala
object A {
  class Foo extends Function1[Int, Int] { def apply(v: Int): Int = v + 1 }
  def foo: Int => Int = x => x + 1
}
println(A.f.getClass.getName)
println(A.foo.getClass.getName)

Scala 2.11:
    A$Foo
    A$$anonfun$foo$1

Scala 2.12:
    A$Foo
    A$$$Lambda$13458/243505737
```
To maintain forward compatibility of models (so one won’t need to retrain models and would be able to load them in Scala 2.11/2.12) we would like to convert all the lambdas into concrete classes.

Changing all the lambdas with concrete classes.

NA

### DIFF
--- a/core/src/main/scala/com/salesforce/op/dsl/RichDateFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichDateFeature.scala
@@ -56,7 +56,7 @@ trait RichDateFeature {
       f.transformWith(
         new UnaryLambdaTransformer[Date, DateList](
           operationName = "dateToList",
-          RichDateFeatureLambdas.toDateList
+          new RichDateFeatureLambdas.ToDateList
         )
       )
     }
@@ -137,7 +137,7 @@ trait RichDateFeature {
       f.transformWith(
         new UnaryLambdaTransformer[DateTime, DateTimeList](
           operationName = "dateTimeToList",
-          RichDateFeatureLambdas.toDateTimeList
+          new RichDateFeatureLambdas.ToDateTimeList
         )
       )
     }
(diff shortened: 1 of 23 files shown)


wsuchy:
Imagine we move all lambdas to a companion object like this:
```class RichMapFeature {
....
(f.map[RealNN](RichMapFeatureLambdas.predictionToReal),
....
}

object RichMapFeatureLambdas{
val predictionToReal = (p: Prediction): RealNN => p.prediction.toRealNN
}
```
Still whenever try to inspect `RichMapFeatureLambdas.predictionToReal`  we get that random name under scala 2.12 and more deterministic value under scala 2.11, still not good.

But, given that we a priori know where are stored all lambdas for that class we could easily inspect our `RichMapFeatureLambdas` object and create a reverse lookup mapping all generated numbers to the property holding the actual function, e.g ` Lambda1$$$Lambda$4624/831965909@6259b7e5 -> RichMapFeatureLambdas.predictionToReal`

Then during the serialization we would simply use our map to replace compiler generated names  with class.property references. Because it would use run-time reflections the mapping would be always correct.

Finally the solution would work for both scala 2.11/2.12 and would be future proof with our macro.
@tovbinm

tovbinm:
@wsuchy Sorry, I dont understand what exactly are you suggesting? what's wrong with the solution I propose in this PR?

wsuchy:
@tovbinm Eventually we don't want our users to have to create manually their lambda functions as concrete classes and we know there is a solution (like the one I described). Although your changes apply only to the internal code  we will end up having two ways of dealing with lambdas. Also, modifying the code where you have to write functions manually isn't the greatest experience. This is the reason I proposed to use macro for generating these instead.

tovbinm:
I think our macros would offer exactly the generation of concrete classes as proposed in this PR (since concrete classes are the most straightforward way to handle lambda functions in all scenarios possible, unless Scala decides to change their FunctionX interface of course, which is unlikely).

wsuchy:
Unfortunately you don't have types present at macro evaluation time so you can't populate type params here: `Function1[???,???]`. Unless we don't want to write a compiler plugin of course.

tovbinm:
Anonymous lambdas are much more difficult to debug and comprehend that simple classes, therefore I tend to assume that if simple macros spent work then we will have to write a compiler plugin if we decide to do so.

tovbinm:
1. I followed all the changes that were done in #274 and changed chose to concrete classes.
2. No, these can be standalone classes that live outside objects.

PR #358, opened 2019-07-10
Aggregate spark metrics during run time instead of post processing by default

gerashegalov:
Replaces the previous PR #341


tovbinm:
```
> Task :utils:scalaStyle
warning file=/home/travis/build/salesforce/TransmogrifAI/utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala message=Use correct indentation line=323
error file=/home/travis/build/salesforce/TransmogrifAI/utils/src/test/scala/com/salesforce/op/utils/spark/OpSparkListenerTest.scala message=Expected token SEMI but got Token(INTERPOLATION_ID,s,4046,s)
```

PR #363, opened 2019-07-11
Replace mapValues with immutable Map where applicable

### LINKED ISSUE #362: Not serializable exception during `computeFeatureStats`
**Description**
Serializability issue during one of the stages of model training:
```
... 
(TID 3067704) had a not serializable result: scala.collection.immutable.MapLike$$anon$2
Serialization stack:
    - object not serializable (class: scala.collection.immutable.MapLike$$anon$2)
...
```

Stacktrace leads to `com.salesforce.op.filters.PreparedFeatures#summaries` method, which is called from `com.salesforce.op.filters.RawFeatureFilter#computeFeatureStats`. The usage 
 of `mapValues` leads to a non serializable Map, which is a long standing [scala bug](https://github.com/scala/bug/issues/7005).

I've attempted to fix that just by adding a `.map(identity)`, which resolved the issue, i.e.:
```
predictors.mapValues(Summary(_)).map(identity)
```
 Not sure about "correct" way of dealing with that

**Additional context**
Transmogrify version: `0.5.1` & `0.5.2`. Though it seems to be applicable to current `master`, as the related code didn't change since then.

tovbinm:
**Related issues**
Fixes #362

PreparedFeatures.summary should produce immutable map to allow their proper serialization with Spark.

NA

### DIFF
--- a/core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
+++ b/core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
@@ -232,10 +232,10 @@ private[op] class OpMultiClassificationEvaluator
     ThresholdMetrics(
       topNs = topNs,
       thresholds = thresholds,
-      correctCounts = agg.mapValues { case (cor, _) => cor.toSeq },
-      incorrectCounts = agg.mapValues { case (_, incor) => incor.toSeq },
-      noPredictionCounts = agg.mapValues { case (cor, incor) =>
-        (Array.fill(nThresholds)(nRows) + cor.map(-_) + incor.map(-_)).toSeq
+      correctCounts = agg.map { case (k, (cor, _)) => k -> cor.toSeq },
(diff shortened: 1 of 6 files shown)


core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -232,10 +232,10 @@ private[op] class OpMultiClassificationEvaluator
     ThresholdMetrics(
       topNs = topNs,
       thresholds = thresholds,
-      correctCounts = agg.mapValues { case (cor, _) => cor.toSeq },
```
gerashegalov: nit: just for completeness
one can materialize mapValues view by 
```scala
agg.mapValues { case (cor, _) => cor.toSeq }.view.toMap
```
or 
```scala
agg.transform { case (_, (cor, _)) => cor.toSeq },
```

tovbinm: Good point. Thanks.

PR #372, opened 2019-07-26
Making model selectors robust to failing models

leahmcguire:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/370

Make model selector no fail when a portion of attempted models don't work. Also exposed parameter to limit max time will wait for modeling to finish in model selector

**Additional context**


core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
```diff
@@ -287,25 +321,41 @@ private[op] trait OpValidator[M <: Model[_], E <: Estimator[_]] extends Serializ
...
-        val numModels = params.length
-        val metrics = new Array[Double](params.length)
-        log.info(s"Train split with multiple sets of parameters.")
-        val models = estimator.fit(train, params).asInstanceOf[Seq[M]]
-        for {i <- 0 until numModels} {
-          val metric = evaluator.evaluate(models(i).transform(test, params(i)))
-          log.info(s"Got metric $metric for model $name trained with ${params(i)}.")
-          metrics(i) = metric
+        val paramsMetrics = params.map { p =>
+          Try {
```
tovbinm: You might wanna consider to parallelize by splits AND models, I.e just use futures instead of tries nested inside futures.

core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala
```diff
@@ -59,11 +61,13 @@ class RegressionModelSelectorTest extends FlatSpec with TestSparkContext
...
   import spark.implicits._
+  val rand = new Random(seed)
 
-  val rawData: Seq[(Double, Vector)] = List.range(0, 100, 1).map(i =>
-    (i.toDouble, Vectors.dense(2 * i, 4 * i)))
+  val rawData: Seq[(Double, Vector)] = List.range(-100, 100, 1).map(i =>
+    (i.toDouble, Vectors.dense(2 * i, 4 * i + 5, rand.nextFloat())))
 
   val data = sc.parallelize(rawData).toDF("label", "features")
+  data.show()
```
tovbinm: No show() in tests.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
```diff
@@ -287,25 +321,41 @@ private[op] trait OpValidator[M <: Model[_], E <: Estimator[_]] extends Serializ
...
       }
     }
-    val summary = SparkThreadUtils.utils.awaitResult(Future.sequence(summaryFuts), Duration.Inf).toArray
+
+    val summaryOfAttempts = summaryFuts.map { f => f.map(Option(_)).recover {
+      case e: Throwable =>
+        log.warn(s"Model attempted in model selector failed with following issue: \n${e.getMessage}")
+        None
+    }}
+    val summary = SparkThreadUtils.utils.awaitResult(Future.sequence(summaryOfAttempts), maxWait).flatten.toArray
```
tovbinm: Perhaps it’s worth to log the maxWait value

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -423,7 +423,7 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
           log.info("Estimate best Model with CV/TS. Stages included in CV are: {}, {}",
             during.flatMap(_.map(_._1.stageName)).mkString(", "), modelSelector.uid: Any
           )
-          modelSelector.findBestEstimator(trainFixed, during, persistEveryKStages)
+          modelSelector.findBestEstimator(trainFixed, Option(during))
```
tovbinm: Why not pass persistEveryK param anymore?

leahmcguire: it wasnt being used

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
leahmcguire: I tried that but I think the copy of the model causes some kind of lock - when I had them both in futures the first test where the model selectors are ran for 30 minutes without ever finishing....

leahmcguire:
If a model fails it doesnt appear in model insights - if you wanted to know which models you tried and failed you would need to compare the modelselector params to the sequence of evaluated models

Jauntbox:
> If a model fails it doesnt appear in model insights - if you wanted to know which models you tried and failed you would need to compare the modelselector params to the sequence of evaluated models

Ok, so for our use case we could still get a list of failed models with that comparison.

What about more complicated cases where we have a random search or a data-dependent hyperparameter search? It looks like for random search we'd still have the info stored in the ParamMap in the models argument to ModelSelector so we'd probably do the same thing for a data-dependent hyperparam search too?

leahmcguire:
Maybe - I still need to think about how we would implement data dependent hyperparmeter search. I can make it a requirement that the models tried exist somewhere...

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
tovbinm: Let’s put a todo there or perhaps Chris S. Can help?

core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala
```diff
@@ -213,6 +216,53 @@ class RegressionModelSelectorTest extends FlatSpec with TestSparkContext
...
+    RegressionEvalMetrics.values.foreach(metric =>
+      assert(metaData.trainEvaluation.toJson(false).contains(s"${metric.entryName}"),
+        s"Metric ${metric.entryName} is not present in metadata: " + metaData)
+    )
+    metaData.validationResults.foreach(println(_))
+    metaData.validationResults.size shouldBe 42
+  }
+
+
+  it should "fail when all models fail" in {
```
gerashegalov: can you explain in comments or assertions which model fails and why in each case?

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
```diff
@@ -93,37 +90,48 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
...
-      label = in1.name, features = in2.name, dag = Option(dag), splitter = splitter,
-      stratifyCondition = validator.isClassification
+  protected[op] def findBestEstimator(data: DataFrame, dag: Option[StagesDAG] = None)
+    (implicit spark: SparkSession): (BestEstimator[E], Option[SplitterSummary], DataFrame) = {
+
+    val PrevalidationVal(splitterSummary, dataOpt) = prepareForValidation(data, in1.name)
+    val dataUse = dataOpt.getOrElse(data)
+
+    val theBestEstimator = validator.validate(modelInfo = modelsUse, dataset = dataUse,
+      label = in1.name, features = in2.name, dag = dag, splitter = splitter
```
gerashegalov: should we use `labelColName` everywhere for consistency?

core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala
gerashegalov: maxWait could be tested explicitly with some short duration.

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpValidator.scala
wsuchy: @tovbinm @leahmcguire I was able to replace `params` with `params.par` and although it worked it was much slower than one would expect. First, does it make sense to fit and eval all models in parallel when there is only one spark context? 
BTW do you have `jvisualvm` avail in Zulu?

gerashegalov: You can always get visualvm standalone https://visualvm.github.io/download.html

wsuchy: https://github.com/salesforce/TransmogrifAI/pull/373 - work fine for me. Though I am not sure about running anything in parallel with spark context as dependency.

tovbinm: Please refer to the implementation of cross validation in Spark (it uses Futures and there is a good reason why).

tovbinm: And that’s why I propose to stay with Futures, but just use them correctly.

wsuchy: @tovbinm If you are referring to https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/tuning/CrossValidator.scala then the only difference with my PR is that they are using `foldMetricFutures.map(ThreadUtils.awaitResult(_, Duration.Inf))` instead of `Future.sequence` therefore can you elaborate more?

leahmcguire:
addressed all the comments - can someone approve?

core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala
```diff
@@ -213,6 +217,69 @@ class RegressionModelSelectorTest extends FlatSpec with TestSparkContext
...
+        numFolds = 4,
+        validationMetric = Evaluators.Regression.mse(),
+        seed = 10L,
+        modelTypesToUse = Seq(RegressionModelsToTry.OpLinearRegression),
+        maxWait = Duration(1L, "microsecond")
+      )
+      .setInput(label, features)
+
+
+    intercept[Exception](testEstimator.fit(data))
```
gerashegalov: let us use more specific exceptions in intercepts. e.g. TimeoutException here

PR #373, opened 2019-07-29
used list of futures

wsuchy:
Alternatively you could replace `params` with `params.par` instead.


tovbinm:
Let’s not use .par please but stay with Futures.

PR #375, opened 2019-07-29
Use compact and compressed model json by default

### LINKED ISSUE #374: compact and compressed json serialization for models
**Problem**
OpWorkflowModelWriter outputs pretty printed large jsons. Pretty printing for disk is an unnecessary feature because there are a lot of tools that can pretty-print on-demand when a human user inspects the file. 

illustration 

| format                | size in bytes |
|-----------------|-------------|
| pretty (original)  | 1698606     |
| compact json     | 963400       |
| gzip pretty          | 203380       |
| gzip compact     | 185558        |    

**Solution**
Output compact json. (Longer term allow for efficient binary formats). Gzip-compress on top of this.

**Alternatives**
making it configurable

gerashegalov:
**Related issues**
Fixes #374 

use compact serialization, apply gzip to it.

make this behavior configurable

**Additional context**
In a test scenario output is reduced from 1.6M to 188K

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
@@ -35,6 +35,7 @@ import com.salesforce.op.filters.RawFeatureFilterResults
 import com.salesforce.op.stages.{OPStage, OpPipelineStageWriter}
 import enumeratum._
 import org.apache.hadoop.fs.Path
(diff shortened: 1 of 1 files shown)


gerashegalov:
> Don’t you need to mention the gzip codec when reading the model?

It's handled by Hadoop TextInputFormat based on the filename extension transparently. On the write path the filename extension is appended based on compression codec. On the read path, you can even mix files with different compression codecs/extensions and uncompressed files in the same dir.

Issue #376, opened 2019-07-30
RecordInsightsLOCO#returnTopPosNeg complexity is =0(n^2 * log n) per record 


gerashegalov:
**Describe the bug**
returnTopPosNeg creates sparse vectors in a loop over bits. `Vectors.sparse` sorts the indices array 

**To Reproduce**
Enable LOCO

**Expected behavior**
Using DenseVector for 0(1) generating of LOCO vectors

**Logs or screenshots**
<img width="1143" alt="Screen Shot 2019-07-30 at 10 29 23 AM" src="https://user-images.githubusercontent.com/3187938/62151326-ee193e80-b2b4-11e9-9507-8075bb0c6e77.png">

PR #377, opened 2019-07-30
Avoid creating SparseVectors for LOCO

gerashegalov:
**Related issues**
#376 

reuse the original SparseVector as a mutable template

**Additional context**
In a scoring job:

|before LOCO| with record insights | with record insights after PR |
|-------------|--------------------|----------|
|  < 10s               | 168 sec                   |   11sec    |


leahmcguire:
Wow!

michaelweilsalesforce:
Wait, if we apply `foreachActive` on a dense vector, wouldn't this look at ALL the element of the vectors?

gerashegalov:
it's still WIP but I have this guard `if oldVal != 0.0` in the pattern match @michaelweilsalesforce

tovbinm:
Neat. What about memory complexity?

gerashegalov:
reworked the solution to avoid the memory overhead of the dense vector.

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -165,20 +156,42 @@ class RecordInsightsLOCO[T <: Model[T]]
...
-
     val minMaxHeap = new MinMaxHeap(k)
     val aggregationMap = mutable.Map.empty[String, (Array[Int], Array[Double])]
-    for {i <- featureArray.indices} {
-      val (oldInd, oldVal) = featureArray(i)
-      val diffToExamine = computeDiffs(i, oldInd, oldVal, featureArray, featureSize, baseScore)
-      val history = histories(oldInd)
 
+    agggregateDiffs(0, Left(featureSparse), indexToExamine, minMaxHeap, aggregationMap,
+      baseScore)
```
leahmcguire: So the sparse features you just put in a value of 0? Cant we just skip adding them to the heap?

gerashegalov: I had the same idea but in one of the iteration I ran into test failures and deferred it to later. I'll recheck now that I have everything green

michaelweilsalesforce: What kind of failures have you encountered?

leahmcguire: it may be that we were doing an unnecessary calculation and that just happened to be captured in the test...

gerashegalov: @michaelweilsalesforce you can reproduce it by commenting out the line 171-172.
```
Aggregate all the derived hashing tf features of rawFeature - text. 0.08025355373244505 was not less than 1.0E-10 expected aggregated LOCO value (0.006978569889777832) should be the same as actual (0.08723212362222289)

Aggregate x_HourOfDay and y_HourOfDay of rawFeature - dateFeature. 0.016493734169231777 was not less than 1.0E-10 expected aggregated LOCO value (0.016493734169231777) should be the same as actual (0.032987468338463555)
```

michaelweilsalesforce: @sanmitra

michaelweilsalesforce: @leahmcguire @gerashegalov The reason for tracking zero values is whenever we want to average LOCOs of a same raw text feature we are also including the zero values. 
E.g if text feature `TextA` has on a row 6 non zero values `loco1`, ..., `loco6` and 4 0s, we are dividing by 10 : 
(`loco1` + `loco2` + ... + `loco6`)/10

michaelweilsalesforce: LEt me write a fix that will not go over the zeros

gerashegalov:
Thanks Michael!

michaelweilsalesforce:
@gerashegalov Here is a proposal that skips the diffs for zero values. Code can be nicer though

gerashegalov:
Thank you, looks good,  just a few polishes

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -116,34 +114,28 @@ class RecordInsightsLOCO[T <: Model[T]]
     Set(FeatureType.typeName[DateMap], FeatureType.typeName[DateTimeMap])
 
   // Indices of features derived from Text(Map)Vectorizer
-  private lazy val textFeatureIndices = getIndicesOfFeatureType(textTypes ++ textMapTypes)
+  private lazy val textFeatureIndices: Seq[Int] = getIndicesOfFeatureType(textTypes ++ textMapTypes,
+    h => h.indicatorValue.isEmpty && h.descriptorValue.isEmpty)
```
leahmcguire: maybe update comment to indicate only getting hashed text values

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -158,70 +150,93 @@ class RecordInsightsLOCO[T <: Model[T]]
...
 
-  private def getRawFeatureName(history: OpVectorColumnHistory): Option[String] = history.grouping match {
-    case Some(grouping) => history.parentFeatureOrigins.headOption.map(_ + "_" + grouping)
-    case None => history.parentFeatureOrigins.headOption
+  private def getRawFeatureName(history: OpVectorColumnHistory): Option[String] = {
+    val groupSuffix = history.grouping.map("_" + _).getOrElse("")
+    val name = history.parentFeatureOrigins.headOption.map(_ + groupSuffix)
+
+    // If the descriptor value of a derived date feature exists, then it is likely to be
+    // from unit circle transformer. We aggregate such features for each (rawFeatureName, timePeriod).
```
leahmcguire: this is true now - but may not always be true. If you want this to apply only for date unit circles should also check that one of the `parentFeatureStages` is a `DateToUnitCircleTransformer` or `DateToUnitCircleVectorizer`

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -158,70 +150,93 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+    agggregateDiffs(featureSparse, indexToExamine, minMaxHeap, aggregationMap,
+      baseScore)
+
+    // Adding LOCO results from aggregation map into heaps
+    for {(name, (indices, ar)) <- aggregationMap} {
+      // The index here is arbitrary
+      val (i, n) = (indices.head, indices.length)
+      val zeroCounts = zeroCountByFeature.get(name).getOrElse(0)
+      val diffToExamine = ar.map(_ / (n + zeroCounts))
+      minMaxHeap enqueue LOCOValue(i, diffToExamine(indexToExamine), diffToExamine)
```
leahmcguire: wait so we are aggregating everything into a map and then putting it into a heap and then just taking it out of the heap? doesn't that defeat the whole purpose of the heap? Shouldn't we be putting each value into the heap as we calculating it rather than aggregating the whole thing?

michaelweilsalesforce: We are only aggregating TF and Date features

leahmcguire: ah ok - can you add a comment to that effect

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
michaelweilsalesforce: This check is not consistent : Unit Circle Transformation in `DateMapVectorizer` is not reflected in the parentStages (`Seq[DateMapVectorizer]` instead).
I think the check on descriptor value is coherent.

michaelweilsalesforce: Or I can check the parentType instead

leahmcguire: If this change is explicitly to deal with date features that are transformed to unit circle then the check needs to be explicitly for that. Otherwise this is also applied to lat lon values (and anything else that we add later) and if we just check the type of the parent it assumes that we will always have unit circle transformation of dates - which could change at some point...

michaelweilsalesforce: I agree, but as I said above checking the `parentFeatureStages` won't work : for instance `DateMapVectorizer` may apply Unit circle transformation

leahmcguire: DateMapVectorizer does days between reference date and the date. The only two that do unit vector are DateToUnitCircleTransformer and DateToUnitCircleVectorizer

michaelweilsalesforce: Then there must be a bug in the shortcut : when `println(s"name ${history.columnName} stage ${history.parentFeatureStages} descriptor value ${history.descriptorValue}")
` I get
```
name dateMapFeature_k0_y_DayOfYear_33 stage ArrayBuffer(vecDateMap_DateMapVectorizer_00000000004c) descriptor value Some(y_DayOfYear)
name dateMapFeature_k1_x_DayOfYear_34 stage ArrayBuffer(vecDateMap_DateMapVectorizer_00000000004c) descriptor value Some(x_DayOfYear)
name dateMapFeature_k1_y_DayOfYear_35 stage ArrayBuffer(vecDateMap_DateMapVectorizer_00000000004c) descriptor value Some(y_DayOfYear)
name dateFeature_x_HourOfDay_0 stage ArrayBuffer() descriptor value Some(x_HourOfDay)
name dateFeature_y_HourOfDay_1 stage ArrayBuffer() descriptor value Some(y_HourOfDay)
```
Those features both use the `.vetcorize` shortcut.

leahmcguire: blarg! you are right there is a bug in the feature history that means we loose info if the same feature undergoes multiple transformations :-( https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala#L53

Can you put a todo to update once the bug is fixed

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -158,70 +150,98 @@ class RecordInsightsLOCO[T <: Model[T]]
...
-    (textFeatureIndices ++ dateFeatureIndices).foreach(i => if (!featuresSparse.indices.contains(i)) res += i -> 0.0)
-    val featureArray = res.toArray
-    val featureSize = featuresSparse.size
+    val zeroValIndices = (textFeatureIndices ++ dateFeatureIndices)
+      .filterNot(featureIndexSet.contains)
+
+    // Count zeros by feature name
+    val zeroCountByFeature = zeroValIndices
+      .groupBy(i => getRawFeatureName(histories(i)).get)
+      .mapValues(_.length).view.toMap
```
tovbinm: What’s the point of .view here?

gerashegalov: to force map materialization

tovbinm:
Shall we merged this one?

PR #381, opened 2019-08-05
Add descaling ability to MinMaxNormEstimator and a custom estimator with enum

erica-chiu:
**Related issues**
N/A

Adding a custom estimator to allow standardization but keeping positive values by subtracting by min instead of mean
Adding an enum to allow for a choice among the current linear estimators for the label
Adding metadata to MinMaxNormEstimator to allow for descaling

The alternative is to create one large estimator that changes the linear function applied depending on the enum given as an input. Would reduce repetitive code, but may become clunky with too many different functions.

**Additional context**
These changes are meant to experiment with GLMs and normalizing the label for regression problems.


leahmcguire:
Is this really what we want? We would need to apply different normalization based on properties of the data - which we do not know beforehand. Rather than creating multiple estimators which do different things don't we want one estimator that examines the data and and does the correct thing for the data?

I would suggest one combined estimator with an enum of transformation types: "Norm, MinMax, StandardZeroMin, Auto"  where auto will select between all possible methods to give the best result for the data.

erica-chiu:
Is this auto transformation selecting the best result based on data distribution or on actual experimentation?

leahmcguire:
you would need to do actual experiments to determine what the best transformation for a given distribution is. My point is that we are trying to build AutoML so we need the transformation to be applied correctly within the stage not based on predefined knowledge about the data.

Jauntbox:
@leahmcguire That is what we'd want to work towards, but we don't know what "auto" should be yet. We're planning on testing these different rescalings in automl workflows to see what would work best before enabling it.

We could have an estimator that takes an enum that just has the three scalings we have for now, and then add auto once we do some experiments.

core/src/test/scala/com/salesforce/op/stages/impl/feature/StandardMinEstimatorTest.scala
```diff
@@ -0,0 +1,96 @@
...
+    val normedOutput = featureNormalizer.getOutput()
+    val metadata = featureNormalizer.fit(inputData).getMetadata()
+
+    val expectedMean = inputValues.sum / inputValues.length
+    val expectedStd = math.sqrt(inputValues.map(value => math.pow(expectedMean - value, 2)).sum
+      / (inputValues.length - 1))
+    val expectedMin = inputValues.min
+    val expectedSlope = 1 / expectedStd
+    val expectedIntercept = - expectedMin / expectedStd
+    ScalerMetadata(metadata) match {
```
gerashegalov: since we simply propagate `err`, maybe no need to do the pattern match here:
```
val meta = ScalerMetadata(metadata).get
meta shouldBe ...
...
```

PR #385, opened 2019-08-19
Model combiner

leahmcguire:
**Related issues**
Want to do different feature engineering for different model types or do model combination. Basically this allows two different feature engineering flows to be passed into different model selectors (eg you want to do one kind of feature eng for random forest and one for regression but want grid search for each). The accuracy of the models can be compared and then the predictions willbe combined by either taking the best model, equal combination, or weighted combination of the predictions.

Made a stage which can combine the results of two model selectors to produce a new score. The inputs are the two prediction outputs and the label. The output is a new prediction which is the combined result of the input predictions. The metadata is updated to reflect the combined prediction.


core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
```diff
@@ -179,6 +179,9 @@ case class SummaryStatistics
     meta.build()
   }
 
+  private[op] def +(sum: SummaryStatistics): SummaryStatistics = new SummaryStatistics(count, sampleFraction,
```
Jauntbox: This summation is mainly to allow cases where there are multiple feature vectors, where each one corresponds to its own sanity checker and model selector? In ModelInsights, it looked like it also allowed multiple sanity checkers leading to the same model. Is that possible?

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
```diff
@@ -313,10 +316,22 @@ case class Correlations
     corrMeta.putString(SanityCheckerNames.CorrelationType, corrType.sparkName)
     corrMeta.build()
   }
+
+  private[op] def +(corr: Correlations): Correlations =
+    new Correlations(featuresIn ++ corr.featuresIn, values ++ corr.values, nanCorrs ++ corr.nanCorrs, corrType)
```
Jauntbox: Would the Correlations have to have the same correlation type for this to make sense?

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
tovbinm: you should be able to use `com.twitter.algebird.macros.caseclass` instead. i.e.
```scala
implicit val summaryStatisticsSG = caseclass[SummaryStatistics]

val (s1: SummaryStatistics, s2: SummaryStatistics) = ???
s1 + s2
```
and same with `Correlations` type as well.

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
```diff
@@ -0,0 +1,246 @@
...
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case o: CombinationStrategy => this.entryName == o.entryName
+      case _ => false
+    }
+  }
+  override def hashCode(): Int = super.hashCode()
+}
+
+object CombinationStrategy extends Enum[CombinationStrategy] {
```
tovbinm: I think this enum should be defined in `feature` module and registered with the json in `OpPipelineStageReadWriteFormats`

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
```diff
@@ -0,0 +1,246 @@
...
+  def setEvaluators(ev: Seq[OpEvaluatorBase[_ <: EvaluationMetrics]]): this.type = {
+    evaluatorList = ev
+    this
+  }
+  override def evaluators: Seq[OpEvaluatorBase[_ <: EvaluationMetrics]] = evaluatorList
+}
+
+
+sealed abstract class CombinationStrategy extends EnumEntry with Serializable {
+  override def equals(obj: Any): Boolean = {
```
tovbinm: so which test require this equals override?

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
```diff
@@ -0,0 +1,245 @@
...
+    val (metricValueOpt1: Option[Double], metricValueOpt2: Option[Double], metricName: EvalMetric) =
+      if (eval1 == eval2) { // same decision metric in validation results
+        (getWinningModelMetric(summary1), getWinningModelMetric(summary2), eval1)
+      } else { // look for overlapping metrics in training results
+        val m2e1 = getMetricValue(summary2.trainEvaluation, eval1)
+        val m1e2 = getMetricValue(summary1.trainEvaluation, eval2)
+        if (m2e1.nonEmpty) {
+          (getMetricValue(summary1.trainEvaluation, eval1), m2e1, eval1)
+        } else if (m1e2.nonEmpty) {
+          (m1e2, getMetricValue(summary2.trainEvaluation, eval2))
```
Jauntbox: Don't you need an `eval2` at the end of the tuple here? How did the Scala compiler miss that actually??

Can you add some tests that cover some of these weird edge cases? I feel like there's a lot of new stuff being added.

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
```diff
@@ -0,0 +1,245 @@
...
+    val probability = p1.probability.zip(p2.probability).map{ case (v1, v2) => v1 * weight1 + v2 * weight2 }
+    val prediction =
+      if (probability.nonEmpty) probability.indexOf(probability.max).toDouble
+      else p1.prediction * weight1 + p2.prediction * weight2
+    Prediction(prediction = prediction, probability = probability, rawPrediction = rawPrediction)
+  }
+
+  lazy val labelColName: String = in1.name
+
+  @transient private var evaluatorList: Seq[OpEvaluatorBase[_ <: EvaluationMetrics]] = Seq.empty
```
Jauntbox: Why does this need to be transient?

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
```diff
@@ -0,0 +1,245 @@
...
+    doc = "Method used to combine predictions",
+    isValid = (in: String) => CombinationStrategy.values.map(_.entryName).contains(in)
+  )
+  def setCombinationStrategy(value: CombinationStrategy): this.type = set(combinationStrategy, value.entryName)
+  def getCombinationStrategy(): CombinationStrategy = CombinationStrategy.namesToValuesMap($(combinationStrategy))
+  setDefault(combinationStrategy, CombinationStrategy.Best.entryName)
+
+}
+
+class SelectedCombiner
```
Jauntbox: Docs please

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
```diff
@@ -0,0 +1,245 @@
...
+sealed abstract class CombinationStrategy extends EnumEntry with Serializable {
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case o: CombinationStrategy => this.entryName == o.entryName
+      case _ => false
+    }
+  }
+}
+
+object CombinationStrategy extends Enum[CombinationStrategy] {
```
Jauntbox: Docs here too

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
```diff
@@ -50,7 +50,7 @@ class OpVectorMetadata private
 (
   val name: String,
   col: Array[OpVectorColumnMetadata],
-  val history: Map[String, FeatureHistory]
+  val history: Map[String, FeatureHistory] // TODO fix map -> causes problems when multiple vectorizers used on feature
```
Jauntbox: I thought this was still an issue? No?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -443,43 +444,69 @@ case object ModelInsights {
...
+    } else None
+
+    log.info(
+      s"Found ${sanityCheckers.length} sanity checkers" +
+        s"${sanityChecker.map("will preferentially use results from checkers in model path:" + _.uid +
+          " to fill in model insights")}"
+    )
+
+
+    val label = model.map(_.getInputFeature[RealNN](0))
```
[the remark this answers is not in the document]
leahmcguire: in a map on a dataframe this is a map over an optional stage - it will be executed on the driver

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
leahmcguire: yes and yes

leahmcguire: but I dont want the count or sample fraction to be added...

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
leahmcguire: yes- but I am not sure it makes sense to error out on that... should I maybe log a warning?

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
leahmcguire: so that we can save the model

features/src/main/scala/com/salesforce/op/utils/spark/OpVectorMetadata.scala
leahmcguire: it is an issue - that is why I put the todo

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -443,43 +444,69 @@ case object ModelInsights {
...
-        // first try out to get vector metadata from sanity checker
-        .flatMap(s => makeMeta(s.parent.asInstanceOf[SanityChecker]).orElse(makeMeta(s)))
-        // fall back to model selector stage metadata
-        .orElse(model.flatMap(m => makeMeta(m.parent.asInstanceOf[ModelSelector[_, _]])))
-        // finally try to get it from the last vector stage
-        .orElse(
-        stages.filter(_.getOutput().isSubtypeOf[OPVector]).lastOption
-          .map(v => OpVectorMetadata(v.getOutputFeatureName, v.getMetadata()))
-      )
+      if (sanityChecker.nonEmpty) { // first try out to get vector metadata from sanity checker
```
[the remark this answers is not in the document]
leahmcguire: I dont think it makes sense to have the flatten function return an option (also I am using an already defined function which has other uses here). I could make a reduce function to wrap the flatten and do a check but that makes the calculation less efficient...though I suppose these are small enough it probably doesnt matter. @tovbinm what is your vote?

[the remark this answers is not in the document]
tovbinm: The way `SanityCheckerSummary.flatten` is defined i.e. `def flatten(checkers: Seq[SanityCheckerSummary]): SanityCheckerSummary` - it should be renamed to `reduce`.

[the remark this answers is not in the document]
leahmcguire: Well I can rename it, then it is inconsistent with the OpVectorMetadata functions naming which does exactly the same thing. Also @Juantbox wants a change in the functionality so that it returns a None rather than an Empty summary, which I can easily do for the Sanity Checker but would change the usage of the OpVectorMetadata in a couple of places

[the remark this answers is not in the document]
tovbinm: My 2c: Empty summary (with all `0.0` values I presume) makes more sense to me. Option would create unnecessary clutter.

core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
```diff
@@ -201,25 +209,32 @@ object ForecastEvalMetrics extends Enum[ForecastEvalMetric] {
...
+    OpEvaluatorNames(name, humanName, largeBetter) {
     override def entryName: String = name.toLowerCase
   }
+  def withName(name: String, isLargerBetter: Boolean): OpEvaluatorNames =
+    Try(super.withName(name)).getOrElse(Custom(name, name, isLargerBetter))
+  override def withName(name: String): OpEvaluatorNames = withName(name, true)
+
+  def withNameInsensitive(name: String, isLargerBetter: Boolean): OpEvaluatorNames =
+    super.withNameInsensitiveOption(name).getOrElse(Custom(name, name, isLargerBetter))
+  override def withNameInsensitive(name: String): OpEvaluatorNames = withNameInsensitive(name, true)
```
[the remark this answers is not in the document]
leahmcguire: these are base methods @tovbinm , hence the override. if they want the fallback behavior your method provides they can use my definitions - but what happens when people call the base methods and there is no default?

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
```diff
@@ -0,0 +1,249 @@
...
+        )
+        setMetadata(summary.toMetadata().toSummaryMetadata())
+      }
+    }
+
+    val (metricValue1, metricValue2) = (getMet(metricValueOpt1), getMet(metricValueOpt2))
+
+    val strategy = getCombinationStrategy()
+    val (weight1, weight2) = strategy match {
+      case CombinationStrategy.Best =>
```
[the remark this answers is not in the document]
leahmcguire: it actually cant unless I move the EvalMetric class to the features module

core/src/main/scala/com/salesforce/op/stages/impl/selector/SelectedCombiner.scala
leahmcguire: So no we do not need a custom serializer - evaluators do not need to be serialized

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
leahmcguire: added custom to reflect if have multiple corr types

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -443,43 +444,69 @@ case object ModelInsights {
...
+      val stages = m.getInputFeatures().map(_.parentStages().toOption.map(_.keySet.map(_.uid)))
+      val uid = stages.collect{ case Some(uids) => uids }
+      uid.fold(Set.empty)(_ + _)
+    }.getOrElse(Set.empty)
+
+    val sanityCheckers = stages.collect { case s: SanityCheckerModel => s }
+    val sanityCheckersForModel = sanityCheckers.filter(s => modelInputStages.contains(s.uid) &&
+      model.exists(_.getInputFeature[RealNN](0) == s.getInputFeature[RealNN](0))).toSeq
+
+    val sanityChecker = if (sanityCheckersForModel.nonEmpty) sanityCheckersForModel else sanityCheckers.lastOption.toSeq
```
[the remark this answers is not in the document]
leahmcguire: this is not how these functions should work

leahmcguire:
@tovbinm @Jauntbox please take a look

PR #386, opened 2019-08-20
Make model selector metadata to metric more robust

erica-chiu:
**Related issues**

Change the matching when parsing model selector metadata into metrics from name of metric to metric keys

N/A

**Additional context**
Allows for parsing of metadata to be based on content rather than the name, allowing for multiple metrics of the same type

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorSummary.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorSummary.scala
@@ -240,19 +240,26 @@ case object ModelSelectorSummary {
     ReflectionUtils.classForName(className) match {
       case n if n == classOf[MultiMetrics] =>
         JsonUtils.fromString[Map[String, Map[String, Any]]](json).map{ d =>
-          val asMetrics = d.flatMap{ case (_, values) => values.map{
+          val asMetrics = d.flatMap{ case (_, values) =>
+            values.map{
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorSummary.scala
```diff
@@ -240,19 +240,26 @@ case object ModelSelectorSummary {
...
+                case `binary` =>
                   nm -> JsonUtils.fromString[BinaryClassificationMetrics](valsJson).get
-                case OpEvaluatorNames.BinScore.humanFriendlyName =>
+                case `binscore` =>
                   nm -> JsonUtils.fromString[BinaryClassificationBinMetrics](valsJson).get
-                case OpEvaluatorNames.Multi.humanFriendlyName =>
+                case `multi` =>
                   nm -> JsonUtils.fromString[MultiClassificationMetrics](valsJson).get
-                case OpEvaluatorNames.Regression.humanFriendlyName =>
+                case `regression` =>
```
leahmcguire: hmm, it is better not to have this hard coded as text, but I noticed a problem with matching in Enumns in my PR as well... @tovbinm is this related to the upgrade?

tovbinm: Weird. Looking at it...

tovbinm:
@erica-chiu proposed tests are defining the custom evaluator incorrectly. Here is how is supposed to work #387

PR #394, opened 2019-08-27
Added roadmap

tovbinm:
**Related issues**
We don't have a roadmap that we communicate with the community. Let's add one.

Added roadmap file + update a few other files.


tovbinm:
Is there anything else I should add here? please feel free to comment and/or edit.

PR #404, opened 2019-09-09
Ensure correct metrics despite model failures on some CV folds

leahmcguire:
**Related issues**
GLM optimizer is not very stable so sometimes runs will fail in some CV folds but not others - this was causing metrics and grids to not be computed correctly since we assumed that the same number of parameter grids successfully run in each cross validation fold 

throw out grids which are not in every cv run (since they are likely to fail in some modeling runs)

use grids with metrics in a map to make sure that the correct metrics are combined in CV and simply take the best metric

**Additional context**

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpCrossValidation.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpCrossValidation.scala
@@ -57,20 +57,19 @@ private[op] class OpCrossValidation[M <: Model[_], E <: Estimator[_]]
     folds: Seq[ValidatedModel[E]]
   ): ValidatedModel[E] = {
     require(folds.map(_.model.uid).toSet.size == 1) // Should be called only on instances of the same model
-    val gridCounts = folds.map(_.grids.map(_ -> 1).toMap).reduce(_ + _)
+
+    val gridCounts = folds.flatMap(_.grids.map(_ -> 1).toMap).sumByKey
     val (_, maxFolds) = gridCounts.maxBy{ case (_, count) => count }
     val gridsIn = gridCounts.filter{ case (_, foldCount) => foldCount == maxFolds }.keySet
-    val gridMetrics = folds.map(f => f.grids.zip(f.metrics).toMap)
-      .reduce{ (m1: Map[ParamMap, Double], m2: Map[ParamMap, Double]) =>
-        val keys = m1.keySet.union(m2.keySet)
-        keys.map(k => k -> (m1.getOrElse(k, 0.0) + m2.getOrElse(k, 0.0))).toMap
-      }
-      .filterKeys(gridsIn.contains)
-      .map{ case (key, met) => key -> met / maxFolds}
-      .toSeq
+
+    val gridMetrics = folds.flatMap{
+      f => f.grids.zip(f.metrics).collect { case (pm, met) if gridsIn.contains(pm) => (pm, met / maxFolds) }.toMap
+    }.sumByKey.toSeq
+
     val ((bestGrid, bestMetric), bestIndex) =
       if (evaluator.isLargerBetter) gridMetrics.zipWithIndex.maxBy{ case ((_, metric), _) => metric}
       else gridMetrics.zipWithIndex.minBy{ case ((_, metric), _) => metric}
+
     val ValidatedModel(est, _, _, _) = folds.head
     log.info(s"Average cross-validation for $est metrics: {}", gridMetrics.mkString(","))
     log.info(s"Best set of parameters:\n$bestGrid")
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpCrossValidation.scala
```diff
@@ -57,16 +57,23 @@ private[op] class OpCrossValidation[M <: Model[_], E <: Estimator[_]]
...
-    blas.dscal(metrics.length, 1.0 / numFolds, metrics, 1)
-    val ValidatedModel(est, _, _, grid) = folds.head
-    log.info(s"Average cross-validation for $est metrics: {}", metrics.toSeq.mkString(","))
-    val (bestMetric, bestIndex) =
-      if (evaluator.isLargerBetter) metrics.zipWithIndex.maxBy(_._1)
-      else metrics.zipWithIndex.minBy(_._1)
-    log.info(s"Best set of parameters:\n${grid(bestIndex)}")
+    require(folds.map(_.model.uid).toSet.size == 1) // Should be called only on instances of the same model
+    val gridCounts = folds.map(_.grids.map(_ -> 1).toMap).reduce(_ + _)
+    val maxFolds = gridCounts.maxBy(_._2)._2
```
tovbinm: For better readability please replace `_._2` in `val maxFolds = gridCounts.maxBy(_._2)._2` with `val (_, maxFolds) = gridCounts.maxBy { case (_, folds) => folds }`
Same in a few operations below.

leahmcguire:
@tovbinm @gerashegalov I discovered some fun behavior using algebird implicit for maps:

val test = Seq(Map("a" -> 1, "b" -> 1), Map("a" -> 1, "b" -> 1))
val combined = test.reduce( _ +  _) // Map("a" -> 2, "b" -> 2)) as expected

val test = Seq(Map("a" -> 0, "b" -> 0), Map("a" -> 0, "b" -> 0))
val combined = test.reduce( _ +  _) // Map()    :-(

Think I may need to write out the reduce function...

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpCrossValidation.scala
```diff
@@ -57,16 +56,27 @@ private[op] class OpCrossValidation[M <: Model[_], E <: Estimator[_]]
...
+    require(folds.map(_.model.uid).toSet.size == 1) // Should be called only on instances of the same model
+    val gridCounts = folds.map(_.grids.map(_ -> 1).toMap).reduce(_ + _)
+    val (_, maxFolds) = gridCounts.maxBy{ case (_, count) => count }
+    val gridsIn = gridCounts.filter{ case (_, foldCount) => foldCount == maxFolds }.keySet
+    val gridMetrics = folds.map(f => f.grids.zip(f.metrics).toMap)
+      .reduce{ (m1: Map[ParamMap, Double], m2: Map[ParamMap, Double]) =>
+        val keys = m1.keySet.union(m2.keySet)
+        keys.map(k => k -> (m1.getOrElse(k, 0.0) + m2.getOrElse(k, 0.0))).toMap
+      }
+      .filterKeys(gridsIn.contains)
```
gerashegalov: let us filter first, so we have less to reduce, maybe

```scala
    val gridMetrics = folds.flatMap(f => f.grids.zip(f.metrics))
      .collect { case (pm, met) if gridsIn.contains(pm) => (pm, met / maxFolds) }
      .sumByKey
```

tovbinm:
@leah Map behaviour is as expected under monoid rules. In order to keep the map keys with 0 values use semigroup for maps. I e.

implicit val mapSG = Semigroup.map[String, Long]

tovbinm:
@leahmcguire  Map behavior is as expected under monoid rules. In order to keep the map keys with 0 values use semigroup for map values. Example for `Map[String, Long]` below:
```scala
import com.twitter.algebird._

implicit val longSemigroup= Semigroup.from[Long](_ + _)
implicit val mapLongMonoid = Monoid.mapMonoid[String, Long](longSemigroup)

// works!
(Map("a" -> 0L) + Map("b" -> 0L)) shouldBe Map("a" -> 0L, "b" -> 0L)
```

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpCrossValidation.scala
```diff
@@ -57,16 +56,26 @@ private[op] class OpCrossValidation[M <: Model[_], E <: Estimator[_]]
...
   ): ValidatedModel[E] = {
-    val metrics = folds.map(_.metrics).reduce(_ + _)
-    blas.dscal(metrics.length, 1.0 / numFolds, metrics, 1)
-    val ValidatedModel(est, _, _, grid) = folds.head
-    log.info(s"Average cross-validation for $est metrics: {}", metrics.toSeq.mkString(","))
-    val (bestMetric, bestIndex) =
-      if (evaluator.isLargerBetter) metrics.zipWithIndex.maxBy(_._1)
-      else metrics.zipWithIndex.minBy(_._1)
-    log.info(s"Best set of parameters:\n${grid(bestIndex)}")
+    require(folds.map(_.model.uid).toSet.size == 1) // Should be called only on instances of the same model
```
gerashegalov: good to be defensive, my concern we call it in iteration in a private method completely in the scope here, and by construction we already know that folds are for the same model.

leahmcguire: Sure, I can remove and just put a description on the method

PR #407, opened 2019-09-12
Fix `ModelInsightsTest` flakiness due to coefficients turning negative 

TuanNguyen27:
[ModelInsightsTest](https://github.com/salesforce/TransmogrifAI/blob/95a77b17269a71bf0d53c54df7d76f0bfe862275/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala#L750-L778) was flaky due to returned coefficients being negative sometimes.

Adding absolute value when summing descaled coefficients


gerashegalov (approved):
LGTM

Scutua (approved):

PR #412, opened 2019-09-30
InsightLOCOTest Refactor 

AdamChit:
**Related issues**
Long test that needs to be refactored with the main goal of making it more readable and breaking it up into separate tests

First Change from FlatSpec to [FunSpec](http://doc.scalatest.org/1.7.1/org/scalatest/FunSpec.html) then break up the test into 5 different test cases 

could of put the data generation code into a function, but generating data is expensive so we wouldn't want to make repeated calls to it

### DIFF
--- a/core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
@@ -96,7 +96,7 @@ class RecordInsightsLOCOTest extends FunSpec with TestSparkContext with RecordIn
     )
   )
   // scalastyle:on
-  it("should work with randomly generated features and binary logistic regression") {
+  it ("should work with randomly generated features and binary logistic regression") {
     val features = RandomVector.sparse(RandomReal.normal(), 40).limit(1000)
     val labels = RandomIntegral.integrals(0, 2).limit(1000).map(_.value.get.toRealNN)
     val (df, f1, l1) = TestFeatureBuilder("features", "labels", features.zip(labels))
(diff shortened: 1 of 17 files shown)


core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -404,10 +411,8 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext with RecordI
    *
    * @param predicate  predicate used by RecordInsights in order to aggregate
    */
-  private def assertAggregatedWithPredicate(
-    predicate: OpVectorColumnHistory => Boolean,
-    testData: RecordInsightsTestData[LogisticRegressionModel]
-  ): Unit = {
+  private def assertAggregatedWithPredicate(predicate: OpVectorColumnHistory => Boolean,
+                                            testData: RecordInsightsTestData[LogisticRegressionModel]): Unit = {
```
crupley: Can you revert the formatting here to the way it was before to match scala style guidelines?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -202,7 +202,8 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext with RecordI
     }
   }
 
-  it should "return the most predictive features for data generated with a strong relation to the label" in {
+  describe("data strongly related to label. ") {
```
crupley: This description doesn't seem to match the following section. Can you add something more descriptive?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -96,7 +96,7 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext with RecordI
     )
   )
   // scalastyle:on
-  Spec[RecordInsightsLOCO[_]] should "work with randomly generated features and binary logistic regression" in {
+  it ("should work with randomly generated features and binary logistic regression") {
```
tovbinm: replace `it` with `Spec[RecordInsightsLOCO[_]]`

AdamChit: Can't directly do that with FunSpec, what worked best was wrapping `describe(Spec[RecordInsightsLOCO[_]])`  around the tests.

What the updated tests look like:

Screen Shot 2019-10-07 at 10 32 14 AM

AdamChit:
code 
https://codecov.io/gh/salesforce/TransmogrifAI/compare/a83bf4639b2a9324ae27d1d49edf66a6245f6ac0...a4a0e653b6d1437dbef7a3b388baa740834c0b58/changes

PR #413, opened 2019-10-03
Regression training limit

AdamChit:
**Related issues**

DataBalancer for binary classification has a parameter that controls the max data passed into modeling - Regression should allow similar limits

Steps:
1. Investigate where the check should occur (somewhere in DataSplitter)

2. Add logic to downsample when the number of records is reached 

3. Add downsampling information to the summary object and log

4. Add tests to DataSplitterTest and RegressionModelSelectorTest to cover the downsampling logic

Not having a limit for any of the model types - this was not optimal because some spark models may have very long runtimes or bad behavior with too much data. So the default will be to downsample once we have passes 1M records and give the user the option to set their own maxTrainingSample if they are ok with working with large dataset

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
@@ -73,9 +73,9 @@ class DataSplitter(uid: String = UID[DataSplitter]) extends Splitter(uid = uid) 
   override def preValidationPrepare(data: Dataset[Row]): PrevalidationVal = {
     val dataSetSize = data.count().toDouble
     val sampleF = getMaxTrainingSample / dataSetSize
-    val DownSampleFraction = if (getMaxTrainingSample < dataSetSize) sampleF else 1
-    summary = Option(DataSplitterSummary(DownSampleFraction))
-    setDownSampleFraction(DownSampleFraction)
+    val downSampleFraction = math.min(sampleF, SplitterParamsDefault.DownSampleFractionDefault)
+    summary = Option(DataSplitterSummary(downSampleFraction))
+    setDownSampleFraction(downSampleFraction)
     PrevalidationVal(summary, None)
   }
 
@@ -114,6 +114,7 @@ trait DataSplitterParams extends Params {
       lowerBound = 0.0, upperBound = 1.0, lowerInclusive = false, upperInclusive = true
     )
   )
+  setDefault(downSampleFraction, SplitterParamsDefault.DownSampleFractionDefault)
 
   private[op] def setDownSampleFraction(value: Double): this.type = set(downSampleFraction, value)
 
--- a/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
@@ -73,7 +73,7 @@ class DataSplitter(uid: String = UID[DataSplitter]) extends Splitter(uid = uid) 
   override def preValidationPrepare(data: Dataset[Row]): PrevalidationVal = {
     val dataSetSize = data.count().toDouble
     val sampleF = getMaxTrainingSample / dataSetSize
-    val DownSampleFraction = if (sampleF < 1) sampleF else 1
+    val DownSampleFraction = if (getMaxTrainingSample < dataSetSize) sampleF else 1
     summary = Option(DataSplitterSummary(DownSampleFraction))
     setDownSampleFraction(DownSampleFraction)
     PrevalidationVal(summary, None)
--- a/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
@@ -71,10 +71,10 @@ class DataSplitter(uid: String = UID[DataSplitter]) extends Splitter(uid = uid) 
    * @return Parameters set in examining data
    */
   override def preValidationPrepare(data: Dataset[Row]): PrevalidationVal = {
-    val dataSetSize = data.count().toDouble
-    val sampleF = getMaxTrainingSample / dataSetSize
+    val dataSetSize = data.count()
+    val sampleF = getMaxTrainingSample / dataSetSize.toDouble
     val downSampleFraction = math.min(sampleF, SplitterParamsDefault.DownSampleFractionDefault)
-    summary = Option(DataSplitterSummary(downSampleFraction))
+    summary = Option(DataSplitterSummary(dataSetSize, downSampleFraction))
     setDownSampleFraction(downSampleFraction)
     PrevalidationVal(summary, None)
   }
@@ -125,7 +125,7 @@ trait DataSplitterParams extends Params {
  * Summary for data splitter run for storage in metadata
  * @param downSamplingFraction down sampling fraction for training set
  */
-case class DataSplitterSummary(downSamplingFraction: Double) extends SplitterSummary {
+case class DataSplitterSummary(preSplitterDataCount: Long, downSamplingFraction: Double) extends SplitterSummary {
 
   /**
    * Converts to [[Metadata]]
@@ -137,6 +137,7 @@ case class DataSplitterSummary(downSamplingFraction: Double) extends SplitterSum
   def toMetadata(skipUnsupported: Boolean): Metadata = {
     new MetadataBuilder()
       .putString(SplitterSummary.ClassName, this.getClass.getName)
+      .putLong(ModelSelectorNames.PreSplitterDataCount, preSplitterDataCount)
       .putDouble(ModelSelectorNames.DownSample, downSamplingFraction)
       .build()
   }
--- a/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
@@ -31,8 +31,9 @@
 package com.salesforce.op.stages.impl.tuning
 
 import com.salesforce.op.UID
+import com.salesforce.op.stages.impl.selector.ModelSelectorNames
 import org.apache.spark.ml.param._
-import org.apache.spark.sql.{Dataset, Row}
+import org.apache.spark.sql.{DataFrame, Dataset, Row}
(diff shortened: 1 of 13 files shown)


AdamChit:
@TuanNguyen27 Could you review

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
```diff
@@ -59,30 +62,69 @@ case object DataSplitter {
...
+      dataPrep
+    }
+    balanced.persist()
+  }
   override def copy(extra: ParamMap): DataSplitter = {
     val copy = new DataSplitter(uid)
     copyValues(copy, extra)
   }
 }
+trait DataSplitterParams extends Params {
```
leahmcguire: Why are we putting down sample fraction as a settable parameter? - this is an internal parameter only it is used in data rebalancing if the target fraction is set

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
```diff
@@ -59,30 +62,69 @@ case object DataSplitter {
...
+  /**
+   * Fraction to down sample data
+   * Value should be in [0.0, 1.0]
+   *
+   * @group param
+   */
+  private[op] final val downSampleFraction = new DoubleParam(this, "downSampleFraction",
```
leahmcguire: these can be protected instead of private to OP

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
```diff
@@ -46,11 +47,13 @@ case object DataSplitter {
...
   def apply(
     seed: Long = SplitterParamsDefault.seedDefault,
-    reserveTestFraction: Double = SplitterParamsDefault.ReserveTestFractionDefault
+    reserveTestFraction: Double = SplitterParamsDefault.ReserveTestFractionDefault,
+    maxTrainingSample: Int = SplitterParamsDefault.MaxTrainingSampleDefault
   ): DataSplitter = {
     new DataSplitter()
       .setSeed(seed)
       .setReserveTestFraction(reserveTestFraction)
+      .setMaxTrainingSample(maxTrainingSample)
```
leahmcguire: is this also exposed for the datacutter class?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
leahmcguire: also maybe set a default value of 1

core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala
```diff
@@ -120,9 +121,32 @@ class RegressionModelSelectorTest extends FlatSpec with TestSparkContext
...
     val modelSelector = RegressionModelSelector()
-    modelSelector.splitter.get.setReserveTestFraction(0.1).setSeed(11L)
+    modelSelector.splitter.get.setReserveTestFraction(0.1).setSeed(11L).setMaxTrainingSample(1000)
+
     modelSelector.splitter.get.getSeed shouldBe 11L
     modelSelector.splitter.get.getReserveTestFraction shouldBe 0.1
+    modelSelector.splitter.get.getMaxTrainingSample shouldBe 1000
+  }
+
+  it should "set maxTrainingSample and down-sample" in {
```
crupley: Did you also mean to check `maxTrainingSample` was set in this test as well?

core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala
```diff
@@ -120,9 +121,32 @@ class RegressionModelSelectorTest extends FlatSpec with TestSparkContext
...
+
+    implicit val vectorEncoder: org.apache.spark.sql.Encoder[Vector] = ExpressionEncoder()
+    implicit val e1 = Encoders.tuple(Encoders.scalaDouble, vectorEncoder)
+    val maxTrainingSample = 100
+    val sampleF = maxTrainingSample / dataCount.toDouble
+    val downSampleFraction = if (maxTrainingSample < dataCount) sampleF else 1
```
crupley: this might be more clear as a `min`
```suggestion
    val downSampleFraction = math.min(sampleF, 1.0)
```

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
```diff
@@ -59,30 +62,69 @@ case object DataSplitter {
...
+   * Function to set the down sampling fraction and parameters before passing into the validation step
    *
    * @param data
    * @return Parameters set in examining data
    */
   override def preValidationPrepare(data: Dataset[Row]): PrevalidationVal = {
-    summary = Option(DataSplitterSummary())
+    val dataSetSize = data.count().toDouble
+    val sampleF = getMaxTrainingSample / dataSetSize
+    val DownSampleFraction = if (getMaxTrainingSample < dataSetSize) sampleF else 1
```
crupley: could be a `min` here too

also: lowerCamelCase

```suggestion
    val downSampleFraction = math.min(sampleF, 1.0)
```

core/src/test/scala/com/salesforce/op/stages/impl/tuning/DataSplitterTest.scala
```diff
@@ -71,8 +98,10 @@ class DataSplitterTest extends FlatSpec with TestSparkContext with SplitterSumma
   it should "keep the data unchanged when prepare is called" in {
     val summary = dataSplitter.preValidationPrepare(data)
     val train = dataSplitter.validationPrepare(data)
+    val sampleF = MaxTrainingSampleDefault / dataCount.toDouble
+    val downSampleFraction = if (sampleF < 1) sampleF else 1
```
crupley:
```suggestion
    val downSampleFraction = math.min(sampleF, 1.0)
```

core/src/test/scala/com/salesforce/op/stages/impl/tuning/DataSplitterTest.scala
```diff
@@ -43,6 +43,7 @@ class DataSplitterTest extends FlatSpec with TestSparkContext with SplitterSumma
 
   val seed = 1234L
   val dataCount = 1000
+  val MaxTrainingSampleDefault = 1E6.toLong
```
crupley: I would give this a different name so we don't confuse it with the actual default in SplitterParams

core/src/test/scala/com/salesforce/op/stages/impl/tuning/DataSplitterTest.scala
```diff
@@ -56,6 +57,32 @@ class DataSplitterTest extends FlatSpec with TestSparkContext with SplitterSumma
...
+    val numRows = 2000
+    val maxRows = numRows / 2
+
+    dataSplitter
+      .setReserveTestFraction(0.0)
+      .setMaxTrainingSample(maxRows)
+
+    dataSplitter.getMaxTrainingSample shouldBe maxRows
+  }
+
```
crupley: It's probably worth checking `downSampleFraction` params was set here as well, for completeness so that you have checked everything in `DataSplitterParams`

AdamChit: made changes here 433d4836b19bdc74596708e3f5de5061cfc7d242

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
AdamChit: Yes, I added it to SplitterParams which datacutter has access to - e4b8a9219a73291d9e1dbc22274ef97f46898f25. So that I can use the same set/get functions across DataBalancer, DataCutter and DataSplitter.

core/src/test/scala/com/salesforce/op/stages/impl/regression/RegressionModelSelectorTest.scala
AdamChit: I moved testing the maxTrainingSample set/get functions to [here](https://github.com/salesforce/TransmogrifAI/pull/413/files/8e2778da356242d0cd10d259282fb72709cbba8c#diff-a915d7f576999b5aa0cfe9b9d3e73d9bR128) and forgot to rename this test. Great catch!

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataSplitter.scala
AdamChit: Added the default value [here](https://github.com/salesforce/TransmogrifAI/pull/413/commits/962e06f605a56453fe33a5d940043f19c7cdddd4#diff-178e84a5a5354047d11da1e11cbedcd3R117)

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
```diff
@@ -152,7 +171,10 @@ private[op] object SplitterSummary {
 
   def fromMetadata(metadata: Metadata): Try[SplitterSummary] = Try {
     metadata.getString(ClassName) match {
-      case s if s == classOf[DataSplitterSummary].getName => DataSplitterSummary()
+      case s if s == classOf[DataSplitterSummary].getName => DataSplitterSummary(
+        preSplitterDataCount = metadata.getLong(ModelSelectorNames.PreSplitterDataCount),
+        downSamplingFraction = metadata.getDouble(ModelSelectorNames.DownSample)
+      )
```
leahmcguire: please add the downsample fraction to the datacutter params as well...

AdamChit: I added downsample fraction into the datacutter params as part of the multi class classification training limit changes. I'll create the PR for it today.

AdamChit:
Unrelated Test fails during the Travis CI check https://travis-ci.com/salesforce/TransmogrifAI/jobs/243091644 I'll create a ticket to increase the tolerance on the test [here](https://github.com/salesforce/TransmogrifAI/blob/master/utils/src/test/scala/com/salesforce/op/utils/stats/OpStatisticsTest.scala#L203)

PR #414, opened 2019-10-08
Multi-class classification training limit

AdamChit:
**Related issues**
DataBalancer for binary classification has a parameter that controls the max data passed into modeling - multiclass classification should allow similar limits

The solution is to downsample once we reach the training set limit

N/A

**Additional context**
similar to https://github.com/salesforce/TransmogrifAI/pull/413

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
@@ -153,7 +153,7 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
     val dataPrep = super.validationPrepare(data)
 
     // check if down sampling is needed
-    val balanced: DataFrame = if (getDownSampleFraction < 1) {
+    val balanced: DataFrame = if (getDownSampleFraction < 1.0) {
       dataPrep.sample( false, getDownSampleFraction, getSeed)
     } else {
       dataPrep
@@ -307,7 +307,7 @@ private[impl] trait DataCutterParams extends SplitterParams {
  */
 case class DataCutterSummary
 (
-  preSplitterDataCount: Long = 0,
+  preSplitterDataCount: Long = 0L,
   downSamplingFraction: Double = SplitterParamsDefault.DownSampleFractionDefault,
   labelsKept: Seq[Double],
   labelsDropped: Seq[Double],
--- a/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
@@ -34,7 +34,7 @@ import com.salesforce.op.UID
 import com.salesforce.op.stages.impl.selector.ModelSelectorNames
 import org.apache.spark.ml.attribute.{MetadataHelper, NominalAttribute}
 import org.apache.spark.ml.param._
-import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.{DataFrame, Dataset, Row}
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types.{Metadata, MetadataBuilder}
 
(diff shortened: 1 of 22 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -123,14 +126,41 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+    setDownSampleFraction(downSampleFraction)
+
     summary = Option(DataCutterSummary(
+      preSplitterDataCount = dataSetSize,
+      downSamplingFraction = getDownSampleFraction,
       labelsKept = getLabelsToKeep,
       labelsDropped = getLabelsToDrop,
       labelsDroppedTotal = getLabelsDroppedTotal
     ))
     PrevalidationVal(summary, Option(dataPrep))
```
leahmcguire: I think it makes more sense to do the downsampling in the pre-validationPrepare than in the validation prepare - the difference being that the validation prepare is called within the CV folds. So since you upsample for binary it needs to be here to prevent label leakage but since there is only downsampling here it can go earlier

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -203,7 +233,11 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
         s" minLabelFraction = $minLabelFract, maxLabelCategories = $maxLabels. \n" +
         s"Label counts were: ${labelCounts.collect().toSeq}")
     }
-    DataCutterSummary(labelsKept.toSeq, labelsDropped.toSeq, labelsDroppedTotal.toLong)
+    DataCutterSummary(
+      labelsKept = labelsKept.toSeq,
```
leahmcguire: do you want to add the downsample fraction?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -123,14 +126,41 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
...
+   *
+   * @param data to prepare for model training. first column must be the label as a double
+   * @return balanced training set and a test set
+   */
+  override def validationPrepare(data: Dataset[Row]): Dataset[Row] = {
+
+    val dataPrep = super.validationPrepare(data)
+
+    // check if down sampling is needed
+    val balanced: DataFrame = if (getDownSampleFraction < 1) {
```
tovbinm: use `1.0`

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -273,6 +307,8 @@ private[impl] trait DataCutterParams extends SplitterParams {
  */
 case class DataCutterSummary
 (
+  preSplitterDataCount: Long = 0,
```
tovbinm: use `0L`

core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala
```diff
@@ -129,6 +129,23 @@ trait SplitterParams extends Params {
   def setReserveTestFraction(value: Double): this.type = set(reserveTestFraction, value)
   def getReserveTestFraction: Double = $(reserveTestFraction)
 
+  /**
+   * Fraction to sample majority data
+   * Value should be in ]0.0, 1.0]
```
tovbinm: `(0.0, 1.0]`

core/src/test/scala/com/salesforce/op/stages/impl/tuning/DataCutterTest.scala
```diff
@@ -65,6 +68,8 @@ class DataCutterTest extends FlatSpec with TestSparkContext with SplitterSummary
       s.labelsKept.length shouldBe 1000
       s.labelsDropped.length shouldBe 0
       s shouldBe DataCutterSummary(
+        preSplitterDataCount = dataSize,
+        downSamplingFraction = 1,
```
tovbinm: Use `1.0` instead of `1` here and everywhere with Double values

tovbinm (approved):
see minor comments

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
AdamChit: The 2 main reasons behind doing it this way:
1. Keeping it consistent with dataSplitter/dataBalancer
2. Easier to implement stratified sampling (or other data balancing techniques) in the future which would upsample the minority classes

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
AdamChit: I didn't want to have to pass the dataset (or the dataset count) to the [estimate](https://github.com/salesforce/TransmogrifAI/blob/213962e4488b43d19fc9c6dc11da1b6869233b12/core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala#L170) function. So I added the dataset count and downsample fraction in the summary variable [here](https://github.com/salesforce/TransmogrifAI/pull/414/files/5aba86cbc9c4bd85551439170daf531d16193afb#diff-b36278615b6efb5334827d4ce645c916L126).

Scutua (approved):

PR #416, opened 2019-10-10
Turn off logging for cardEstimate & track token length instead of token value

TuanNguyen27:
Tracking most frequent token value is not feasible with TextStats. Default to track token length instead. Also turned off logging for cardEstimate to keep log shorter


leahmcguire:
why is it not feasible to track the tokens?

TuanNguyen27:
> why is it not feasible to track the tokens?

`TextStats` stops tracking the count of existing tokens inside `valueCounts` when the size of `valueCounts` exceed `RawFeatureFilter.MaxCardinality`, hence the frequency of token values are always undercounted

PR #418, opened 2019-10-11
Option to calculate LOCO for dates/texts by Leaving Out Entire Vector.

sanmitra:
**Related issues**
Raw features of types - texts and dates are converted to vectors during feature engineering. To calculate locos for such features, we calculate loco for each element in the vector and then average them. So, the complexity of calculating LOCO for each date/text feature is **O(n*m)** 
where n is the size of entire feature vector (i.e containing all the features) and m is the size of individual text/date feature vector.

Alternate way to calculate the locos for date/text feature is zero out the entire vector of that feature and then calculate loco. So, the complexity of calculating LOCO for each date/text feature is  **O(n)**. The enum for this new approach is `LeaveOutVector` and old approach is `Avg`, and user one can choose between them.


core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -168,82 +192,85 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+    strategy match {
+      case VectorAggregationStrategy.Avg =>
+        aggIndices
+          .map { case (i, oldInd) => computeDiff(featureSparse.copy.updated(i, oldInd, 0.0), baseScore) }
+          .foldLeft(Array.empty[Double])(sumArrays)
+          .map( _ / featureCount)
+
+      case VectorAggregationStrategy.LeaveOutVector =>
+        val copyFeatureSparse = featureSparse.copy
+        aggIndices.map {case (i, oldInd) => copyFeatureSparse.updated(i, oldInd, 0.0)}
```
michaelweilsalesforce: Maybe `foreach` instead of `map`?
Also doesn't `copyFeatureSparse` need to be a `var`?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -406,34 +450,46 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext with RecordI
...
+              val oldVal = v(i)
+              featureArray.update(i, 0.0)
+              val newScore = model.transformFn(l.toRealNN, featureArray.toOPVector).score.toSeq
+              featureArray.update(i, oldVal)
+              baseScore.zip(newScore).map { case (b, n) => b - n }
+            }
+            val sumLOCOs = locos.reduce((a1, a2) => a1.zip(a2).map { case (l, r) => l + r })
+            sumLOCOs.map(_ / indices.length)
+          case VectorAggregationStrategy.LeaveOutVector =>
+            indices.map { i => featureArray.update(i, 0.0) }
```
michaelweilsalesforce: `foreach` also here

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -406,34 +450,46 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext with RecordI
...
+            val newScore = model.transformFn(l.toRealNN, featureArray.toOPVector).score.toSeq
+            baseScore.zip(newScore).map { case (b, n) => b - n }
         }
```
michaelweilsalesforce: Don't you need to revert `featureArray` with the old vals ?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -168,82 +192,85 @@ class RecordInsightsLOCO[T <: Model[T]]
...
     val baseResult = modelApply(labelDummy, features)
     val baseScore = baseResult.score
-    val featureSize = features.value.size
 
     // TODO: sparse implementation only works if changing values to zero - use dense vector to test effect of zeros
     val featuresSparse = features.value.toSparse
-    val featureIndexSet = featuresSparse.indices.toSet
-
-    // Besides non 0 values, we want to check the text/date features as well
-    val zeroValIndices = (textFeatureIndices ++ dateFeatureIndices)
```
michaelweilsalesforce: How did you manage to remove the zero val indices logic?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
sanmitra: because its a copy see line `471` `val featureArray = v.copy.toArray`

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: Sure, i will use `foreach`
But I am not sure about whether i should make it `var`. Reassignment of `val` is not allowed, but object can have its internal state modified, So guess `val` is fine.

```
val a = Array(10, 20)
a.update(0, 15) // works
a = Array(20, 30) // fails
```

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: From `featureSparse` vector we know the active indices count, but to calculate the `average` LOCO for each date/text field, we needed the zero val indices logic. We used to calculate the total count of indices for a date/text feature in each transformation of a individual record/row in `transformFn` function. There is no need to do this, we can just calculate the total count of indices per date/text feature only once using `OpVectorColumnHistory` at the global level (i.e outside `transformFn` function ), this is what I did in line `139`

```
private lazy val textFeaturesCount: Map[String, Int] = getFeatureCount(isTextIndex)
private lazy val dateFeaturesCount: Map[String, Int] = getFeatureCount(isDateIndex) 
```

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -406,34 +450,46 @@ class RecordInsightsLOCOTest extends FlatSpec with TestSparkContext with RecordI
...
-          baseScore.zip(newScore).map { case (b, n) => b - n }
+        val featureArray = v.copy.toArray
+        val baseScore = model.transformFn(l.toRealNN, v.toOPVector).score.toSeq
+        strategy match {
+          case VectorAggregationStrategy.Avg =>
+            val locos = indices.map { i =>
+              val oldVal = v(i)
+              featureArray.update(i, 0.0)
+              val newScore = model.transformFn(l.toRealNN, featureArray.toOPVector).score.toSeq
+              featureArray.update(i, oldVal)
```
michaelweilsalesforce: Then why do you update back, since `featureArray` is a copy?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -64,9 +64,32 @@ trait RecordInsightsLOCOParams extends Params {
...
+
+  final val textAggregationStrategy = new Param[String](parent = this, name = "dateAggregationStrategy",
+    doc = "Whether text vector is aggregated by " +
+      "1. LeaveOutVector strategy - calculate the loco by leaving out the entire vector or" +
+      "2. Avg strategy - calculate the loco for each column of the vector and then average all the locos."
+  )
+  def setTextAggregationStrategy(strategy: VectorAggregationStrategy): this.type =
+    set(textAggregationStrategy, strategy.entryName)
+  def getTextAggregationStrategy: VectorAggregationStrategy = VectorAggregationStrategy.withName(
+    $(textAggregationStrategy))
```
leahmcguire: do we really want/need to expose strategies for each feature type? I imagine that there are other types of features me may eventually want to control as either average or leave out vector and putting in individual settings for each seems excessive...maybe one parameter to control how ALL vector treated features are handled?

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
sanmitra: Here, we have to because featureArray is a copy created before we calculate loco for each element, i.e
```
val featureArray = v.copy.toArray
val locos = indices.map { i =>
   featureArray.update(i, 0.0)
   calculateLOCO(featureArray, ....)
   featureArray.update(i, 0.0)
}
```

If we refactor the above code like below, then we don't need to update back.

```
val locos = indices.map { i =>
  val featureArray = v.copy.toArray
  featureArray.update(i, 0.0)
  calculateLOCO(featureArray,...)
}
```

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: Yeah, I agree. One parameter should be enough.

sanmitra: done

sanmitra:
@tovbinm can you please take a look at this PR ?

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -113,21 +125,22 @@ class RecordInsightsLOCO[T <: Model[T]]
   private val dateMapTypes =
     Set(FeatureType.typeName[DateMap], FeatureType.typeName[DateTimeMap])
 
-  // Indices of features derived from hashed Text(Map)Vectorizer
-  private lazy val textFeatureIndices: Seq[Int] = getIndicesOfFeatureType(textTypes ++ textMapTypes,
-    h => h.indicatorValue.isEmpty && h.descriptorValue.isEmpty)
+  // Map[FeatureName(Date/Text), VectorSize]
```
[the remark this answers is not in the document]
sanmitra: done.

sanmitra:
@tovbinm @gerashegalov I have addressed your review comments. Thanks

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -168,82 +183,72 @@ class RecordInsightsLOCO[T <: Model[T]]
...
+    strategy match {
+      case VectorAggregationStrategy.Avg =>
+        aggIndices
+          .map { case (i, oldInd) => computeDiff(featureSparse.copy.updated(i, oldInd, 0.0), baseScore) }
+          .foldLeft(Array.empty[Double])(sumArrays)
+          .map( _ / featureSize)
+
+      case VectorAggregationStrategy.LeaveOutVector =>
+        val copyFeatureSparse = featureSparse.copy
+        aggIndices.foreach {case (i, oldInd) => copyFeatureSparse.updated(i, oldInd, 0.0)}
```
tovbinm: why are you using `updated` here?

tovbinm: I think this should be `aggIndices.foreach {case (i, oldInd) => copyFeatureSparse.update(i, oldInd, 0.0)}`

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
```diff
@@ -168,82 +183,72 @@ class RecordInsightsLOCO[T <: Model[T]]
     }
   }
 
+  private def aggregateDiffs(
+    featureSparse: SparseVector,
+    aggIndices: Array[(Int, Int)],
+    strategy: VectorAggregationStrategy,
+    baseScore: Array[Double],
+    featureSize: Int
+  ): Array[Double] = {
```
sanmitra: The cost of computeDiff is O(n). 
When `VectorAggregationStrategy=Avg` note that, we do computeDiff **m** times and  
When `VectorAggregationStrategy=LeaveOutVector`, we do computeDiff only 1 time.
So the former computation is O(n*m) and the latter is O(n)

where n is the size of entire feature vector (i.e containing all the features) and m is the size of individual text/date feature vector which is to be aggregated,

core/src/main/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCO.scala
sanmitra: The idea is to zero out the entire vector and then calculate the loco in case of `VectorAggregationStrategy.LeaveOutVector`

sanmitra: And in case of `VectorAggregationStrategy.Avg` the idea is to calculate the loco by zeroing out each column of the date/text feature vector and then average out all the locos calculated.

tovbinm: well, `updated` returns the updated array. I think you want to modify in place (see my previous comment)

sanmitra: I think it does modify in place and returns the same array. Below the definition of that method -

```implicit class RichSparseVector(val v: SparseVector) extends AnyVal {
    def updated(index: Int, indexVal: Int, value: Double): SparseVector = {
      require(v.indices(index) == indexVal,
        s"Invalid index: indices($index)==${v.indices(index)}, expected: $indexVal")
      v.values(index) = value
      v
    }
  }```

sanmitra:
@tovbinm can you please merge this PR ?

tovbinm:
@sanmitra build timed out. restarted.

tovbinm:
@sanmitra the build is failing on master branch due to LOCO test failure. please have a look. thanks.

PR #420, opened 2019-10-11
removed logging of tokens

leahmcguire:
**Related issues**

**Additional context**


core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -167,7 +167,6 @@ case class FeatureDistribution
       "nulls" -> nulls.toString,
       "distribution" -> distribution.mkString("[", ",", "]"),
       "summaryInfo" -> summaryInfo.mkString("[", ",", "]"),
-      "cardinality" -> cardEstimate.map(_.toString).getOrElse(""),
```
gerashegalov: @leahmcguire was there any value in what @TuanNguyen27 was doing in https://github.com/salesforce/TransmogrifAI/commit/e0ea2ed6db1d03093025b88e343bccbbe86d7314#diff-528878f8bd336b485dd82f1fbef950f7R281

```
seq.map(_.size.toString)
```
which we could still do in `toString `

Scutua (approved):

PR #423, opened 2019-10-16
Achit/loco test bug fix

AdamChit:
**Related issues**
test refactor made in https://github.com/salesforce/TransmogrifAI/pull/412 causes one of the test cases to fail.

Specially this test assumed that at-least 1 of the 4 features would be selected https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala#L236

Test should allow the case where none of the features are picked

**Additional context**
example of the test failing:
Screen Shot 2019-10-15 at 9 24 36 PM

### DIFF
--- a/core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
+++ b/core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
@@ -96,165 +96,153 @@ class RecordInsightsLOCOTest extends FunSpec with TestSparkContext with RecordIn
     )
   )
   // scalastyle:on
-  it ("should work with randomly generated features and binary logistic regression") {
-    val features = RandomVector.sparse(RandomReal.normal(), 40).limit(1000)
-    val labels = RandomIntegral.integrals(0, 2).limit(1000).map(_.value.get.toRealNN)
-    val (df, f1, l1) = TestFeatureBuilder("features", "labels", features.zip(labels))
-    val l1r = l1.copy(isResponse = true)
-    val dfWithMeta = addMetaData(df, "features", 40)
-    val sparkModel = new OpLogisticRegression().setInput(l1r, f1).fit(df)
+  describe(Spec[RecordInsightsLOCO[_]]) {
+    it("should work with randomly generated features and binary logistic regression") {
+      val features = RandomVector.sparse(RandomReal.normal(), 40).limit(1000)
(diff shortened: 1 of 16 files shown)


core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -233,8 +233,8 @@ class RecordInsightsLOCOTest extends FunSpec with TestSparkContext with RecordIn
       info("Each feature vector should only have either three or four non-zero entries. One each from country and " +
         "picklist, while currency can have either two (if it's null the currency column will be filled with the mean)" +
         " or just one if it's not null.")
-      it("should pick between 1 and 4 of the features") {
-        all(parsed.map(_.size)) should (be >= 1 and be <= 4)
+      it("should pick between 0 and 4 features") {
+        all(parsed.map(_.size)) should (be >= 0 and be <= 4)
```
leahmcguire: this seems like it is not checking very much... can you do a check for the conditions described above and then do the size check?

AdamChit:
> Why would it not pick any features?

So in the test the labels are generated based on the PickList feature:
```val labelData: Seq[RealNN] = pickListData.map(p =>
        p.value match {
          case Some("A") | Some("B") | Some("C") => RealNN(1.0)
          case _ => RealNN(0.0)
        }
      )
```
And all the features have a chance of being empty:
```
      val countryData: Seq[Country] = RandomText.countries.withProbabilityOfEmpty(0.3).take(numRows).toList
      val pickListData: Seq[PickList] = RandomText.pickLists(domain = List("A", "B", "C", "D", "E", "F", "G"))
        .withProbabilityOfEmpty(0.1).limit(numRows)
      val currencyData: Seq[Currency] = RandomReal.logNormal[Currency](mean = 10.0, sigma = 1.0)
        .withProbabilityOfEmpty(0.3).limit(numRows)
```

One case that could happen: 
2) All the feature are set to empty
3) LOCO changes the pickList from empty to non-empty and because it is not in the set {A,B,C} it is **NOT** a strong of a predictor
=> label wouldn't change => pickList would not have a strong insight => No feature is selected

Note: the chance of this happening is very low which explains why it doesn't always appear in the CI build.

tovbinm:
@AdamChit are you still working on adding a more robust test?

tovbinm:
any updates? @AdamChit

PR #424, opened 2019-10-16
added call of validation prepare before model selection when no dag i…

leahmcguire:
…s passed

**Related issues**

**Additional context**


tovbinm:
Is there a way to test it? Should a user be able to disable it if needed?

leahmcguire:
You disable by not having a splitter. And it is not really testable in our current config since it was a bug in the internals of the validation model selection which may or may not effect which model wins but would not effect the final trained model since the selected model did have this stage applied correctly

gerashegalov:
Oh this is what I was looking in #246 but then it turned out a non-issue in my investigation. We need indeed this covered by tests (re @tovbinm) given the splitter lifecycle issues we have faced.

PR #426, opened 2019-10-17
Make EmailVectorizer not clean the email domains by default.

sanmitra:
**Related issues**
Users want to see the text of the email domain as an indicator variable, the way a true email address does (including punctuation)
EG: Today, email field marc.benioff@salesforce.com will have a column with indicator value salesforcecom, because the text of the domain name has been cleaned. Instead we would like it say "salesforce.com"

Added a `case class CleanTextParams(ignoreCase: Boolean, cleanPunctuations: Boolean)` which will give us more control on how the text is going to cleaned in general. In future more parameters can be added. 
By default across all features, this would be `CleanTextParams(true, true)` except for email/emailMap features, in which it would be `CleanTextParams(true, false)`

### DIFF
--- a/core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
@@ -33,7 +33,7 @@ package com.salesforce.op.dsl
 import com.salesforce.op.features.FeatureLike
 import com.salesforce.op.features.types._
 import com.salesforce.op.stages.impl.feature._
-import com.salesforce.op.utils.text.Language
+import com.salesforce.op.utils.text.{CleanTextParams, Language}
 import org.apache.spark.ml.linalg.Vectors
 
 import scala.reflect.runtime.universe._
@@ -97,13 +97,15 @@ trait RichMapFeature {
       blackListKeys: Array[String] = Array.empty,
       trackNulls: Boolean = TransmogrifierDefaults.TrackNulls,
       others: Array[FeatureLike[T]] = Array.empty,
-      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality
+      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality,
+      cleanTextParams: CleanTextParams = TransmogrifierDefaults.CleanParams
     ): FeatureLike[OPVector] = {
       new TextMapPivotVectorizer[T]()
         .setInput(f +: others)
         .setTopK(topK)
         .setCleanKeys(cleanKeys)
         .setCleanText(cleanText)
+        .setCleanTextParams(cleanTextParams)
         .setMinSupport(minSupport)
         .setWhiteListKeys(whiteListKeys)
         .setBlackListKeys(blackListKeys)
@@ -1026,7 +1028,8 @@ trait RichMapFeature {
       blackListKeys: Array[String] = Array.empty,
       trackNulls: Boolean = TransmogrifierDefaults.TrackNulls,
       others: Array[FeatureLike[EmailMap]] = Array.empty,
-      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality
+      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality,
+      cleanTextParams: CleanTextParams = CleanTextParams(true, false)
     ): FeatureLike[OPVector] = {
       val domains: Array[FeatureLike[PickListMap]] = (f +: others).map { e =>
         new EmailToPickListMapTransformer().setInput(e).getOutput()
@@ -1034,7 +1037,7 @@ trait RichMapFeature {
 
       domains.head.vectorize(
         topK = topK, minSupport = minSupport, cleanText = cleanText, cleanKeys = cleanKeys,
-        whiteListKeys = whiteListKeys, blackListKeys = blackListKeys,
+        cleanTextParams = cleanTextParams, whiteListKeys = whiteListKeys, blackListKeys = blackListKeys,
         others = domains.tail, trackNulls = trackNulls, maxPctCardinality = maxPctCardinality
       )
     }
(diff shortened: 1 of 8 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/feature/Transmogrifier.scala
```diff
@@ -519,24 +522,34 @@ trait TrackTextLenParam extends Params {
...
+        case (k: String, v: String) => cleanTextFn(k, shouldCleanKey, TransmogrifierDefaults.CleanParams) ->
+          cleanTextFn(v, shouldCleanValue, $(cleanTextParams))
         case (k: String, v: Traversable[_]) =>
           if (v.headOption.exists(_.isInstanceOf[String])) {
-            cleanTextFn(k, shouldCleanKey) -> v.asInstanceOf[Traversable[String]].map(cleanTextFn(_, shouldCleanValue))
+            cleanTextFn(k, shouldCleanKey, TransmogrifierDefaults.CleanParams) ->
+              v.asInstanceOf[Traversable[String]].map(cleanTextFn(_, shouldCleanValue, $(cleanTextParams)))
           } else {
-            cleanTextFn(k, shouldCleanKey) -> v
+            cleanTextFn(k, shouldCleanKey, TransmogrifierDefaults.CleanParams) -> v
```
tovbinm: replace ` TransmogrifierDefaults.CleanParams` with `shouldCleanValue`

core/src/main/scala/com/salesforce/op/stages/impl/feature/Transmogrifier.scala
```diff
@@ -519,24 +522,34 @@ trait TrackTextLenParam extends Params {
...
           }
-        case (k: String, v) => cleanTextFn(k, shouldCleanKey) -> v
+        case (k: String, v) => cleanTextFn(k, shouldCleanKey, TransmogrifierDefaults.CleanParams) -> v
```
tovbinm: same - replace ` TransmogrifierDefaults.CleanParams` with `shouldCleanValue`

core/src/main/scala/com/salesforce/op/stages/impl/feature/Transmogrifier.scala
```diff
@@ -608,8 +621,8 @@ trait MapPivotParams extends Params {
   protected def filterKeys[V](m: Map[String, V], shouldCleanKey: Boolean, shouldCleanValue: Boolean): Map[String, V] = {
     val map = cleanMap[V](m, shouldCleanKey, shouldCleanValue)
     val (whiteList, blackList) = (
-      $(whiteListKeys).map(cleanTextFn(_, shouldCleanKey)),
-      $(blackListKeys).map(cleanTextFn(_, shouldCleanKey))
+      $(whiteListKeys).map(cleanTextFn(_, shouldCleanKey, TransmogrifierDefaults.CleanParams)),
```
tovbinm: and same here - replace ` TransmogrifierDefaults.CleanParams` with `shouldCleanValue`

core/src/main/scala/com/salesforce/op/dsl/RichTextFeature.scala
```diff
@@ -600,11 +602,12 @@ trait RichTextFeature {
       minSupport: Int,
       trackNulls: Boolean = TransmogrifierDefaults.TrackNulls,
       others: Array[FeatureLike[Email]] = Array.empty,
-      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality
+      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality,
+      cleanTextParams: CleanTextParams = CleanTextParams(true, false)
```
tovbinm: name boolean arguments

core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
```diff
@@ -1026,15 +1028,16 @@ trait RichMapFeature {
       blackListKeys: Array[String] = Array.empty,
       trackNulls: Boolean = TransmogrifierDefaults.TrackNulls,
       others: Array[FeatureLike[EmailMap]] = Array.empty,
-      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality
+      maxPctCardinality: Double = OpOneHotVectorizer.MaxPctCardinality,
+      cleanTextParams: CleanTextParams = CleanTextParams(true, false)
```
gerashegalov: How are we going to evolve these params, keep adding booleans? should this be a list of enum values with corresponding text transformers?

utils/src/main/scala/com/salesforce/op/utils/text/TextUtils.scala
```diff
@@ -30,23 +30,47 @@
...
-    raw.map(t => cleanString(t, splitOn))
+  def cleanOptString(raw: Option[String], splitOn: String = " ",
+    cleanTextParams: CleanTextParams = defaultCleanParams): Option[String] =
+    raw.map(t => cleanString(t, splitOn, cleanTextParams))
 
-  def cleanString(raw: String, splitOn: String = " "): String = {
-    raw
-      .toLowerCase
+  def cleanString(raw: String, splitOn: String = " ", cleanTextParams: CleanTextParams = defaultCleanParams): String = {
+    val l = if (cleanTextParams.ignoreCase) raw.toLowerCase else raw
```
gerashegalov: looks like JDK has an interesting twist on ignoreCase and actually prefers to normalize to the upper case unless for the Georgian alphabet http://hg.openjdk.java.net/jdk7u/jdk7u6/jdk/file/8c2c5d63a17e/src/share/classes/java/lang/String.java#l1356
Consider more explicit flags: lowerCase, upperCase instead of more general ignoreCase

sanmitra:
@gerashegalov For now I am closing this PR since I am going to just turn off the email cleaning directly in AutoML and leave TMOG as it is. In future we can make changes to TMOG to provide more granular control on how the text is cleaned if there are many use-cases which require it.

PR #427, opened 2019-10-18
Property-based GLM test

michaelweilsalesforce:
**Related issues**
Adding more Tests for our ML capabilities (TDD)

Property based tests to check each of our the Regression Models (Linear Regression, Random Forest, GLM) will be picked based on the nature of the data (and response) generated

Doing the same tests on one single dataset instead


tovbinm:
@michaelweilsalesforce please update PR description. also, is it ready for review?

core/src/main/scala/com/salesforce/op/utils/stages/FitStagesUtil.scala
```diff
@@ -262,6 +261,7 @@ private[op] case object FitStagesUtil {
     val stages = stagesLayer.map(_._1)
     val (estimators, noFit) = stages.partition(_.isInstanceOf[Estimator[_]])
     val fitEstimators = estimators.map { case e: Estimator[_] =>
+      println(e.getInputFeatures().toSeq)
```
tovbinm: remove println

michaelweilsalesforce:
@Jauntbox These tests are actually painful for our runtime (~15-20 mins/ unit test)

tovbinm:
Builds will be killed if not output is produced for 10 minutes, so we can’t have an individual test take longer than that.

PR #428, opened 2019-10-18
Detecting names in text fields (deprecated)

MWYang:
`SmartTextVectorizer` now has an optional flag `detectSensitive` that will guess, using a combination of dictionary lookup and conditional logic, whether any of the input columns are names (which we don't want in our models in case of bias). For right now, just a warning is logged to console that there may be such names in the input fields. In the future, the `removeSensitive` flag will remove those columns from contributing to the output vector. Also in the future, the gender information that is extracted from name columns (using government data) will be used to check for model fairness.

A unary estimator `HumanNameIdentifier` is also included as a standalone drop-in for custom workflows.

**Additional context**
I completed this work as part of my ongoing Salesforce internship.


tovbinm:
@MWYang thanks for the contribution. I would appreciate to get some context about the proposed changes in the PR description.

core/src/main/scala/com/salesforce/op/dsl/RichTextFeature.scala
```diff
@@ -235,6 +235,8 @@ trait RichTextFeature {
       hashSpaceStrategy: HashSpaceStrategy = TransmogrifierDefaults.HashSpaceStrategy,
       defaultLanguage: Language = TextTokenizer.DefaultLanguage,
       hashAlgorithm: HashAlgorithm = TransmogrifierDefaults.HashAlgorithm,
+      detectSensitive: Boolean = false,
```
gerashegalov: may make sense to have an enum of treatment modes instead of adding many booleans. 
`SensitvieFeatureMode = { DETECT_ONLY, DETECT_AND_REMOVE}`

tovbinm:
@MWYang why do we need a new `FeatureType` for `Name`? what's so specific about that `Text` cannot handle?

MWYang:
> @MWYang why do we need a new `FeatureType` for `Name`? what's so specific about that `Text` cannot handle?

You're right, there's nothing special about `Name`. At the beginning of my ideation process, I had thought that it would make sense to have a different output type for the fields that get flagged as names, but it's no longer a part of the feature. Now, it seems the preferred way to store information about which fields are names is in the metadata, which is what I'm working towards.

Should I revert the changes for creating the new `Name` feature type then?

MWYang:
I'm closing because I made #440, which reduces the PR size by removing out a different feature that I didn't mean to commit into this branch and by removing the extraneous `Name` types, per @gerashegalov's suggestion. Would appreciate a review of the new PR so I can know what to work on!

tovbinm:
Thank you @MWYang I will have a look.

PR #429, opened 2019-10-18
Eval dataset should not be balanced

AdamChit:
**Related issues**
validationPrepare was being called on the validation datasets which will balance the dataset. This should not be done because the validation set should represent the real distribution. 

remove the call to validationPrepare


tovbinm:
@leahmcguire @AdamChit didn't we just changed it in this PR - https://github.com/salesforce/TransmogrifAI/pull/424 ?

leahmcguire:
Yes @tovbinm I was too quick to make the change :-)

PR #434, opened 2019-11-08
data cutter logging

AdamChit:
**Related issues**

**Additional context**


core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -125,9 +125,9 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
     val dataPrep = data
       .filter(r => labelSet.contains(r.getDouble(labelColIdx)))
       .withColumn(labelColName, data(labelColName).as(labelColName, metadataNA.toMetadata))
-
+    log.info(s"number of rows after only using top 100 labels ${dataPrep.count()}")
     // calculate the down sample fraction
-    val dataSetSize = data.count()
+    val dataSetSize = dataPrep.count()
```
tovbinm: so this was a bug then?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -125,9 +125,9 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
     val dataPrep = data
       .filter(r => labelSet.contains(r.getDouble(labelColIdx)))
       .withColumn(labelColName, data(labelColName).as(labelColName, metadataNA.toMetadata))
-
+    log.info(s"number of rows after only using top 100 labels ${dataPrep.count()}")
```
tovbinm: Please use upper case in the first word, e.g. `Number of rows...` (same applies for other log statements.)

core/src/main/scala/com/salesforce/op/stages/impl/tuning/OpCrossValidation.scala
```diff
@@ -136,6 +139,7 @@ private[op] class OpCrossValidation[M <: Model[_], E <: Estimator[_]]
     // Await for all the evaluations to complete
     val modelSummaries = SparkThreadUtils.utils.awaitResult(Future.sequence(modelSummariesFuts.toSeq), maxWait)
 
+    log.info(s"finding best model")
```
tovbinm: is this important?

core/src/main/scala/com/salesforce/op/stages/impl/tuning/DataCutter.scala
```diff
@@ -151,13 +151,15 @@ class DataCutter(uid: String = UID[DataCutter]) extends Splitter(uid = uid) with
   override def validationPrepare(data: Dataset[Row]): Dataset[Row] = {
 
     val dataPrep = super.validationPrepare(data)
+    log.info(s"count of dataset before down-sampling for this fold ${dataPrep.count()}")
```
tovbinm: 1. counts in spark are not free. do you really need it `info`? perhaps in debug only?
2. also use `log.info("count blah blah {}", data.count())` that way if logging is disabled there is no cost for the count

gerashegalov:
I assume we don't need this anymore, feel free to reopen @AdamChit if needed

PR #437, opened 2019-11-14
Remove names from SmartTextVectorizer; Add metadata features for sensitive fields

MWYang:
**Related issues**
This PR completes the work that #440 started (itself a rework of #428). First merge #440 before this PR.

The optional flag `DetectAndRemove` in `SmartTextVectorizer` now works properly.

A new `SensitiveFeatureInformation` is now an optional attribute on `FeatureInsight` objects. Individual estimators, like `SmartTextVectorizer` (implemented in this PR), can write sensitive feature information to the metadata, which will then be extracted into `FeatureInsight` objects by `ModelInsights`.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/ModelInsights.scala
+++ b/core/src/main/scala/com/salesforce/op/ModelInsights.scala
@@ -333,7 +333,7 @@ case class Discrete(domain: Seq[String], prob: Seq[Double]) extends LabelInfo
  * @param metrics          sequence containing metrics computed in RawFeatureFilter
  * @param distributions    distribution information for the raw feature (if calculated in RawFeatureFilter)
  * @param exclusionReasons exclusion reasons for the raw feature (if calculated in RawFeatureFilter)
- *
+ * @param sensitiveInformation derived information about sensitive field checks (if performed)
  */
 case class FeatureInsights
 (
@@ -342,7 +342,8 @@ case class FeatureInsights
   derivedFeatures: Seq[Insights],
   metrics: Seq[RawFeatureFilterMetrics] = Seq.empty,
   distributions: Seq[FeatureDistribution] = Seq.empty,
-  exclusionReasons: Seq[ExclusionReasons] = Seq.empty
+  exclusionReasons: Seq[ExclusionReasons] = Seq.empty,
+  sensitiveInformation: Option[SensitiveFeatureInformation] = None
 )
 
 /**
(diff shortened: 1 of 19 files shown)


core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -697,8 +698,34 @@ case object ModelInsights {
...
-          FeatureInsights(featureName = fname, featureType = ftype, derivedFeatures = seq.map(_._2),
-            metrics = metrics, distributions = distributions, exclusionReasons = exclusionReasons)
+          val sensitiveFeatureInformation = vectorInfo.flatMap(_.sensitive.get(fname))
+          FeatureInsights(
+            featureName = fname, featureType = ftype, derivedFeatures = seq.map(_._2),
+            metrics = metrics, distributions = distributions, exclusionReasons = exclusionReasons,
+            sensitiveInformation = sensitiveFeatureInformation
+          )
+      }.toSeq ++ {
+        // Add FeatureInsights for removed sensitive fields that do not have a column in OpVectorMetadata
```
leahmcguire: If there is no column in OpVectorMetadata all the metadata will be screwed up since this is an indexed list for the vector

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -77,6 +93,120 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
 
+  /* CODE FOR DETECTING SENSITIVE FEATURES BEGIN */
+  var guardCheckResults: Option[Array[Boolean]] = None
+
+  override def fit(dataset: Dataset[_]): SequenceModel[T, OPVector] = {
+    /* Set instance variable for guardCheck results here.
+    We compute guardChecks here because all of the logical checks can be implemented efficiently in native Spark,
+    and we would like to use the individual Text columns before they are combined in SequenceEstimator.fit().
+    It is not directly possible to do this computation in the treeAggregate() call below in fitFn() because one of the
+    logical checks computes a standard deviation, which requires knowing the mean beforehand. We could implement the
```
leahmcguire: Standard deviation can be approximated in a single pass - https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -77,6 +93,120 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
+    }
+
+    def aggregateSeqResults(
+      one: Seq[NameIdentificationAccumulator], two: Seq[NameIdentificationAccumulator]
+    ): Seq[NameIdentificationAccumulator] = {
+      one zip two map { case (x, y) => aggregateTwoResults(x, y) }
+    }
+
+    import com.salesforce.op.features.types.NameStats.GenderStrings._
+    def computeResults(input: T#Value): NameIdentificationAccumulator = {
```
leahmcguire: lets set up some time to talk about doing this in the fitFn with monoid aggregators

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -96,15 +226,19 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
+    val nameIdentificationResults = if (getRemoveSensitive) Some(detectSensitive(dataset)) else None
+
     val smartTextParams = SmartTextVectorizerModelArgs(
       isCategorical = isCategorical,
       topValues = topValues,
       shouldCleanText = shouldCleanText,
       shouldTrackNulls = $(trackNulls),
-      hashingParams = makeHashingParams()
+      hashingParams = makeHashingParams(),
+      isName = nameIdentificationResults.map(_.isName).getOrElse(Array.empty[Boolean]),
```
leahmcguire: is name the only type of sensitive information we ever want to hide? perhaps this should be an enum so we leave ourselves room to expand in the future

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -151,7 +297,48 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
+              |Percentage Likely Male Names: ${results.pctMale(index)}
+              |Percentage Likely Female Names: ${results.pctFemale(index)}
+              |Percentage Where No Gender Found: ${results.pctOther(index)}
+              |""".stripMargin
+            }
+          }
+        }
+
+        // Transform into the tuple from feature name to SensitiveFeatureInformation
+        inN.zipWithIndex map { case (feature: TransientFeature, index: Int) =>
```
leahmcguire: does it need to be created for every feature passed in? why not just a map from the feature names found to be sensitive to the information? very large metadata can have a cost

core/src/main/scala/com/salesforce/op/utils/stages/NameIdentificationUtils.scala
```diff
@@ -0,0 +1,165 @@
...
+ * Provides shared helper functions and variables (namely, broadcast dictionaries) for name identification
+ * and name to gender transformation.
+ * @tparam T     the FeatureType (subtype of Text) to operate over
+ */
+private[op] trait NameIdentificationFun[T <: Text] extends Logging {
+  import com.salesforce.op.utils.stages.NameIdentificationUtils._
+
+  def preProcess(s: T#Value): Seq[String] = {
+    s.getOrElse("").toLowerCase().split("\\s+").map((token: String) =>
+      token.replace("\\P{L}", "")
```
leahmcguire: we have standard text cleaning utils - lets use them please

MWYang:
Closing to rework on comments from reviewers.

PR #440, opened 2019-11-21
Detecting names in text fields

MWYang:
`SmartTextVectorizer` now has an optional flag `detectSensitive` that will guess, using a combination of dictionary lookup and conditional logic, whether any of the input columns are names (which we don't want in our models in case of bias). For right now, just a warning is logged to console that there may be such names in the input fields. In the future, the `removeSensitive` flag will remove those columns from contributing to the output vector. Also in the future, the gender information that is extracted from name columns (using government data) will be used to check for model fairness.

A unary estimator `HumanNameIdentifier` is also included as a standalone drop-in for custom workflows.

**Additional context**
I completed this work as part of my ongoing Salesforce internship. This PR cleans up and replaces #428.


MWYang:
Closing to rework on comments from reviewers.

PR #441, opened 2019-11-22
make helloworld a subproject

nicodv:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/87

helloworld is now a subproject

N/A

**Additional context**
See discussion in https://github.com/salesforce/TransmogrifAI/pull/254

I've tested the build and the helloworld examples, which still run fine.


nicodv:
Played around with Gradle and made helloworld a subproject. This DRYs up the code, but not sure if it's desired. Let me know if it's useful.

tovbinm:
@nicodv thanks for the effort. The idea was to have a Helloworld to build against the latest release and not the master branch. I would rather extract it out as a separate project. In fact we already did so with for sbt - https://github.com/salesforce/transmogrifai-helloworld-sbt

nicodv:
I see, @tovbinm . I interpreted `helloworld` as example code against the current API (i.e. master), and the `templates` directory the thing you would want to separate out for others to base their own work on and adapt.

gerashegalov:
Thanks @nicodv, your RP is what I was thinking in terms of #254. But it's indeed better to move all the examples to transmogrifai-helloworld-sbt as @tovbinm points out to increase DRYness even more :). And we can see if we can set up a build chain such that PR check includes the build of transmogrifai-helloworld-sbt on the TransmogrifAI PR branch as we do internally.

PR #442, opened 2019-11-25
Added Chinese and Korean examples to TextTokenizerTest

Jauntbox:
**Related issues**
n/a

n/a

n/a

**Additional context**
This is a small change to better allow testing of alternatives to the CJK tokenizer (that we've already replaced for Japanese). The CJK tokenizer uses bigrams for its tokenization, rather than trying to extract words, so most of the tokens from a text sample will have length 2 (not all, since other languages can be mixed in). Some of the simpler ID detection calculations will look at the distributions of token lengths, so they may incorrectly think that text from languages using the CJK tokenizer is IDs.


tovbinm:
LOCO test is failing @sanmitra @Jauntbox

sanmitra:
@tovbinm The LOCO test - `com.salesforce.op.stages.impl.insights.RecordInsightsLOCOTest` is succeeding. Where exactly you are seeing the failure of LOCO test ?

tovbinm:
It’s a flaky one. See previous runs.

PR #445, opened 2019-12-09
Unary estimator for detecting names and transforming to gender

MWYang:
The PR implements the following algorithm for detecting names in text fields and inferring gender as a standalone unary estimator:
- First, a majority (or some custom threshold) of entries must be found in a large name dictionary
- If the threshold is reached, the transformer uses several heuristics (the first or last token; RegEx patterns; presence of honorifics like Mr. or Ms.) to find the likely first name and infer the gender.

**Additional context**
I completed this work as part of my ongoing Salesforce internship. This PR cleans up and replaces many previous ones: #440, #437, #428. Per internal review with @leahmcguire, I'm including my name identification and name-to-gender code in two PRs. This first one implements the standalone unary estimator where most of the actual logic is in a utils file, `NameDetectUtils.scala`. The code also makes use of Algebird monoids in order to run in a single pass over the data. Pending approval on this PR, I will then make a separate PR with the proper (and hopefully fewer) changes made in SmartTextVectorizer.


tovbinm:
I am still not convinced why do we need to add a brand new time, instead of just returning a Map type. And if we do decided to add a new type I would better add it as a separate PR, since it requires a bit more work that was done here.

leahmcguire:
I think this is a very appropriate use of a new type...

MWYang:
> I am still not convinced why do we need to add a brand new feature type, instead of just returning a Map type. And if we do decided to add a new type I would better add it as a separate PR, since it requires a bit more work that was done here.

I think the main thing is that I want to define the keys/values for the Map return object in a standard way, and so creating the custom type is a way of signaling that one should expect those (and only those) keys/values. I don't have to add an extra TMOG type to do that though. Maybe there's also a better way for me to solve this problem too.

core/src/test/scala/com/salesforce/op/stages/impl/feature/HumanNameDetectorTest.scala
```diff
@@ -0,0 +1,287 @@
...
+
+  it should "ignore null values in calculating stats" in {
+    val names = RandomText.names.withProbabilityOfEmpty(0.90).take(200).toList
+    val (newData, newFeature) = TestFeatureBuilder(names)
+    val model = estimator.setInput(newFeature).fit(newData)
+    model.asInstanceOf[HumanNameDetectorModel[Text]].treatAsName shouldBe true
+
+    val countNullsModel = estimator.setIgnoreNulls(false).setInput(newFeature).fit(newData)
+    countNullsModel.asInstanceOf[HumanNameDetectorModel[Text]].treatAsName shouldBe false
+  }
```
leahmcguire: can you add a test for mixed data and the threshold for name detection

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -291,6 +291,46 @@ object GeolocationMap {
...
+}
+object NameStats {
+  object Keys {
+    val OriginalName = "original"
+    val IsNameIndicator = "isName"
+    val FirstName = "firstName"
+    val LastName = "lastName"
+    val Gender = "gender"
+  }
+  object BooleanStrings {
```
leahmcguire: what are the boolean strings for?

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -291,6 +291,46 @@ object GeolocationMap {
...
+    val Male = "male"
+    val Female = "female"
+    val GenderNA = "NA"
+    val GenderNotInferred = "NotInferred"
+  }
+
+  def apply(value: Map[String, String]): NameStats = new NameStats(value)
+  def empty: NameStats = FeatureTypeDefaults.NameStats
+}
+
```
leahmcguire: please add your feature type to the feature type tests here: https://github.com/salesforce/TransmogrifAI/tree/master/features/src/test/scala/com/salesforce/op/features/types

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,143 @@
...
+  operationName = operationName
+) with NameDetectFun[T] {
+
+  def fitFn(dataset: Dataset[T#Value]): HumanNameDetectorModel[T] = {
+    val spark = dataset.sparkSession
+    // Load name and gender data into broadcast variables
+    val broadcastNameDict: Broadcast[NameDictionary] = spark.sparkContext.broadcast(NameDictionary())
+    val broadcastGenderDict: Broadcast[GenderDictionary] = spark.sparkContext.broadcast(GenderDictionary())
+    // Create HyperLogLog factory
+    val hllMonoid = new HyperLogLogMonoid(NameDetectUtils.HLLBits)
```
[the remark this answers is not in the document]
MWYang: This is my bad with the comment—the `HyperLogLogMonoid` object is not a factory, but it is something that I need to reuse. I'll update the comment.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,377 @@
...
+      tokens.map({ token: String => if (dict.value.value contains token) 1 else 0}).sum.toDouble / tokens.length
+    }
+  }
+
+  def getNameFromCustomIndex(tokens: Seq[String], index: Int): String = {
+    if (tokens.isEmpty) ""
+    else if (tokens.length == 1) tokens.head
+    else {
+      // Mod to accept -1 as valid index
+      tokens((index + tokens.length) % tokens.length)
```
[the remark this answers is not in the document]
MWYang: This might just be being too Pythonic in Scala-land but I wanted functionality for getting the last element in a list. I could make this more explicit in the `GenderDetectStrategy` enum by making `GetFirst` and `GetLast` options but I thought it would be more useful to just specify it in a `ByIndex` class so that somebody could specify arbitrary indices to check.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,377 @@
...
+    val entryName: String = parts(0)
+    entryName match {
+      case "ByIndex" => ByIndex(parts(1).toInt)
+      case "ByRegex" => ByRegex(parts(1).r)
+      case "FindHonorific" => FindHonorific()
+    }
+  }
+}
+
+private[op] trait NameDetectParams extends Params {
```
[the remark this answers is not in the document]
MWYang: I'm not sure what you're suggesting, can you clarify?

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
MWYang: So the `NameStats` type is a `Map[String, String]` (b/c for other values, namely the gender, I need multiple possible values. But then for another key  in `NameStats` the value I want is basically a boolean (see `transformFn` in the companion transformer in `HumanNameDetector`).

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,377 @@
...
+
+import scala.io.Source
+import scala.util.Try
+import scala.util.matching.Regex
+
+/**
+ * Provides shared helper functions and variables for name identification and name to gender transformation.
+ * @tparam T     the FeatureType (subtype of Text) to operate over
+ */
+private[op] trait NameDetectFun[T <: Text] extends NameDetectParams with Logging {
```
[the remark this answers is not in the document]
MWYang: I would prefer that as well but I need to pass the type parameter `T` as a subtype of `Text` if I'm to use the most semantically accurate type annotation `T#Value` in a few places. I could replace all the instances of `T#Value` with `Option[String]` and then change the trait to an object if you think that would be better?

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,377 @@
...
+          AveragedValue(1L, dictCheck(tokens, nameDict)),
+          computeGenderResultsByStrategy(text, tokens, genderDict)
+        )
+      case None if $(ignoreNulls) => NameDetectStats.empty
+      case _ =>
+        val tokens = Seq.empty[String]
+        NameDetectStats(
+          computeGuardCheckQuantities("", tokens, hll),
+          AveragedValue(1L, dictCheck(tokens, nameDict)),
+          computeGenderResultsByStrategy("", tokens, genderDict)
```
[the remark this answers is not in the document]
MWYang: If not ignoring nulls, we need to create the `NameDetectStats` accumulator with a count of 1. If ignoring nulls, then we use `NameDetectStats.empty`, which has counts of 0.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,377 @@
...
+ *
+ * Name and gender data are maintained by and taken from this repository:
+ *  https://github.com/MWYang/InternationalNames
+ * which itself sources data from:
+ *  https://ec.europa.eu/jrc/en/language-technologies/jrc-names (currently unused)
+ *  https://github.com/OpenGenderTracking/globalnamedata
+ *  https://github.com/first20hours/google-10000-english
+ */
+private[op] object NameDetectUtils {
+  case class NameDictionary
```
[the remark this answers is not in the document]
MWYang: I made case classes in order to provide a more informative type name to the Spark broadcast variables that I was passing around. I think I can address your comment just by instantiating the case classes I defined inside the `NameDetectUtils` object, something like `val DefaultNameDictionary = ...`. I can also remove the case class entirely if you would prefer that.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,377 @@
...
+  )
+
+  /**
+   * Number of bits used for hashing in HyperLogLog (HLL). Error is about 1.04/sqrt(2^{bits}).
+   * Default is 12 bits for 1% error which means each HLL instance is about 2^{12} = 4kb per instance.
+   */
+  val HLLBits = 12
+
+  val MaleHonorifics: Set[String] = Set("mr", "mister", "sir")
+  val FemaleHonorifics: Set[String] = Set("ms", "mrs", "miss", "madam")
```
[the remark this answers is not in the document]
MWYang: As an MVP, no. 😄 This should be a separate discussion but multilingual name/gender detection is a whole other ballgame (in terms of data availability and implementation). When that time comes, I agree the code might have to be refactored significantly to do a top-level language/country of origin detection that then changes all of the rest of the logic.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,377 @@
...
+        ),
+        l.guardCheckQuantities.approxNumUnique + r.guardCheckQuantities.approxNumUnique
+      ),
+      l.dictCheckResult + r.dictCheckResult,
+      l.genderResultsByStrategy + r.genderResultsByStrategy
+    )
+    override def zero: NameDetectStats = NameDetectStats.empty
+  }
+
+  def empty: NameDetectStats = NameDetectStats(GuardCheckStats(), AveragedGroup.zero, Map.empty[String, GenderStats])
```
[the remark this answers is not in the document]
MWYang: I thought this would be easy to do but then quickly realized that the whole reason I have to implement a custom `toJson` is because Algebird doesn't serialize nicely yet (https://github.com/twitter/algebird/issues/119). So, I don't have an easy way of deserializing HLL objects, and I also won't be using a `fromJson` method for this PR (the `toJSON` is just for printing/debugging, and I've made that more clear in the comment now).

MWYang:
@leahmcguire I've implemented the straightforward changes that you requested. When you have a moment, could you go through the remaining conversations and either follow-up or resolve?

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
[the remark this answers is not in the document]
tovbinm: I think there is a way to serialize algebird types with Kryo. It should be somewhere in the code already.

tovbinm:
Adding an extra feature type is somewhat involved (for example see the internal tmog repo when we added a Prediction type). Thats why if you folks decide to do it then I think it should be done in a separate PR.

tovbinm:
@MWYang found it! HLL serialized with kryo - https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/stages/impl/feature/OpOneHotVectorizer.scala#L370

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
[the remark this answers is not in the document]
leahmcguire: can you make another method for this creation type on NameDetectStats?

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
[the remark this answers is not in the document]
leahmcguire: yeah I would go ahead and just remove the case classes- just have the vals

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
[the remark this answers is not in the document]
leahmcguire: ok can you make a not in the description that this only works for english?

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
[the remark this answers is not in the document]
leahmcguire: consistency across the code base

core/src/test/scala/com/salesforce/op/stages/impl/feature/TempTest.scala
```diff
@@ -0,0 +1,78 @@
...
+import com.salesforce.op.testkit.{DataSources, RandomText}
+import com.salesforce.op.utils.stages.GenderDetectStrategy
+import com.salesforce.op.utils.stages.GenderDetectStrategy.FindHonorific
+import org.apache.spark.sql.DataFrame
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class TempTest
```
leahmcguire: ??

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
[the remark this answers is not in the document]
MWYang: I made a different change that removed this ugliness :)

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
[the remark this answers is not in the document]
MWYang: Oh I see. These params will get re-used across the unary estimator, SmartTextVectorizer, and SmartTextMapVectorizer though.

core/src/main/scala/com/salesforce/op/dsl/RichTextFeature.scala
```diff
@@ -434,6 +434,15 @@ trait RichTextFeature {
       toLowercase: Boolean = TextTokenizer.ToLowercase
     ): FeatureLike[Binary] =
       f.transformWith(new SubstringTransformer[T, T2]().setToLowercase(toLowercase), f2)
+
+    /**
+     * Check if feature is actual human names, and if so, return related demographic information
```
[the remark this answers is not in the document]
MWYang: Yes, I can update the comment to make this more clear.

core/src/test/scala/com/salesforce/op/stages/impl/feature/TempTest.scala
MWYang: My bad, I accidentally committed this.

MWYang:
> @MWYang found it! HLL serialized with kryo - https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/stages/impl/feature/OpOneHotVectorizer.scala#L370

Hmm, I couldn't get this to work. I think my goal here is different than what this code is trying to do. I'm just trying to get the stats object to print correctly when the `JsonLike` trait doesn't seem to work with Algebird types.

I did figure out how to use custom serializers with `JsonUtils` though, which is what I've done now: https://github.com/MWYang/TransmogrifAI/blob/464fe523cc9539cb61c75acb3e86d5537bc3fd21/core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala#L312

tovbinm:
@MWYang got it! In this case I would remove `JsonLike` base trait for `GuardCheckStats` and `GenderStats`, to avoid someone calling it by mistake.

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,172 @@
...
+(
+  override val uid: String,
+  val treatAsName: Boolean,
+  val orderedGenderDetectStrategies: Seq[GenderDetectStrategy] = Seq.empty[GenderDetectStrategy]
+)(implicit tti: TypeTag[T])
+  extends UnaryModel[T, NameStats]("human name detector", uid) with NameDetectFun[T] {
+
+  var broadcastGenderDict: Option[Broadcast[GenderDictionary]] = None
+
+  override def transform(dataset: Dataset[_]): DataFrame = {
```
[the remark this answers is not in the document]
MWYang: You're right, this did not work. Thanks for catching it. Do you have any suggestions on how to fix it? The basic problem is that I need to have the Spark dataset in scope to access `sparkSession` in order to make the broadcast variable for the gender dictionary. Alternatively, I could just not load the gender dictionary into a broadcast variable.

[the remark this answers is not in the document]
MWYang: I'm just going to go with the second (not using broadcast variable for the transformer part). It doesn't really matter for right now.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  private[op] def computeGuardCheckQuantities(
+    text: String,
+    tokens: Seq[String],
+    hllMonoid: HyperLogLogMonoid
+  ): GuardCheckStats = {
+    val textLength = text.length
+    GuardCheckStats(
+      countBelowMaxNumTokens = if (tokens.length < 10) 1 else 0,
```
leahmcguire: why 10? can you add a comment?

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,162 @@
...
+  val GenderResultsByStrategyKey = "genderResultsByStrategy"
+
+  def fromMetadata(metadata: Metadata): HumanNameDetectorMetadata = {
+    val genderResultsMetadata = metadata.getMetadata(GenderResultsByStrategyKey)
+    val genderResultsByStrategy: Map[String, GenderStats] = {
+      NameDetectUtils.GenderDetectStrategies map { strategy: GenderDetectStrategy =>
+        val strategyString = strategy.toString
+        val resultsArray = genderResultsMetadata.getDoubleArray(strategyString)
+        require(resultsArray.length == 3, "There must be exactly three values for each gender detection strategy.")
+        strategyString -> GenderStats(resultsArray(0).toInt, resultsArray(1).toInt, resultsArray(2).toInt)
```
tovbinm: please name arguments to `GenderStats`

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -55,7 +55,7 @@ import scala.reflect.runtime.universe.TypeTag
  * plus occurrences of non top k values and a null indicator (if enabled).
  * Non-categoricals will be converted into a vector using the hashing trick. In addition, a null indicator is created
  * for each non-categorical (if enabled).
- *
+ *r
```
tovbinm: remove `r`

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+import scala.util.Try
+
+/**
+ * Provides shared helper functions and variables for name identification and name to gender transformation.
+ */
+private[op] trait NameDetectFun extends Logging with NameDetectParams {
+  import GenderDetectStrategy._
+  import NameDetectUtils._
+
+  private[op] implicit def textValue2String[T <: Text](input: T#Value): String = input.getOrElse("")
```
tovbinm: do these function really provide that much value - `textValue2String` & `string2Text`? if you would like to check if `Text` value is empty do:
```scala
import com.salesforce.op.feature.types._
Text("abc") match {
   case SomeValue(Some(s)) => // we got some string value
   case _ => // handle empty case
}
```
or if you would like to check the string value length as well do:
```scala
Text("abc") match {
   case SomeValue(Some(s)) if s.length > 0 => // we got some non empty string value
   case _ => // handle empty case 
}
```

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,162 @@
...
+ * @param uid           uid for instance
+ * @param operationName unique name of the operation this stage performs
+ * @param tti           type tag for input
+ * @param ttiv          type tag for input value
+ * @tparam T            the FeatureType (subtype of Text) to operate over
+ */
+class HumanNameDetector[T <: Text]
+(
+  uid: String = UID[HumanNameDetector[T]],
+  operationName: String = "human name detector"
```
tovbinm: please make sure operation is short and does not contain spaces, e.g. `humanNameDetect`

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+  private[op] implicit def string2Text[T <: Text](input: String): T#Value = input match {
+    case "" => Text.empty.value
+    case _ => Text(input).value
+  }
+
+  private[op] def preProcess(input: String): Seq[String] = {
+    TextTokenizer.tokenize(Text(input)).tokens.toArray
```
tovbinm: `toArray` seems redundant

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+      N < 10 || stats.approxMomentsOfTextLength.stddev > 0.05,
+      // check that the number of unique entries is at least 10
+      N < 100 || hllMonoid.sizeOf(stats.approxNumUnique).estimate > 10
+    )
+    checks.forall(identity)
+  }
+
+  private[op] def dictCheck(tokens: Seq[String], dict: NameDictionary): Double = {
+    if (tokens.isEmpty) 0.0 else {
+      tokens.map({ token: String => if (dict contains token) 1 else 0}).sum.toDouble / tokens.length
```
tovbinm: please be explicit on value types `1.0` and `0.0` to avoid conversions

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  private[op] def computeGenderResultsByStrategy(
+    text: String,
+    tokens: Seq[String],
+    genderDict: GenderDictionary
+  ): Map[String, GenderStats] = {
+    GenderDetectStrategies map { strategy: GenderDetectStrategy =>
+      val genderResult: String = identifyGender(text, tokens, strategy, genderDict)
+      strategy.toString -> GenderStats(
```
tovbinm: here this ` implicit def booleanToInt(v: Boolean): Int = if (v) 1 else 0` can be justified

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+    } toMap
+  }
+
+  private[op] def computeResults(
+    input: String,
+    nameDict: Broadcast[NameDictionary],
+    genderDict: Broadcast[GenderDictionary],
+    hll: HyperLogLogMonoid
+  ): NameDetectStats = input match {
+    case "" if $(ignoreNulls) => NameDetectStats.empty
```
tovbinm: I think it is a bad practice to handle String values. Better operate on our own `Text` or `Option[String]` and check for emptiness as I pointed our above.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+      NameDetectStats(
+        computeGuardCheckQuantities(input, tokens, hll),
+        AveragedValue(1L, dictCheck(tokens, nameDict.value)),
+        computeGenderResultsByStrategy(input, tokens, genderDict.value)
+      )
+  }
+
+  type NameDetectMapFun[T <: Text] = T#Value => NameDetectStats
+  private[op] def makeMapFunction[T <: Text](spark: SparkSession): NameDetectMapFun[T] = {
+    val broadcastNameDict: Broadcast[NameDictionary] = spark.sparkContext.broadcast(DefaultNameDictionary)
```
tovbinm: what are the performance implications of the broadcast variable here? why not use `object` instead?

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,162 @@
...
+  override val ttiv: TypeTag[T#Value]
+) extends UnaryEstimator[T, NameStats](
+  uid = uid,
+  operationName = operationName
+) with NameDetectFun {
+
+  def fitFn(dataset: Dataset[T#Value]): HumanNameDetectorModel[T] = {
+    implicit val (nameDetectStatsEnc, nameDetectStatsMonoid) = (NameDetectStats.kryo, NameDetectStats.monoid)
+    val mapFun: T#Value => NameDetectStats = makeMapFunction(dataset.sparkSession)
+    val aggResults: NameDetectStats = dataset.map(mapFun).reduce(_ + _)
```
tovbinm: fyi, if the dataset is empty the reduce will yield  an error.

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,162 @@
...
+    val tokens = preProcess(input.value)
+    if (treatAsName) {
+      require(orderedGenderDetectStrategies.nonEmpty, "There must be a gender extraction strategy if treating as name.")
+      // TODO: Figure out how to use a broadcast variable for the gender dictionary within a unary transformer
+      val genders: Seq[String] = orderedGenderDetectStrategies map {
+        identifyGender(input.value.getOrElse(""), tokens, _, NameDetectUtils.DefaultGenderDictionary)
+      }
+      val gender = genders.find(_ != GenderNA).getOrElse(GenderNA)
+      NameStats(Map(
+        IsNameIndicator -> True,
```
tovbinm: `IsNameIndicator -> treatAsName`

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,162 @@
...
+  val PredictedNameProbKey = "predictedNameProb"
+  val GenderResultsByStrategyKey = "genderResultsByStrategy"
+
+  def fromMetadata(metadata: Metadata): HumanNameDetectorMetadata = {
+    val genderResultsMetadata = metadata.getMetadata(GenderResultsByStrategyKey)
+    val genderResultsByStrategy: Map[String, GenderStats] = {
+      NameDetectUtils.GenderDetectStrategies map { strategy: GenderDetectStrategy =>
+        val strategyString = strategy.toString
+        val resultsArray = genderResultsMetadata.getDoubleArray(strategyString)
+        require(resultsArray.length == 3, "There must be exactly three values for each gender detection strategy.")
```
tovbinm: minor: name the values - `s"There must be exactly three values for each gender detection strategy: $TreatAsNameKey, $PredictedNameProbKey, $GenderResultsByStrategyKey"`

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+  ): String = {
+    strategy match {
+      case FindHonorific() =>
+        val matched = tokens filter { NameDetectUtils.AllHonorifics contains }
+        if (matched.length == 1) {
+          if (MaleHonorifics contains matched.head) Male else Female
+        }
+        else GenderNA
+      case ByIndex(index) =>
+        val nameToCheckGenderOf = tokens.lift(index).getOrElse("")
```
tovbinm: again, I think it's a bad idea to operate on `""`, instead either use `Option[String]` or `Text.empty`

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+
+  private[op] def identifyGender(
+    text: String,
+    tokens: Seq[String],
+    strategy: GenderDetectStrategy,
+    genderDict: GenderDictionary
+  ): String = {
+    strategy match {
+      case FindHonorific() =>
+        val matched = tokens filter { NameDetectUtils.AllHonorifics contains }
```
tovbinm: Ho about a bit direct way:
```scala
(tokens collect {
    case v if MaleHonorifics.contains(v) => Male
    case v if FemaleHonorifics.contains(v) => Female
}) match {
    case h :: Nil => h
    case _ => GenderNA     
}
```

features/src/main/scala/com/salesforce/op/stages/impl/feature/GenderDetectStrategy.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+  case class ByRegex(pattern: Regex) extends GenderDetectStrategy {
+    override def toString: String = "ByRegex" + delimiter + pattern.toString
+  }
+  case class FindHonorific() extends GenderDetectStrategy {
+    override def toString: String = "FindHonorific"
+  }
+
+  def fromString(s: String): GenderDetectStrategy = {
+    val parts = s.split(delimiter)
```
tovbinm: better and safer:
```scala
Option(s).map(_.split(delimiter)) match {
    case Some(Array("ByIndex", index)) if Try(index.toInt).isSuccess => ByIndex(index.toInt)
    // etc.
   case None => sys.error("Attempted to deserialize GenderDetectStrategy but no matching entry found for empty value")
   case _ => sys.error(s"Attempted to deserialize GenderDetectStrategy but no matching entry found for value '$s'")
}
```

features/src/main/scala/com/salesforce/op/features/types/FeatureTypeFactory.scala
```diff
@@ -110,6 +110,8 @@ case object FeatureTypeFactory {
         if (value == null) FeatureTypeDefaults.IDMap else new IDMap(value.asInstanceOf[Map[String, String]])
       case t if t =:= weakTypeOf[IntegralMap] => (value: Any) =>
         if (value == null) FeatureTypeDefaults.IntegralMap else new IntegralMap(value.asInstanceOf[Map[String, Long]])
+      case t if t =:= weakTypeOf[NameStats] => (value: Any) =>
```
tovbinm: are these all the places you need to add the `NameStats` type? how about `FeatureBuilder` and `TestFeatureBuilder`?

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+ * which itself sources data from:
+ *  https://ec.europa.eu/jrc/en/language-technologies/jrc-names
+ *  https://github.com/OpenGenderTracking/globalnamedata
+ *  https://github.com/first20hours/google-10000-english
+ */
+private[op] object NameDetectUtils {
+  import GenderDetectStrategy._
+
+  type NameDictionary = Set[String]
+  val DefaultNameDictionary: NameDictionary = {
```
tovbinm: make this a `lazy val DefaultNameDictionary`

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+      nameDictionary += name
+    }
+    buffer.close
+    nameDictionary.toSet[String]
+  }
+  // Use the following line to use the smaller but less noisy gender dictionary as a source for names
+  // val DefaultNameDictionary: NameDictionary = DefaultGenderDictionary.value.keySet
+
+  type GenderDictionary = Map[String, Double]
+  val DefaultGenderDictionary: GenderDictionary = {
```
tovbinm: same here `lazy val DefaultGenderDictionary`  - to avoid loading this every time

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
```diff
@@ -0,0 +1,337 @@
...
+
+  val MaleHonorifics: Set[String] = Set("mr", "mister", "sir")
+  val FemaleHonorifics: Set[String] = Set("ms", "mrs", "miss", "madam")
+  val AllHonorifics: Set[String] = MaleHonorifics ++ FemaleHonorifics
+
+  /**
+   * The strategies to use for transforming name to gender; Order does not matter.
+   *
+   * The first RegEx pattern will extract all text after the first comma;
+   * The second RegEx pattern will extract all text after both the first comma and the immediately next token,
```
tovbinm: why not name the regex based on you comments? e.g.
```scala
val TextAfterFirstComma = """.*,(.*)""".r
val CommaAndNextToken = """.*,\s+.*?\s+(.*)""".r
```

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
MWYang: `TextTokenizer.tokenize().tokens` is a `TextList`

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
MWYang: This seems like the clearest use case for Spark's broadcast variables (a large read-only Map to use in RDD or dataset operations). Without it, Spark would serialize/deserialize the name dictionaries for each worker node. Locally, it does make tests slower, and probably for our many small-to-medium size orgs we wouldn't see that much of a difference. But for those XL orgs it probably makes a difference. Open to other thoughts though; it doesn't look like TMOG uses broadcast variables at any other place.

MWYang: It was definitely a headache to figure out how to write the name detection code in a shared utils file while also using broadcast variables so there are some code maintainability gains to be had. :)

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
MWYang: I will add a `require` check—is that enough?

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
MWYang: Is that clearer? This bit of code is in the conditional branch when `treatAsName` already.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
MWYang: Using this thread to discuss all of the other related comments about Text vs. String:
1) The implicit conversions between `T#Value` and `String` are there to make my life easier when I'm porting over to `SmartTextMapVectorizer`, where the values are `String` and not `Text`.
2) As I discovered when erroneously making #449, TMOG does distinguish between `Text.empty` and `Text("")` (where the former is realized with Map types by excluding the relevant key/value from the Map altogether). However, for my application, I would like to treat `Text.empty` and `Text("")` the same way. This is clearer when I just operate over `String`.

With these two points in mind, would you still recommend that I operate over `Text` or `Option[String]`? It's not a big deal, I'm happy to implement it that way.

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
MWYang: They're arbitrarily chosen. I will break out these numbers into their own params, I should've done that before.

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -89,7 +91,7 @@ class HumanNameDetectorModel[T <: Text]
   val treatAsName: Boolean,
   val orderedGenderDetectStrategies: Seq[GenderDetectStrategy] = Seq.empty[GenderDetectStrategy]
 )(implicit tti: TypeTag[T])
-  extends UnaryModel[T, NameStats]("human name detector", uid) with NameDetectFun {
+  extends UnaryModel[T, NameStats]("humanNameDetect", uid) with NameDetectFun {
```
tovbinm: make `operationName` a ctor argument for the class so it would be easier to debug

core/src/test/scala/com/salesforce/op/stages/impl/feature/HumanNameDetectorTest.scala
```diff
@@ -193,6 +193,13 @@ class HumanNameDetectorTest
     ).toText)
     // scalastyle:on
     model.asInstanceOf[HumanNameDetectorModel[Text]].treatAsName shouldBe true
+    val metadataMap = model.getMetadata()
+    val metadata = HumanNameDetectorMetadata.fromMetadata(metadataMap)
+    println(metadata.genderResultsByStrategy)
```
tovbinm: remove println

core/src/main/scala/com/salesforce/op/utils/stages/NameDetectUtils.scala
tovbinm: I do this that using container types are beneficial than nullable values such as String. Aa for implicit conversions, they should be avoided as they are difficult to debug and spot in pull request reviews, therefore I would recommend against using them unless absolutely necessary.

tovbinm:
let's please follow the convention of camelCasing the parameters, e.g. replace `guard_maxNumberOfTokens` with `guardMaxNumberOfTokens` @MWYang

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -305,6 +305,14 @@ object NameStats {
     val LastName = "lastName"
     val Gender = "gender"
   }
+  // TODO: Use enumeratum for this
```
tovbinm: so why not use Enumerantum indeed? @MWYang

testkit/src/main/scala/com/salesforce/op/testkit/RandomMap.scala
```diff
@@ -157,6 +158,33 @@ object RandomMap {
   RandomMap[Long, M] =
     mapsOf[Long, M](valueGenerator.numbers.producer, minSize, maxSize, sources = Some(valueGenerator))
 
+  /**
```
tovbinm: nice!!

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
MWYang: Oops, just forgot to do it 😅

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -293,35 +293,34 @@ object StreetMap {
...
+    case object True  extends BooleanStrings
+    case object False extends BooleanStrings
   }
-  object GenderStrings {
-    val Male = "male"
-    val Female = "female"
-    val GenderNA = "NA"
-    val GenderNotInferred = "NotInferred"
+  sealed class GenderStrings extends EnumEntry
+  case object GenderStrings extends Enum[GenderStrings] {
```
tovbinm: Nit: IMHO, `GenderValues` is better. @MWYang

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -293,35 +293,34 @@ object StreetMap {
...
-    Keys.IsNameIndicator,
-    Keys.FirstName,
-    Keys.LastName,
-    Keys.Gender
-  )
-  object BooleanStrings {
-    val True = "true"
-    val False = "false"
+  sealed class BooleanStrings extends EnumEntry
+  case object BooleanStrings extends Enum[BooleanStrings] {
```
tovbinm: I think `BooleanStrings` enum is reduntant. simply use `true.toString` and `false.toString` instead @MWYang

tovbinm:
BTW, where are the `GenderDictionary_USandUK.csv` and `Names_JRC_Combined.txt` coming from? What's the license on these datasets? @MWYang

features/src/main/scala/com/salesforce/op/features/types/Maps.scala
```diff
@@ -280,6 +280,47 @@ object StreetMap {
...
+ * - "original"
+ * - "isName"
+ * - "firstName"
+ * - "lastName"
+ * - "gender"
+ */
+class NameStats(value: Map[String, String]) extends TextMap(value) {
+  import NameStats.Key._
+
+  def isName: Boolean = value.getOrElse(IsName.toString, "false") == "true"
```
tovbinm: replace `==` with `compareToIgnoreCase`

MWYang:
> BTW, where are the `GenderDictionary_USandUK.csv` and `Names_JRC_Combined.txt` coming from? What's the license on these datasets? @MWYang

They are both from governmental organizations. For the first, US and UK birth registry data, and for the second, the EU science commission. I also documented the pre-processing of those files in a separate repo: https://github.com/MWYang/InternationalNames

core/src/test/scala/com/salesforce/op/stages/impl/feature/HumanNameDetectorTest.scala
```diff
@@ -0,0 +1,367 @@
...
+      "Benjamin L. Cardin",
+      "Lisa Maria Blunt Rochester",
+      "Thomas Robert Carper",
+      "Jennifer González-Colón"
+    ).toText)
+    // scalastyle:on
+    model.asInstanceOf[HumanNameDetectorModel[Text]].treatAsName shouldBe true
+    val resultingMaps = result.collect().toSeq.map(row => row.get(1)).asInstanceOf[Seq[Map[String, String]]]
+    val identifiedGenders = resultingMaps.map(_.get(Gender.toString))
+    identifiedGenders shouldBe Seq(
```
tovbinm: here and in other tests you can replace
```
Seq(
      Some(Male.toString),
      Some(Female.toString),
      Some(Male.toString),
      Some(Female.toString),
      Some(Male.toString),
      Some(Female.toString)
    )
```
with
```scala
Seq(
      Male,
      Female,
      Male,
      Female,
      Male,
      Female
    ).map(v => Some(v.toString))
```

core/src/main/scala/com/salesforce/op/stages/impl/feature/HumanNameDetector.scala
```diff
@@ -0,0 +1,167 @@
...
+(
+  implicit tti: TypeTag[T],
+  override val ttiv: TypeTag[T#Value]
+) extends UnaryEstimator[T, NameStats](
+  uid = uid,
+  operationName = operationName
+) with NameDetectFun[T] {
+
+  def fitFn(dataset: Dataset[T#Value]): HumanNameDetectorModel[T] = {
+    require(!dataset.isEmpty, "Input dataset cannot be empty")
```
gerashegalov: Is it absolutely necessary that this layer checks for dataset emptiness? Look up the internal demo "Big Data is not empty" showing the downside of creating always-checks for rare cases. Replacing reduce with fold and the right definition of the zero for the use case often helps.

MWYang: I buy that the RDD emptiness check is expensive—this is a `Dataset` though, and the `isEmpty` function is provided by our own `RichDataset` helper and uses the `head` function to check the first element. `Dataset`s also don't have a `foldLeft` or `treeAggregate` function, so I would have to convert to an `RDD` first which is expensive.

gerashegalov: `head` triggers a query execution with split calculation. with dataset you can have custom `agg`. you can also take care of the issue by doing nothing and rely on reduce throwing NoSuchElementException. Or you can catch it and add more diagnostic if it's helpful.

tovbinm: Another option is to remove the check and live without it ;)

tovbinm:
@MWYang kudos on your 1st contribution!!! 🔥

PR #447, opened 2019-12-11
Skip serializing cardinality estimates in FeatureDistributions

winterslu:
**Related issues**
Tokens of text field are printing out while internal usage. 

Ignore cardinality field in serialization

**Additional context**
Previous PR #420 & #438

### DIFF
--- a/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
+++ b/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
@@ -31,30 +31,28 @@
 package com.salesforce.op
 
 import com.salesforce.op.evaluators._
-import com.salesforce.op.features.types._
+import com.salesforce.op.features.types.{Real, _}
 import com.salesforce.op.features.{Feature, FeatureDistributionType, FeatureLike}
 import com.salesforce.op.filters._
 import com.salesforce.op.stages.impl.classification._
+import com.salesforce.op.stages.impl.feature.{CombinationStrategy, TextStats}
 import com.salesforce.op.stages.impl.preparators._
 import com.salesforce.op.stages.impl.regression.{OpLinearRegression, OpXGBoostRegressor, RegressionModelSelector}
 import com.salesforce.op.stages.impl.selector.ModelSelectorNames.EstimatorType
-import com.salesforce.op.stages.impl.selector.{SelectedModelCombiner, SelectedCombinerModel, SelectedModel}
 import com.salesforce.op.stages.impl.selector.ValidationType._
+import com.salesforce.op.stages.impl.selector.{SelectedCombinerModel, SelectedModel, SelectedModelCombiner}
 import com.salesforce.op.stages.impl.tuning.{DataCutter, DataSplitter}
 import com.salesforce.op.test.{PassengerSparkFixtureTest, TestFeatureBuilder}
 import com.salesforce.op.testkit.RandomReal
 import com.salesforce.op.utils.spark.{OpVectorColumnMetadata, OpVectorMetadata}
+import com.twitter.algebird.Moments
 import org.apache.spark.ml.param.ParamMap
 import org.apache.spark.ml.tuning.ParamGridBuilder
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.functions._
 import org.junit.runner.RunWith
-import com.salesforce.op.features.types.Real
-import com.salesforce.op.stages.impl.feature.{CombinationStrategy, TextStats}
-import com.twitter.algebird.Moments
-import org.apache.spark.sql.{DataFrame, Dataset}
-import org.scalactic.Equality
 import org.scalatest.FlatSpec
 import org.scalatest.junit.JUnitRunner
-import org.apache.spark.sql.functions._
 
 import scala.util.{Failure, Success}
 
@@ -175,7 +173,7 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
     val insights = model.modelInsights(inputModel)
     val featureMoments = insights.features.map(f => f.featureName -> f.distributions.head.moments.get).toMap
     val featureCardinality = insights.features.map(f => f.featureName -> f.distributions.head.cardEstimate.get).toMap
-    return (featureMoments, featureCardinality)
+    featureMoments -> featureCardinality
   }
 
   val params = new OpParams()
@@ -795,10 +793,9 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
     }
     }
 
-    cardinality.foreach { case (featureName, value) => {
-      val actualUniques = df.select(featureName).as[Double].collect().toSet
-      value.valueCounts.keySet.map(_.toDouble).subsetOf(actualUniques) shouldBe true
-    }
+    cardinality.foreach { case (featureName, value) =>
+        val actualUniques = df.select(featureName).as[Double].distinct.collect.toSet
+        actualUniques should contain (value.valueCounts.keySet)
     }
   }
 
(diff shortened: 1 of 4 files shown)


TuanNguyen27 (approved):
LGTM, discussed internally.

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -793,6 +794,12 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
       math.abs((value.variance - expectedVariance) / expectedVariance) < varTol shouldBe true
     }
     }
+
+    cardinality.foreach { case (featureName, value) => {
+      val actualUniques = df.select(featureName).as[Double].collect().toSet
```
tovbinm: First do distinct then do collect.

tovbinm (commented):
See comments

gerashegalov (approved):
LGTM

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
```diff
@@ -238,4 +241,12 @@ class FeatureDistributionTest extends FlatSpec with PassengerSparkFixtureTest wi
     intercept[IllegalArgumentException](fd1.jsDivergence(fd1.copy(name = "boo"))) should have message
       "requirement failed: Name must match to compare or combine feature distributions: A != boo"
   }
+
+  it should "not serialize cardEstimate field" in {
+    val fd1 = FeatureDistribution("A", None, 10, 1, Array(1, 4, 0, 0, 6),
+      Array.empty, Some(Moments(1.0)), Some(TextStats(Map("foo" -> 1, "bar" ->2))),
+      FeatureDistributionType.Scoring)
+
+    FeatureDistribution.toJson(Seq(fd1)) shouldNot contain theSameElementsAs "cardEstimate"
```
gerashegalov: actually the test seems incorrect, it won't fail without the fix:
```scala
scala> "blah cardEstimate blah" shouldNot contain theSameElementsAs "cardEstimate"
res2: org.scalatest.Assertion = Succeeded
```

gerashegalov: include is captures it:
```scala
scala> "blah cardEstimate blah" shouldNot include ("cardEstimate")
org.scalatest.exceptions.TestFailedException: "blah cardEstimate blah" included substring "cardEstimate"
  at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:343)
  at org.scalatest.Matchers$ShouldMethodHelper$.shouldNotMatcher(Matchers.scala:6731)
  at org.scalatest.Matchers$AnyShouldWrapper.shouldNot(Matchers.scala:7151)
  ... 36 elided

scala> "blah cardEstimat blah" shouldNot include ("cardEstimate")
res4: org.scalatest.Assertion = Succeeded
```

gerashegalov (changes requested):

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
WintersLu: Good catch! I guess `contain` here works differently from `contains` to string

gerashegalov: http://www.scalatest.org/user_guide/using_matchers#workingWithAggregations
theSameElementAs was not the right choice because it can generate false positive for a permutation of the string order, and it does not have to be consecutive.

gerashegalov (approved):

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
```diff
@@ -238,4 +241,12 @@ class FeatureDistributionTest extends FlatSpec with PassengerSparkFixtureTest wi
     intercept[IllegalArgumentException](fd1.jsDivergence(fd1.copy(name = "boo"))) should have message
       "requirement failed: Name must match to compare or combine feature distributions: A != boo"
   }
+
+  it should "not serialize cardEstimate field" in {
+    val fd1 = FeatureDistribution("A", None, 10, 1, Array(1, 4, 0, 0, 6),
+      Array.empty, Some(Moments(1.0)), Some(TextStats(Map("foo" -> 1, "bar" ->2))),
+      FeatureDistributionType.Scoring)
+
+    FeatureDistribution.toJson(Seq(fd1)) shouldNot include ("cardEstimate")
```
tovbinm: How about testing reading from json?

tovbinm (approved):

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
```diff
@@ -238,4 +241,21 @@ class FeatureDistributionTest extends FlatSpec with PassengerSparkFixtureTest wi
...
       "requirement failed: Name must match to compare or combine feature distributions: A != boo"
   }
+
+  it should "not serialize cardEstimate field" in {
+    val fd1 = FeatureDistribution("A", None, 10, 1, Array(1, 4, 0, 0, 6),
+      Array.empty, Some(Moments(1.0)), Some(TextStats(Map("foo" -> 1, "bar" ->2))),
+      FeatureDistributionType.Scoring)
+    FeatureDistribution.toJson(Seq(fd1)) shouldNot include ("cardEstimate")
+
+    val json =
```
gerashegalov: You already computed this string on Line 249, you can store it in a variable.

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
```diff
@@ -238,4 +241,21 @@ class FeatureDistributionTest extends FlatSpec with PassengerSparkFixtureTest wi
...
+      """[{"name":"A","count":10,"nulls":1,"distribution":[1.0,4.0,0.0,0.0,6.0],
+        |"summaryInfo":[],"moments":{"m0":1,"m1":1.0,"m2":0.0,"m3":0.0,"m4":0.0},
+        |"cardEstimate":{"valueCounts":{"foo":1,"bar":2}},"type":"Scoring"}]
+        |""".stripMargin
+    FeatureDistribution.fromJson(json) match {
```
gerashegalov: more concisely:
```
FeatureDistribution.fromJson(json) shouldBe Success(Seq(fd1))
```

gerashegalov: I think we need the test case for being able to deserialize where cardEstimate is present in JSON

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
WintersLu: that json doesn't contains cardEstimate obj, which ignored while deserialize, i construct another example with it, which should be tested that fromJson can still deserialize this obj without ignoring it.

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
WintersLu: yes, this test case is doing so

gerashegalov: sorry, I meant the complementary case then without cardEstimate in JSON

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
gerashegalov: you can generate it too:
```
val jsonWithCardEstimate = Serialization.write(fd1)(DefaultFormats +
      EnumEntrySerializer.json4s[FeatureDistributionType](FeatureDistributionType))
```

leahmcguire (approved):

gerashegalov (approved):

Scutua (approved):

Scutua (approved):

PR #448, opened 2019-12-16
Add support for ignoring text that looks like IDs in SmartTextVectorizer

Jauntbox:
**Related issues**
N/A

Adds a few parameters to SmartTextVectorizer to allow for ignoring text fields that would be hashed (eg. not categorical) if they have a token length variance below a specified threshold (eg. to catch machine-generated IDs).

Other alternatives are a sort of topK token counting (eg. with a countMinSketch). This works, but is difficult to robustly scale with dataset size, and may be implemented later via Algebird's TopKCMS data structure. Filtering data by raw text length std dev, or by how well the text length distribution fits a poisson distribution performed better on synthetic data and requires less modifications to SmartTextVectorizer.

**Additional context**
Extra thing we need to be careful of is that we still use the CJK tokenizer for Chinese and Korean text (Japanese uses a proper language-specific tokenizer already), and this tokenizer always splits the text into character bigrams which would cause it to fail any length distribution checks. We will need to update the Korean & Chinese tokenizers to language-specific ones that will pick out words rather than bigrams.

We plan to also add a way to filter based on goodness of fit of the text length distribution to a Poisson distribution in a future PR. All the information is already available to do this, so the modifications should be straightforward.


core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -81,23 +81,27 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
     val maxCard = $(maxCardinality)
+    val minLenStdDev = $(minLengthStdDev)
     val shouldCleanText = $(cleanText)
 
     implicit val testStatsMonoid: Semigroup[TextStats] = TextStats.monoid(maxCard)
     val valueStats: Dataset[Array[TextStats]] = dataset.map(_.map(computeTextStats(_, shouldCleanText)).toArray)
     val aggregatedStats: Array[TextStats] = valueStats.reduce(_ + _)
 
-    val (isCategorical, topValues) = aggregatedStats.map { stats =>
+    val (isCategorical, isIgnorable, topValues) = aggregatedStats.map { stats =>
```
leahmcguire: can something be both ignorable and categorical? if not perhaps we should replace with an enum rather than multiplying our booleans...

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -81,23 +81,27 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
       val isCategorical = stats.valueCounts.size <= maxCard
+      val isIgnorable = stats.lengthStdDev <= minLenStdDev
```
leahmcguire: is this really all you need to identify IDs?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -117,18 +121,23 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
     require(inN.length == smartTextParams.isCategorical.length)
 
-    val (categoricalFeatures, textFeatures) =
+    val (categoricalFeatures, allTextFeatures) =
       SmartTextVectorizer.partition[TransientFeature](inN, smartTextParams.isCategorical)
+    // Also need to partition the masking array so that the correct features are still split out as ignorable
+    val (_, isIgnorableText) =
+      SmartTextVectorizer.partition[Boolean](smartTextParams.isIgnorable, smartTextParams.isCategorical)
+    val (textFeaturesIgnorable, textFeatures) = SmartTextVectorizer
+      .partition[TransientFeature](allTextFeatures, isIgnorableText)
```
leahmcguire: lets convert to an enum rather than additing booleans for every case...

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -156,38 +165,59 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
-      if (l.valueCounts.size > maxCardinality) l
-      else if (r.valueCounts.size > maxCardinality) r
-      else TextStats(l.valueCounts + r.valueCounts)
+      val newValueCounts = if (l.valueCounts.size > maxCardinality) l.valueCounts
+        else if (r.valueCounts.size > maxCardinality) r.valueCounts
+        else l.valueCounts + r.valueCounts
+
+      val newLengthCounts = if (l.lengthCounts.size > maxCardinality) l.lengthCounts
+        else if (r.lengthCounts.size > maxCardinality) r.lengthCounts
+        else l.lengthCounts + r.lengthCounts
```
leahmcguire: can we make this a function?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
Jauntbox: This is the simplest check that worked on most of the synthetic data. The topK counts from a count-min sketch worked about as well, but it was harder to construct a threshold that worked across different dataset sizes and required more changes to the TextStats class. We could add that in the future, but I wanted to try out the trivial one on real data first.

The best performing criteria was a goodness of fit threshold on how well the text lengths follow an MLE Poisson fit. I was going to add that one in another PR, but could put it in this PR too. It's simple enough that it shouldn't need any curve fitting libraries.

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
Jauntbox: Not currently. Right now "ignorable" refers just to fields that would have been hashed otherwise. If a field is already low cardinality, I don't think we'd have a reason to ignore it in general, even if it was an ID or something.

We could make this into an enum with three choices (Pivot, Hash, Ignore) if you think that's clearer.

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
Jauntbox: Probably, there's a chunk for `Map[String, Int]` and a chunk for `Map[Int, Int]` so we should be able to combine since each has a monoid.

MWYang:
@Jauntbox Regarding creating an enum to replace the booleans in SmartTextVectorizer, I have done this already on my personal branch for incorporating name detection in STV (https://github.com/MWYang/TransmogrifAI/pull/1/files) (Look for `SmartTextVectorizerAction`.) Hopefully that's helpful, even though my changes are a lot to look through right now. 😅

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -117,18 +121,23 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
   }
 
   private def computeTextStats(text: T#Value, shouldCleanText: Boolean): TextStats = {
-    val valueCounts = text match {
-      case Some(v) => Map(cleanTextFn(v, shouldCleanText) -> 1)
-      case None => Map.empty[String, Int]
+    val (valueCounts, lengthCounts) = text match {
+      case Some(v) => (Map(cleanTextFn(v, shouldCleanText) -> 1), Map(cleanTextFn(v, shouldCleanText).length -> 1))
```
gerashegalov: not specific here, but will we have `Int` overflows in larger texts with more than 2G uniques?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -72,7 +72,8 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
     textMap: T#Value, shouldCleanKeys: Boolean, shouldCleanValues: Boolean
   ): TextMapStats = {
     val keyValueCounts = textMap.map{ case (k, v) =>
-      cleanTextFn(k, shouldCleanKeys) -> TextStats(Map(cleanTextFn(v, shouldCleanValues) -> 1))
+      cleanTextFn(k, shouldCleanKeys) ->
+        TextStats(Map(cleanTextFn(v, shouldCleanValues) -> 1), Map(cleanTextFn(v, shouldCleanValues).length -> 1))
```
tovbinm: if looks like you are going to have some merge conflicts - https://github.com/salesforce/TransmogrifAI/pull/449

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -81,23 +81,27 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
+      val isIgnorable = stats.lengthStdDev <= minLenStdDev
       val topValues = stats.valueCounts
         .filter { case (_, count) => count >= $(minSupport) }
         .toSeq.sortBy(v => -v._2 -> v._1)
         .take($(topK)).map(_._1)
-      isCategorical -> topValues
-    }.unzip
+
+      (isCategorical, isIgnorable, topValues)
+    }.unzip3
```
tovbinm: nice! I did not even know this existed ;)

core/src/test/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizerTest.scala
```diff
@@ -191,6 +249,7 @@ class SmartTextVectorizerTest
     val transformed = new OpWorkflow().setResultFeatures(smartVectorized).transform(inputData)
 
     val meta = OpVectorMetadata(transformed.schema(smartVectorized.name))
+    meta.columns.foreach(println)
```
tovbinm: please remove println-s

leahmcguire:
"oof!" someone has been hanging out with @snabar :-P

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
Jauntbox: Replaced! Mostly - I still need to propagate the changes to SmartTextMapVectorizer, but this did make things more readable. Good suggestion!

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
Jauntbox: hmmm good point. That would require a huge dataset but it's an easy change to switch these to Longs

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
Jauntbox: and... back to `unzip` with the enum refactoring, lol

core/src/test/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizerTest.scala
Jauntbox: Oops, thought I found them all

snabar:
> "oof!" someone has been hanging out with @snabar :-P

Ooooooofffff!

tovbinm:
@Jauntbox lgtm. 

Compilation failed though - https://travis-ci.com/salesforce/TransmogrifAI/jobs/269487360#L695

tovbinm:
Also 
```
warning file=/home/travis/build/salesforce/TransmogrifAI/features/src/main/scala/com/salesforce/op/stages/impl/feature/TextVectorizationMethod.scala message=Header does not match expected text line=2
```
https://travis-ci.com/salesforce/TransmogrifAI/jobs/272953824#L508

PR #449, opened 2019-12-17
Fixed SmartTextMapVectorizer counting empty string entries

MWYang:
**Related issues**
Tangentially related to the name detection changes I've been making as part of my internship (#445 and other deprecated PRs).

Previously, `SmartTextMapVectorizer` would treat empty string values in input maps as their own category. This could lead to edge case behavior where an otherwise categorical input would be treated as non-categorical due to the presence of empty strings (probably unlikely to happen IRL but it has tripped me up in tests for my other work). This PR implements:
- the one-line fix in `SmartTextMapVectorizer` to not count empty strings when determining which input features are categorical
- a test for `SmartTextMapVectorizer` against this behavior
- the one-line fix in `TextMapPivotVectorizer` to treat empty string entries as null values rather than other values

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
@@ -72,7 +72,9 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
     textMap: T#Value, shouldCleanKeys: Boolean, shouldCleanValues: Boolean
   ): TextMapStats = {
     val keyValueCounts = textMap.map{ case (k, v) =>
-      cleanTextFn(k, shouldCleanKeys) -> TextStats(Map(cleanTextFn(v, shouldCleanValues) -> 1))
+      cleanTextFn(k, shouldCleanKeys) -> TextStats(
+        if (v.isEmpty) Map.empty[String, Int] else Map(cleanTextFn(v, shouldCleanValues) -> 1)
+      )
     }
     TextMapStats(keyValueCounts)
   }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -72,7 +72,9 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
     textMap: T#Value, shouldCleanKeys: Boolean, shouldCleanValues: Boolean
   ): TextMapStats = {
     val keyValueCounts = textMap.map{ case (k, v) =>
-      cleanTextFn(k, shouldCleanKeys) -> TextStats(Map(cleanTextFn(v, shouldCleanValues) -> 1))
+      cleanTextFn(k, shouldCleanKeys) -> TextStats(
```
tovbinm: do we have more places like this in the code? @Jauntbox @MWYang

Jauntbox:
Oh dear, I may have misinterpreted the issue you brought up @MWYang

Right now, we should be treating the empty string `""` differently from an empty dataframe entry (a true `null`). A mediocre example as to why they could be different: `""` could be the contents of a survey that was left blank, while `null` could correspond to someone who hasn't been presented with said survey.

So the main thing is that a Text column corresponding to `("A", null, "C")` should be treated the same as a TextMap column containing `(Map("f1" -> "A"), Map.empty, Map("f1" -> "C"))` while a Text column corresponding to `("A", "", "C")` should be treated the same as a TextMap column containing `(Map("f1" -> "A"), Map("f1" -> ""), Map("f1" -> "C"))`, which I think it is.

Part of the confusion is because we store empty elements slightly differently between map and non-map features. An empty element in a Text field will show up as a None inside our Text object, while an empty element in a TextMap field will not be added to the map at all (since the values are String rather than Option[String])

PR #455, opened 2020-01-11
Add support for ignoring text that looks like IDs in SmartTextMapVectorizer

Jauntbox:
**Related issues**
This is the map version of #448 

Adds a few parameters to SmartTextMapVectorizer to allow for ignoring text fields that would be hashed (eg. not categorical) if they have a token length variance below a specified threshold (eg. to catch machine-generated IDs).

Other alternatives are a sort of topK token counting (eg. with a countMinSketch). This works, but is difficult to robustly scale with dataset size, and may be implemented later via Algebird's TopKCMS data structure. Filtering data by raw text length std dev, or by how well the text length distribution fits a poisson distribution performed better on synthetic data and requires less modifications to SmartTextVectorizer.

**Additional context**
Extra thing we need to be careful of is that we still use the CJK tokenizer for Chinese and Korean text (Japanese uses a proper language-specific tokenizer already), and this tokenizer always splits the text into character bigrams which would cause it to fail any length distribution checks. We will need to update the Korean & Chinese tokenizers to language-specific ones that will pick out words rather than bigrams.

We plan to also add a way to filter based on goodness of fit of the text length distribution to a Poisson distribution in a future PR. All the information is already available to do this, so the modifications should be straightforward.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
@@ -316,7 +316,6 @@ private[op] trait MapHashingFun extends HashingFun {
   (
     hashFeatures: Array[TransientFeature],
     ignoreFeatures: Array[TransientFeature],
-    features: Array[TransientFeature],
     params: HashingFunctionParams,
     hashKeys: Seq[Seq[String]],
     ignoreKeys: Seq[Seq[String]],
@@ -337,25 +336,27 @@ private[op] trait MapHashingFun extends HashingFun {
         }.toArray
       } else {
         for {
-          (keys, f) <- hashKeys.toArray.zip(hashFeatures)
+          // Need to filter out empty key sequences since the hashFeatures only contain a map feature if one of their
+          // keys is to be hashed, but hashKeys contains a sequence per map (whether it's empty or not)
+          (keys, f) <- hashKeys.filter(_.nonEmpty).zip(hashFeatures)
           key <- keys
           i <- 0 until numHashes
         } yield f.toColumnMetaData().copy(grouping = Option(key))
-      }
+      }.toArray
 
     // All columns get null tracking or text length tracking, whether their contents are hashed or ignored
-    val allKeys = hashKeys.zip(ignoreKeys).map{ case(h, i) => h ++ i }
-    // val allFeatures =
+    val allTextKeys = hashKeys.zip(ignoreKeys).map{ case(h, i) => h ++ i }
+    val allTextFeatures = hashFeatures ++ ignoreFeatures
     val nullColumns = if (shouldTrackNulls) {
       for {
-        (keys, f) <- allKeys.toArray.zip(features)
+        (keys, f) <- allTextKeys.toArray.zip(allTextFeatures)
         key <- keys
       } yield f.toColumnMetaData(isNull = true).copy(grouping = Option(key))
     } else Array.empty[OpVectorColumnMetadata]
 
     val lenColumns = if (shouldTrackLen) {
       for {
-        (keys, f) <- allKeys.toArray.zip(features)
+        (keys, f) <- allTextKeys.toArray.zip(allTextFeatures)
         key <- keys
       } yield f.toColumnMetaData(descriptorValue = OpVectorColumnMetadata.TextLenString).copy(grouping = Option(key))
     } else Array.empty[OpVectorColumnMetadata]
--- a/core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
@@ -314,32 +314,38 @@ private[op] trait MapHashingFun extends HashingFun {
 
   protected def makeVectorColumnMetadata
   (
+    hashFeatures: Array[TransientFeature],
+    ignoreFeatures: Array[TransientFeature],
     features: Array[TransientFeature],
     params: HashingFunctionParams,
-    allKeys: Seq[Seq[String]],
+    hashKeys: Seq[Seq[String]],
+    ignoreKeys: Seq[Seq[String]],
     shouldTrackNulls: Boolean,
     shouldTrackLen: Boolean
   ): Array[OpVectorColumnMetadata] = {
     val numHashes = params.numFeatures
-    val numFeatures = allKeys.map(_.length).sum
+    val numFeatures = hashKeys.map(_.length).sum
     val hashColumns =
       if (isSharedHashSpace(params, Some(numFeatures))) {
         (0 until numHashes).map { i =>
           OpVectorColumnMetadata(
-            parentFeatureName = features.map(_.name),
-            parentFeatureType = features.map(_.typeName),
+            parentFeatureName = hashFeatures.map(_.name),
+            parentFeatureType = hashFeatures.map(_.typeName),
             grouping = None,
             indicatorValue = None
           )
         }.toArray
       } else {
         for {
-          (keys, f) <- allKeys.toArray.zip(features)
+          (keys, f) <- hashKeys.toArray.zip(hashFeatures)
           key <- keys
           i <- 0 until numHashes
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
```diff
@@ -314,32 +314,38 @@ private[op] trait MapHashingFun extends HashingFun {
...
   ): Array[OpVectorColumnMetadata] = {
     val numHashes = params.numFeatures
-    val numFeatures = allKeys.map(_.length).sum
+    val numFeatures = hashKeys.map(_.length).sum
     val hashColumns =
       if (isSharedHashSpace(params, Some(numFeatures))) {
         (0 until numHashes).map { i =>
           OpVectorColumnMetadata(
-            parentFeatureName = features.map(_.name),
-            parentFeatureType = features.map(_.typeName),
```
leahmcguire: is features still used elsewhere?

core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
```diff
@@ -314,32 +314,38 @@ private[op] trait MapHashingFun extends HashingFun {
...
         for {
-          (keys, f) <- allKeys.toArray.zip(features)
+          (keys, f) <- hashKeys.toArray.zip(hashFeatures)
           key <- keys
           i <- 0 until numHashes
         } yield f.toColumnMetaData().copy(grouping = Option(key))
       }
 
+    // All columns get null tracking or text length tracking, whether their contents are hashed or ignored
+    val allKeys = hashKeys.zip(ignoreKeys).map{ case(h, i) => h ++ i }
```
leahmcguire: what about pivoted keys?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -104,9 +104,24 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
       )
     } else Array.empty[OpVectorColumnMetadata]
 
-    val textColumns = if (args.textFeatureInfo.flatten.nonEmpty) {
+
+    /*
```
leahmcguire: why is this all commented out?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -116,28 +131,65 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
...
+        hashFeatures = hashFeatures.toArray,
+        ignoreFeatures = ignoreFeatures.toArray,
+        features = mapFeatures.toArray,
+        params = makeHashingParams(),
+        hashKeys = hashKeys,
+        ignoreKeys = ignoreKeys,
+        shouldTrackNulls = args.shouldTrackNulls,
+        shouldTrackLen = $(trackTextLen)
+      )
+    } else Array.empty[OpVectorColumnMetadata]
```
leahmcguire: do you not need to tract categorical?

core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
Jauntbox: Nope, removing it did not cause any problems anywhere else.

core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
Jauntbox: The pivot function should already do the necessary null tracking

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
Jauntbox: Oops, it was removed in the next commit

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
Jauntbox: Categorical tracking is done above in `categoricalColumns`

Jauntbox:
@leahmcguire Sorry, you looked at it right before I got a fix and new test in. It should be ready to go now.

core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
Jauntbox: I can rename to `allTextKeys` if your point is that it's a poor name

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -221,11 +249,20 @@ case class SmartTextMapVectorizerModelArgs
...
+    val ignoreGroup = groups.getOrElse(TextVectorizationMethod.Ignore, Seq.empty)
+
+    (catGroup, hashGroup, ignoreGroup)
+  }.unzip3
+
+  val categoricalKeys = categoricalFeatureInfo.map(_.map(_.key))
+  val hashKeys = hashFeatureInfo.map(_.map(_.key))
+  val ignoreKeys = ignoreFeatureInfo.map(_.map(_.key))
+
+  val textKeys = hashKeys.zip(ignoreKeys).map{ case (hk, ik) => hk ++ ik }
```
tovbinm: I am no sure what's going on between the lines 252-265. Perhaps add some docs?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -248,32 +285,47 @@ final class SmartTextMapVectorizerModel[T <: OPMap[String]] private[op]
   )
 
   private def partitionRow(row: Seq[OPMap[String]]):
-  (Seq[OPMap[String]], Seq[Seq[String]], Seq[OPMap[String]], Seq[Seq[String]]) = {
+  (Seq[OPMap[String]], Seq[Seq[String]], Seq[OPMap[String]], Seq[Seq[String]], Seq[OPMap[String]], Seq[Seq[String]]) = {
```
tovbinm: this return type is bizzare! let's add a private `case class PartitionRes(rowCategorical, keysCategorical, rowHash, keysHash, rowIgnore, keysIgnore)`

Jauntbox: Yeah, it's a bit silly - I can make it a case class for this

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
Jauntbox: I added some for the next commit

core/src/main/scala/com/salesforce/op/stages/impl/feature/OPCollectionHashingVectorizer.scala
```diff
@@ -314,42 +314,49 @@ private[op] trait MapHashingFun extends HashingFun {
...
+          (keys, f) <- hashKeys.filter(_.nonEmpty).zip(hashFeatures)
           key <- keys
           i <- 0 until numHashes
         } yield f.toColumnMetaData().copy(grouping = Option(key))
-      }
+      }.toArray
 
+    // All columns get null tracking or text length tracking, whether their contents are hashed or ignored
+    val allTextKeys = hashKeys.zip(ignoreKeys).map{ case(h, i) => h ++ i }
+    val allTextFeatures = hashFeatures ++ ignoreFeatures
```
leahmcguire: I still dont understand how this covers all of the features? what about the pivoted features?

PR #456, opened 2020-01-14
Incorporate name detection into SmartTextVectorizer

MWYang:
**Describe the proposed solution**
Incorporates the changes in #445 and #457 into `SmartTextVectorizer` and `SmartTextMapVectorizer`.

**Additional context**
Merge #457 before merging this PR. Compare the diff between this PR and that one on [my forked repo](https://github.com/MWYang/TransmogrifAI/compare/my/sensitive-metadata...MWYang:my/stv-detect-names).

Changes from #455 needs to be merged before this PR is ready.

PR #457, opened 2020-01-14
Metadata changes for sensitive feature information

MWYang:
Adds a place in `FeatureInsights` to track information about sensitive feature information (detected by something like #456). `OpVectorMetadata` now has a constructor argument for a map from feature names to a new class, `SensitiveFeatureInformation`, which will then be read out correctly into `FeatureInsights`.

**Additional context**
Merge this PR before merging #456.


utils/src/main/scala/com/salesforce/op/SensitiveFeatureInformation.scala
```diff
@@ -0,0 +1,165 @@
...
+ * @param name        the name of the raw feature
+ * @param key         optionally, the name of the key (if the raw feature is a Map type)
+ * @param actionTaken whether the handling of the raw feature changed b/c it was detected as sensitive
+ */
+sealed class SensitiveFeatureInformation
+(
+  val name: String,
+  val key: Option[String] = None,
+  val actionTaken: Boolean = false
+) extends EnumEntry with JsonLike {
```
Jauntbox: This is a bit confusing compared to the existing Enums we have (see eg. `CorrelationType`, `BinaryClassEvalMetrics`, `HashAlgorithm`, `TextVectorizationStrategy`, etc.)

Usually, an enum is a fixed categorical entity that we then use as input into a class that may do more complicated things like generate metadata. I don't see any examples of other enums where they themselves also implement extra methods or contain case classes (like the `SensitiveFeatureInformation` object you defined below. Is there a reason for this different type of structure?

I'm not 100% sure the design goals here, but I would structure it somewhat like
case class SensitiveFeatureInformation
(val name: String,
  val key: Option[String] = None,
  val actionTaken: Boolean = false) {

def toMetadata = ...
(Only define this here if there's some shared implementation. If there isn't, and you're just matching against different instances, then it's clearer to implement it in the child classes)

}

case class SensitiveNameInformation(
   val probName: Double,
   val probMale: Double,
   ...
 extends SensitiveFeatureInformation(name, key, actionTaken) {
   (implement the specific metadata generation here)
}

// Companion object
object SensitiveNameInformation {
   val EntryName = "Name"
   ...
}

I'm not sure you actually need an enum here, unless you want to restrict the value of a param somewhere to a discrete set of choices.

MWYang: Yeah, I think I was pre-optimizing a bit too much. I thought of doing the enum in order to add exhaustive pattern match checks—Basically, any of the reasons to use `enumeratum` over just inheriting a shared abstract class/trait. But none of those special `enumeratum` features are really used anywhere, so I'll just remove the enum.

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -690,6 +705,62 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
     insights.features.foreach(f => f.distributions shouldBe empty)
   }
 
+  it should
+    """include sensitive feature information
+      |even for sensitive features that are removed from output vector and output vector metadata""".stripMargin in {
+    // Copy metadata from above but add new feature that was removed in vectorizing to sensitive info
+    val f_notInMeta = Feature[Text]("f_notInMeta", false, null, Seq(), "test")
+    val newMeta = OpVectorMetadata(
```
Jauntbox: Can you clean up the creation of this object here? It would be easier to read if you construct the arguments individually in their own vals and then plug them in at the end so that you don't have to go to the definition of OpVectorMetadata to figure out what the args are supposed to be.

features/src/test/scala/com/salesforce/op/utils/spark/OPVectorMetadataTest.scala
```diff
@@ -72,24 +76,51 @@ class OPVectorMetadataTest extends PropSpec with TestCommon with PropertyChecks
   )
   val arrVecColTupleGen: Gen[Array[OpVectorColumnTuple]] = Gen.containerOf[Array, OpVectorColumnTuple](vecColTupleGen)
 
+  val sensitiveGen: Gen[SensitiveTuple] = for {
```
Jauntbox: Nice - I like the generators!

Maybe call this one `sensitiveNameGen` since it's specifically making info for name fields?

utils/src/test/scala/com/salesforce/op/SensitiveFeatureInformationTest.scala
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.spark.sql.types.MetadataBuilder
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class SensitiveFeatureInformationTest extends FlatSpec with TestCommon {
+
+  val probName = 1.0
+  val genderDetectResults: Seq[String] = Seq("ByIndex", "AnotherStrategy")
```
Jauntbox: You may want an enum here though. Is there a finite set of strategies?

utils/src/test/scala/com/salesforce/op/SensitiveFeatureInformationTest.scala
```diff
@@ -0,0 +1,134 @@
...
+package com.salesforce.op
+
+import com.salesforce.op.test.TestCommon
+import org.apache.spark.sql.types.MetadataBuilder
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class SensitiveFeatureInformationTest extends FlatSpec with TestCommon {
```
Jauntbox: Nice tests here!

utils/src/test/scala/com/salesforce/op/SensitiveFeatureInformationTest.scala
MWYang: There are a finite set of strategies, and they already are part of an enum. I think this comment is pointing at something else, which is that `SensitiveFeatureInformation.Name.genderDetectResults` is a `Seq[String]`—when I was writing it, I couldn't figure out the best way to serialize the gender detection results, which is composed of: a) the strategy b) the percentage of unidentified rows. Typing it out now, I think what I can do is make a new case class that just holds those two pieces of information, and then make the `SensitiveFeatureInformation.Name.genderDetectResults` a `Seq` of that case class. Does that make sense?

Jauntbox: Yeah, we do similar things in SmartTextVectorizer and for hashing in general. The args are packed into a case class that can be serialized. See eg this PR for the `TextVectorizationMethod` enum: https://github.com/salesforce/TransmogrifAI/pull/448/files and note the addition to `OpPipelineStageReaderWriter`

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -697,8 +698,41 @@ case object ModelInsights {
...
+            metrics = metrics, distributions = distributions, exclusionReasons = exclusionReasons,
+            sensitiveInformation = sensitiveFeatureInformation
+          )
+      }.toSeq ++ {
+        /*
+          Add FeatureInsights for removed sensitive fields that do not have a column in OpVectorMetadata.
+          With current TMOG settings, this will not happen unless null tracking is turned off since
+          null indicators are created for all text features, even ignored ones.
+        */
+        vectorInfo match {
```
Jauntbox: Hmmm, not a fan of having to do this. To make sure I understand, this is necessary because we determine ignorability (when not to hash or pivot) in the SmartTextVectorizer but determine the sensitive feature info in a separate estimator?

I thought the sensitive feature information was on a raw feature level, not a derived feature level. If so, then wouldn't sensitive feature info already be decoupled from the size of the derived features? Am I misunderstanding this?

MWYang: he sensitive feature information _is_ on the raw feature level _but_ in my earlier testing (when I was completely dropping any derived features for name columns, including null and length trackers), I found that `FeatureInsights` are only created for raw features that have derived features. I can probably(?) take this code out now though, since we've agreed that the `Ignore` STV method will still generate null/length tracker columns, so long as those features are enabled.

If I take the code out though, there is the situation that if we decide to turn off null/length tracking _and_ some fields are ignored, then sensitive feature information won't be generated for those raw features. But maybe this situation is unlikely enough that it's fine to remove this bit of code.

Jauntbox: Hmmm, I guess we should leave it in. It's not hurting anything, I'm just unhappy with our inconsistent treatment of when we decide to remove features.

tovbinm:
Congratulations  @MWYang!! This PR took code coverage to 87%!!!! 👍

PR #459, opened 2020-01-23
Fjiang/chat parsing

feifjiang:
**Related issues**
Adding a new Transformer and related tests to parse LiveChatTranscripts into a list of maps with key'value pairs of speaker and uttrances


build.gradle
```diff
@@ -64,11 +65,13 @@ configure(allProjs) {
         scalaGraphVersion = '1.12.5'
         scalafmtVersion = '1.5.1'
         hadoopVersion = 'hadoop2'
+        scalajCollVersion = '0.1.2'
```
tovbinm: Where is this being used?

build.gradle
```diff
@@ -64,11 +65,13 @@ configure(allProjs) {
...
         json4sVersion = '3.2.11' // matches Spark dependency version
         jodaTimeVersion = '2.9.4'
         jodaConvertVersion = '1.8.1'
         algebirdVersion = '0.13.4'
         jacksonVersion = '2.7.3'
+        jsoupVersion = '1.12.1'
```
tovbinm: Can we worked it out with Jackson or Json4s without adding an additional dependency?

features/src/main/scala/com/salesforce/op/features/types/Lists.scala
```diff
@@ -78,3 +78,12 @@ object DateTimeList {
   def empty: DateTimeList = FeatureTypeDefaults.DateTimeList
 }
 
+/**
+ * A list of map values
+ */
+class MapList(val value: Seq[TextMap]) extends OPList[TextMap]
```
tovbinm: Is it possible to use MultiPicklistMap instead of adding a new type?

leahmcguire: @feifjiang is the idea that this is a general HTML parsing format? Eg sequence of segments with a map(tag -> value)? And is this a general format? or is it specific to chats?

core/src/main/scala/com/salesforce/op/stages/impl/feature/ChatTranscriptTransformer.scala
```diff
@@ -0,0 +1,134 @@
...
+}
+
+  /** in cleanUtterrance,  matcher.matches() or matcher.find() needs to be called to allow group()
+   *  Stack overflow post: http://tiny.cc/dpm0iz
+   */
+object ChatTranscriptParser {
+
+  val agentExpr = Pattern.compile("""Agent ([\w\s]+)""")
+  val chatBotExpr = Pattern.compile("""Chatbot successfully""")
+  val responseTimeExpr = Pattern.compile("""\( [\w\s]+ \)""")
```
leahmcguire: should these be settable with these patterns as the default?

core/src/test/scala/com/salesforce/op/stages/impl/feature/ChatTranscriptTransformerTest.scala
```diff
@@ -0,0 +1,84 @@
...
+    """ Thank you for verifying that information, one moment while I access your account.  I understand you are trying
+      |to watch on your Mac Mini, and there is a blackscreen.""".stripMargin)
+
+  val map4 = Map("speaker"->"Customer", "text"->" yes")
+
+  val expectedResult = Seq(new MapList(List(
+    new TextMap(map1),
+    new TextMap(map2),
+    new TextMap(map3),
+    new TextMap(map4))))
```
gerashegalov: are standard assertions in OpTransformerSpec sufficient?

build.gradle
feifjiang: This shouldn't be there (It was from a version that was 1~2 days earlier than when I forked the repo). I'll fix it in the new PR.

build.gradle
feifjiang: We want a good java/scala libary to parse html, and based on my research Jsoup or HtmlCleaner came out as top options. https://tomassetti.me/parsing-html/  https://www.reddit.com/r/java/comments/4u8daa/best_html_parser_for_java/  Jsoup has better documentation and seems to be more widely-used than HtmlCleaner, hence the choice of Jsoup here.

features/src/main/scala/com/salesforce/op/features/types/Lists.scala
feifjiang: @leahmcguire  This was not intended as a general HTML parsing format - it's specific to Salesforce Live Chat Transcript. The sequence only consists of the agent/customer conversation, and each textmap within that sequence is a map indicating the speaker and his/her utterance, e.g. {"speaker": "Agent Tom", "text": "Hello!"}

build.gradle
```diff
@@ -90,6 +93,8 @@ configure(allProjs) {
         akkaSlf4jVersion = '2.3.11'
         mleapVersion = '0.13.0'
         memoryFilesystemVersion = '2.1.0'
+
+        mainClassName = 'com.salesforce.Main'
```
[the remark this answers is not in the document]
feifjiang: Sorry this was from a version that was 1~2 days earlier than when I forked the repo. I'll fix it in the new PR.

core/src/main/scala/com/salesforce/op/stages/impl/feature/ChatTranscriptTransformer.scala
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright (c) 2017, Salesforce.com, Inc.
```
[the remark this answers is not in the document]
feifjiang: Thanks for the note! I'll fix it in the new PR

features/src/main/scala/com/salesforce/op/features/types/Lists.scala
feifjiang: @tovbinm MultiPicklistMap wouldn't be applicable here, because the values are sets while we care about the order of the utterances from live chat.

build.gradle
tovbinm: Have you looked into using Tika HTMLParser? https://tika.apache.org/0.7/formats.html#HyperText_Markup_Language

tovbinm: Not sure which one brings less dependencies though and which one is more stable, but it would be nice to have the one with less dependencies and better support ;)

features/src/main/scala/com/salesforce/op/features/types/Lists.scala
tovbinm: you can always prefix values with index, e.g. `0 hello`, `1 how`, `2 are`, `3 you` ;)

I dont mind, it's just adding a new type is somewhat a hassle to maintain. wdyt? @leahmcguire @Jauntbox

core/src/main/scala/com/salesforce/op/stages/impl/feature/ChatTranscriptTransformer.scala
```diff
@@ -0,0 +1,134 @@
...
+
+import org.jsoup.Jsoup
+
+/** the Html parser is selected based on https://stackoverflow.com/questions/2168610/which-html-parser-is-the-best
+ */
+class ChatTranscriptTransformer[I <: Text](uid: String = UID[ChatTranscriptTransformer[_]])(implicit tti: TypeTag[I])
+  extends UnaryTransformer[I, MapList](operationName = "chatParsing", uid = uid) {
+
+  override def transformFn: I => MapList = (chats: I) => {
+    if ( !chats.isEmpty ) {
```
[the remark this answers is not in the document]
tovbinm: or pattern match and use SomeValue, e.g.
```scala
import op.salesforce.feature.types._

Text("test me") match {
   case SomeValue(Some(v)) => println(v)
   case  _ => // empty value ;(
}
```
Example - https://github.com/salesforce/TransmogrifAI/blob/57c9f7145fa631b0f594880f3e492618318f9ece/core/src/test/scala/com/salesforce/op/stages/impl/feature/NumericVectorizerTest.scala#L55

core/src/main/scala/com/salesforce/op/stages/impl/feature/ChatTranscriptTransformer.scala
```diff
@@ -0,0 +1,134 @@
...
+      val elements = Jsoup.parse( chats.value.get ).body().getElementsByTag("p").eachText().asScala.toList
```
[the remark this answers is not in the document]
leahmcguire: tag should be a parmeter

features/src/main/scala/com/salesforce/op/features/types/Lists.scala
feifjiang: @tovbinm  I chatted with Leah this morning and got the hassle of adding a new type :) . Will try to fit my use case to an existing TMOG type.

PR #460, opened 2020-01-31
[WIP] Experimenting with adaptive hash size based on token cardinality

TuanNguyen27:
**Related issues**

N/A

Current feature engineering for raw text feature uses a hard-coded size for hash space. For raw features that contain free text, this hash space isn't enough to fully express their content. This PR approximates token cardinality for each raw feature of text type using [HyperLogLog](https://twitter.github.io/algebird/datatypes/approx/hyperloglog.html) and uses this information to set the hash space size. [Related article on how to set this size](https://booking.ai/dont-be-tricked-by-the-hashing-trick-192a6aae3087)

Embeddings for text field, which currently don't work as well ? (To be further investigated). There's also an approach that combines [hashing and embeddings](https://arxiv.org/pdf/1709.03933.pdf) that saves a lot of memory. 

**Additional context**


core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -40,16 +40,16 @@ import com.twitter.algebird.Monoid
  * @param sum   sum of values for double, total number of tokens for text
  * @param count number of doubles for double, number of texts for text
  */
-case class Summary(min: Double, max: Double, sum: Double, count: Double)
+case class Summary(min: Double, max: Double, sum: Double, count: Double, hll: HLL)
```
tovbinm: please update docs on the Summary class

core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -59,8 +59,8 @@ case object Summary {
    */
   def apply(preppedFeature: ProcessedSeq): Summary = {
     preppedFeature match {
-      case Left(v) => Summary(v.size, v.size, v.size, 1.0)
-      case Right(v) => monoid.sum(v.map(d => Summary(d, d, d, 1.0)))
+      case Left(v) => Summary(v.size, v.size, v.size, 1.0, v.map(d => hllMonoid.create(d.getBytes)).reduce(_+_))
```
tovbinm: what's the type of `d` here?

core/src/main/scala/com/salesforce/op/filters/Summary.scala
```diff
@@ -59,8 +59,8 @@ case object Summary {
...
+      case Right(v) => monoid.sum(v.map(d => Summary(d, d, d, 1.0, hllMonoid.create(Array(d.toByte)))))
```
tovbinm: `d.toByte` seems invalid to me. Google around on how to convert Double value to array of bytes.

core/src/main/scala/com/salesforce/op/filters/Summary.scala
TuanNguyen27: `v` is a Seq[String] so d is a `String`

core/src/main/scala/com/salesforce/op/filters/Summary.scala
TuanNguyen27: Most solutions i found are kinda clunky. How do you feel about `hllMonoid.create(d.toString.getBytes)` ?

tovbinm: Either way you do it: a) it has to be fast b) is has to be consistent. I bet there is already some utility in our dependencies that does Double -> Array[Byte] conversion for you.

PR #463, opened 2020-02-26
Standalone minimum variance estimator

clin-projects:
**Related issues**
N/A

Standalone unary estimator to perform a minimum variance filter on derived features. Move shared functionality out of SanityChecker into DerivedFeatureFilterUtils object

_Alternative 1: Gated Params_
* Put all SanityChecker filters behind params so that they can be disabled
* Currently, SanityChecker member functions are private; we would make public those functions needed by UnsupervisedSanityChecker
* UnsupervisedSanityChecker = unaryEstimator that calls functions from SanityChecker
* Keep SanityCheckerParams the same for both (there will just be some extra that do not apply to unary case)

_Alternative 2: Minimal Wrapper Function_
* Create a wrapper function that creates a new SanityChecker stage using a dummy response and sets trivial thresholds to deactivate filters that use the response

**Additional context**
We have a need for a minimum variance filter in an _unsupervised_ (i.e., label-less) setting. While SanityChecker already has a minimum variance filter, it is a BinaryEstimator and assumes a (response, features) pair as input

### DIFF
--- a/core/src/main/scala/com/salesforce/op/dsl/RichVectorFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichVectorFeature.scala
@@ -143,6 +143,13 @@ trait RichVectorFeature {
       new DropIndicesByTransformer(matchFn = matchFn).setInput(f).getOutput()
     }
 
+    /**
+     * Apply filter that removes computed features that have variance <= `minVariance``
+     *
+     * @param minVariance
+     * @param removeBadFeatures
+     * @return
+     */
     def minVariance
     (
       minVariance: Double = MinVarianceFilter.MinVariance,
--- a/core/src/main/scala/com/salesforce/op/dsl/RichVectorFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichVectorFeature.scala
@@ -35,6 +35,7 @@ import com.salesforce.op.features.FeatureLike
 import com.salesforce.op.features.types._
 import com.salesforce.op.stages.impl.classification.{Impurity, OpRandomForestClassifier}
 import com.salesforce.op.stages.impl.feature.{DropIndicesByTransformer, OpLDA}
+import com.salesforce.op.stages.impl.preparators.MinVarianceFilter
 import com.salesforce.op.stages.sparkwrappers.specific.OpEstimatorWrapper
 import com.salesforce.op.utils.spark.{OpVectorColumnMetadata, OpVectorMetadata}
 import org.apache.spark.ml.feature.{IDF, IDFModel}
@@ -74,7 +75,8 @@ trait RichVectorFeature {
      * @param thresholds
      * @return
      */
-    def randomForest(
+    def randomForest
+    (
       label: FeatureLike[RealNN],
       maxDepth: Int = 5,
       maxBins: Int = 32,
@@ -132,6 +134,7 @@ trait RichVectorFeature {
      * Allows columns to be dropped from a feature vector based on properties of the
      * metadata about what is contained in each column (will work only on vectors)
      * created with [[OpVectorMetadata]]
+     *
      * @param matchFn function that goes from [[OpVectorColumnMetadata]] to boolean for dropping
      *                columns (cases that evaluate to true will be dropped)
      * @return new Vector with columns removed by function
@@ -139,6 +142,19 @@ trait RichVectorFeature {
     def dropIndicesBy(matchFn: OpVectorColumnMetadata => Boolean): FeatureLike[OPVector] = {
       new DropIndicesByTransformer(matchFn = matchFn).setInput(f).getOutput()
     }
+
+    def minVariance
+    (
+      minVariance: Double = MinVarianceFilter.MinVariance,
+      removeBadFeatures: Boolean = MinVarianceFilter.RemoveBadFeatures
+    ): FeatureLike[OPVector] = {
+      val filter = new MinVarianceFilter()
+      filter.setInput(f)
+        .setMinVariance(minVariance)
+        .setRemoveBadFeatures(removeBadFeatures)
+        .getOutput()
+    }
+
   }
 
 }
--- a/core/src/main/scala/com/salesforce/op/dsl/RichVectorFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichVectorFeature.scala
@@ -150,7 +150,7 @@ trait RichVectorFeature {
      * @param removeBadFeatures
      * @return
      */
-    def minVariance
+    def filterMinVariance
     (
       minVariance: Double = MinVarianceFilter.MinVariance,
       removeBadFeatures: Boolean = MinVarianceFilter.RemoveBadFeatures
--- a/core/src/main/scala/com/salesforce/op/stages/impl/preparators/DerivedFeatureFilterUtils.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/preparators/DerivedFeatureFilterUtils.scala
@@ -89,7 +89,7 @@ object DerivedFeatureFilterUtils {
    * @param corrIndices       Indices that we actually compute correlations for (eg. can ignore hashed text features)
    * @param categoricalStats  Array of CategoricalGroupStats for each group of feature vector indices corresponding
    *                          to a categorical feature
-   * @return
+   * @return Array of ColumnStatistics objects, one for each column in `metaCols`
    */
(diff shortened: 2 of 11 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/preparators/MinVarianceFilter.scala
```diff
@@ -0,0 +1,258 @@
...
+    logInfo("Calculating features to remove")
+    val toDropFeatures = if (removeBad) getFeaturesToDrop(stats) else Array.empty[ColumnStatistics]
+    val toDropSet = toDropFeatures.flatMap(_.column).toSet
+    val outputFeatures = vectorMetaColumns.filterNot { col => toDropSet.contains(col) }
+    val indicesToKeep = outputFeatures.map(_.index)
+
+    val outputMeta = OpVectorMetadata(getOutputFeatureName, outputFeatures, vectorMeta.history)
+
+    val summaryMetadata = {
+      val featuresStatistics = new SummaryStatistics(colStats, sample = 1.0)
```
leahmcguire: In general we ant to have a case class for metadata with a toMetadata and fromMetadata method to make working with these easier

core/src/main/scala/com/salesforce/op/stages/impl/preparators/MinVarianceFilter.scala
```diff
@@ -0,0 +1,254 @@
...
+ * (2) only filters features by variance
+ */
+class MinVarianceFilter
+(
+  operationName: String = classOf[MinVarianceFilter].getSimpleName,
+  uid: String = UID[MinVarianceFilter]
+) extends UnaryEstimator[OPVector, OPVector](operationName = operationName, uid = uid)
+  with MinVarianceFilterParams {
+
+  private def makeColumnStatistics
```
leahmcguire: rather than copy pasting these functions is it possible to move them to a shared companion object and just have default empty values for the parts you dont use?

core/src/main/scala/com/salesforce/op/dsl/RichVectorFeature.scala
```diff
@@ -132,13 +134,34 @@ trait RichVectorFeature {
...
     }
+
+    /**
+     * Apply filter that removes computed features that have variance <= `minVariance``
+     *
+     * @param minVariance
+     * @param removeBadFeatures
+     * @return
+     */
+    def minVariance
```
tovbinm: perhaps name this `filterMinVariance`

core/src/main/scala/com/salesforce/op/stages/impl/preparators/DerivedFeatureFilterUtils.scala
```diff
@@ -0,0 +1,418 @@
...
+          maxRuleConfidences = Seq.empty,
+          supports = Seq.empty
+        )
+        (labelStats +: featuresStats).toArray
+      }
+    }
+    columnStatistics
+  }
+
+  def getFeaturesToDrop
```
tovbinm: docs please

core/src/main/scala/com/salesforce/op/stages/impl/preparators/DerivedFeatureFilterUtils.scala
```diff
@@ -0,0 +1,418 @@
...
+        removedGroups = ruleConfGroupsToDrop
+      )
+      if reasons.nonEmpty
+    } yield {
+      val warning = s"Removing ${col.name} due to: ${reasons.mkString(",")}"
+      (col, warning)
+    }
+  }
+
+  def removeFeatures
```
tovbinm: same here - docs please

core/src/main/scala/com/salesforce/op/stages/impl/preparators/MinVarianceFilter.scala
```diff
@@ -0,0 +1,161 @@
...
+  override def fitFn(data: Dataset[OPVector#Value]): UnaryModel[OPVector, OPVector] = {
+    // Set the desired log level
+    if (isSet(logLevel)) {
+      Option(log).collect { case l: Log4jLoggerAdapter =>
+        LogManager.getLogger(l.getName).setLevel(Level.toLevel($(logLevel)))
+      }
+    }
+    val removeBad = $(removeBadFeatures)
+
+    logInfo("Getting vector rows")
```
tovbinm: Do these have to be log info statements? debug seem more appropriate.

core/src/test/scala/com/salesforce/op/stages/impl/preparators/MinVarianceFilterMetadataTest.scala
```diff
@@ -0,0 +1,76 @@
...
+    retrieved.featuresStatistics.variance should contain theSameElementsAs summary.featuresStatistics.variance
+    retrieved.names should contain theSameElementsAs summary.names
+  }
+
+  it should "convert to and from JSON and give the same values" in {
+    val meta = summary.toMetadata()
+    val json = meta.wrapped.prettyJson
+    val recovered = Metadata.fromJson(json)
+
+    // recovered shouldBe meta
```
tovbinm: I would prefer to compare properly, instead of the hashcode (since it's difficult to troubleshoot).

core/src/test/scala/com/salesforce/op/stages/impl/preparators/MinVarianceFilterTest.scala
```diff
@@ -0,0 +1,287 @@
...
+import com.salesforce.op.stages.base.unary.{UnaryEstimator, UnaryModel}
+import com.salesforce.op.test.{OpEstimatorSpec, TestFeatureBuilder}
+import com.salesforce.op.utils.spark.{OpVectorColumnMetadata, OpVectorMetadata}
+import org.apache.spark.ml.linalg.{Vector, Vectors}
+import org.apache.spark.sql.types.Metadata
+import org.apache.spark.sql.{DataFrame, Row}
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+case class UnlabeledTextRawData
```
tovbinm: move this class to the end of the file

core/src/main/scala/com/salesforce/op/stages/impl/preparators/MinVarianceFilter.scala
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.log4j.{Level, LogManager}
+import org.apache.spark.ml.linalg.{DenseVector, SparseVector}
+import org.apache.spark.mllib.linalg.{Vector => OldVector, Vectors => OldVectors}
+import org.apache.spark.mllib.stat.Statistics
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.Dataset
+import org.slf4j.impl.Log4jLoggerAdapter
+
+
+trait MinVarianceFilterParams extends DerivedFeatureFilterParams {
```
tovbinm: move this trait to the end of the file

leahmcguire: If you are just setting defaults for these values you dont need a separate trait just set them in the class

core/src/test/scala/com/salesforce/op/stages/impl/preparators/MinVarianceFilterTest.scala
```diff
@@ -0,0 +1,288 @@
...
+    val metadataOrig = transformedDataSchema(outputColName).metadata
+    val metadata = validateEncodeDecode(metadataOrig)
+    metadata
+  }
+
+  private def validateEncodeDecode(metadata: Metadata): Metadata = {
+    val metaParam = new MetadataParam("parent", "name", "doc")
+    val jsonEncodedMeta = metaParam.jsonEncode(metadata)
+    val jsonDecodedMeta = metaParam.jsonDecode(jsonEncodedMeta)
+    jsonDecodedMeta.hashCode() shouldEqual metadata.hashCode()
```
tovbinm: @clin-projects is there no other way to compare except the hashcode?

clin-projects: Changed to string comparison

clin-projects:
@leahmcguire @tovbinm All comments addressed. Let me know if further changes are needed, or whether we can merge, thanks!!!

tovbinm:
Great work!

PR #464, opened 2020-03-05
Allow TextStats length distribution to be token-based and refactor for testability

Jauntbox:
**Related issues**
n/a

Tests did not catch that the token length distributions added to TextStats were actually entry length distributions. This PR refactors some of the functions in TextTokenizer, SmartTextVectorizer, and SmartTextMapVectorizer so that they are directly testable. It also adds more robust tests to check desired behavior of the TextStats object.

n/a

**Additional context**
n/a

### DIFF
--- a/core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
@@ -273,6 +273,10 @@ trait RichMapFeature {
      * @param defaultLanguage           default language to assume in case autoDetectLanguage is disabled or
      *                                  failed to make a good enough prediction.
      * @param hashAlgorithm             hash algorithm to use
+     * @param tokenizeForLengths        If true, then the length counts will be lengths of the tokens in the entries.
+     *                                  If false, then the length counts will be the lengths of the entire entries
+     * @param minLengthStdDev           minimum standard deviation of the lengths of tokens in a text field for it to
+     *                                  be hashed instead of ignored
      * @param others                    additional text features
      * @return result feature of type Vector
      */
@@ -298,6 +302,8 @@ trait RichMapFeature {
       hashSpaceStrategy: HashSpaceStrategy = TransmogrifierDefaults.HashSpaceStrategy,
       defaultLanguage: Language = TextTokenizer.DefaultLanguage,
       hashAlgorithm: HashAlgorithm = TransmogrifierDefaults.HashAlgorithm,
+      tokenizeForLengths: Boolean = SmartTextVectorizer.TokenizeForLengths,
+      minLengthStdDev: Double = SmartTextVectorizer.MinTextLengthStdDev,
       others: Array[FeatureLike[TextMap]] = Array.empty
     ): FeatureLike[OPVector] = {
       // scalastyle:on parameter.number
@@ -322,6 +328,8 @@ trait RichMapFeature {
         .setHashSpaceStrategy(hashSpaceStrategy)
         .setHashAlgorithm(hashAlgorithm)
         .setBinaryFreq(binaryFreq)
+        .setTokenizeForLengths(tokenizeForLengths)
+        .setMinLengthStdDev(minLengthStdDev)
         .getOutput()
     }
   }
@@ -418,6 +426,10 @@ trait RichMapFeature {
      * @param defaultLanguage           default language to assume in case autoDetectLanguage is disabled or
      *                                  failed to make a good enough prediction.
      * @param hashAlgorithm             hash algorithm to use
+     * @param tokenizeForLengths        If true, then the length counts will be lengths of the tokens in the entries.
+     *                                  If false, then the length counts will be the lengths of the entire entries
+     * @param minLengthStdDev           minimum standard deviation of the lengths of tokens in a text field for it to
+     *                                  be hashed instead of ignored
      * @param others                    additional text features
      * @return result feature of type Vector
      */
@@ -443,6 +455,8 @@ trait RichMapFeature {
       hashSpaceStrategy: HashSpaceStrategy = TransmogrifierDefaults.HashSpaceStrategy,
       defaultLanguage: Language = TextTokenizer.DefaultLanguage,
       hashAlgorithm: HashAlgorithm = TransmogrifierDefaults.HashAlgorithm,
+      tokenizeForLengths: Boolean = SmartTextVectorizer.TokenizeForLengths,
+      minLengthStdDev: Double = SmartTextVectorizer.MinTextLengthStdDev,
       others: Array[FeatureLike[TextAreaMap]] = Array.empty
     ): FeatureLike[OPVector] = {
       // scalastyle:on parameter.number
@@ -467,6 +481,8 @@ trait RichMapFeature {
         .setHashSpaceStrategy(hashSpaceStrategy)
         .setHashAlgorithm(hashAlgorithm)
         .setBinaryFreq(binaryFreq)
+        .setTokenizeForLengths(tokenizeForLengths)
+        .setMinLengthStdDev(minLengthStdDev)
         .getOutput()
     }
   }
--- a/core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
+++ b/core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
@@ -273,8 +273,8 @@ trait RichMapFeature {
      * @param defaultLanguage           default language to assume in case autoDetectLanguage is disabled or
      *                                  failed to make a good enough prediction.
      * @param hashAlgorithm             hash algorithm to use
-     * @param tokenizeForLengths        If true, then the length counts will be lengths of the tokens in the entries.
-     *                                  If false, then the length counts will be the lengths of the entire entries
+     * @param textLengthType            Method to use for constructing text length distribution in TextStats. Current
+     *                                  options are from the full entry or from the tokens
      * @param minLengthStdDev           minimum standard deviation of the lengths of tokens in a text field for it to
      *                                  be hashed instead of ignored
      * @param others                    additional text features
@@ -302,7 +302,7 @@ trait RichMapFeature {
       hashSpaceStrategy: HashSpaceStrategy = TransmogrifierDefaults.HashSpaceStrategy,
       defaultLanguage: Language = TextTokenizer.DefaultLanguage,
       hashAlgorithm: HashAlgorithm = TransmogrifierDefaults.HashAlgorithm,
-      tokenizeForLengths: Boolean = SmartTextVectorizer.TokenizeForLengths,
+      textLengthType: TextLengthType = SmartTextVectorizer.LengthType,
       minLengthStdDev: Double = SmartTextVectorizer.MinTextLengthStdDev,
       others: Array[FeatureLike[TextMap]] = Array.empty
     ): FeatureLike[OPVector] = {
@@ -328,7 +328,7 @@ trait RichMapFeature {
         .setHashSpaceStrategy(hashSpaceStrategy)
         .setHashAlgorithm(hashAlgorithm)
         .setBinaryFreq(binaryFreq)
-        .setTokenizeForLengths(tokenizeForLengths)
+        .setTextLengthType(textLengthType)
         .setMinLengthStdDev(minLengthStdDev)
         .getOutput()
     }
@@ -455,7 +455,7 @@ trait RichMapFeature {
       hashSpaceStrategy: HashSpaceStrategy = TransmogrifierDefaults.HashSpaceStrategy,
       defaultLanguage: Language = TextTokenizer.DefaultLanguage,
       hashAlgorithm: HashAlgorithm = TransmogrifierDefaults.HashAlgorithm,
-      tokenizeForLengths: Boolean = SmartTextVectorizer.TokenizeForLengths,
+      textLengthType: TextLengthType = SmartTextVectorizer.LengthType,
       minLengthStdDev: Double = SmartTextVectorizer.MinTextLengthStdDev,
       others: Array[FeatureLike[TextAreaMap]] = Array.empty
     ): FeatureLike[OPVector] = {
@@ -481,7 +481,7 @@ trait RichMapFeature {
         .setHashSpaceStrategy(hashSpaceStrategy)
         .setHashAlgorithm(hashAlgorithm)
         .setBinaryFreq(binaryFreq)
-        .setTokenizeForLengths(tokenizeForLengths)
+        .setTextLengthType(textLengthType)
         .setMinLengthStdDev(minLengthStdDev)
         .getOutput()
     }
--- a/core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
+++ b/core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
@@ -33,7 +33,8 @@ package com.salesforce.op.filters
 import java.util.Objects
 
 import com.salesforce.op.features.{FeatureDistributionLike, FeatureDistributionType}
-import com.salesforce.op.stages.impl.feature.{HashAlgorithm, Inclusion, NumericBucketizer, TextStats}
+import com.salesforce.op.stages.impl.feature.{HashAlgorithm, Inclusion, NumericBucketizer, TextStats,
+  SmartTextVectorizer}
 import com.salesforce.op.utils.json.EnumEntrySerializer
 import com.twitter.algebird.Monoid._
 import com.twitter.algebird._
@@ -282,7 +283,15 @@ object FeatureDistribution {
    * @return TextStats object containing a Map from a value to its frequency (histogram)
    */
   private def cardinalityValues(values: ProcessedSeq): TextStats = {
-    TextStats(countStringValues(values.left.getOrElse(values.right.get)), Map.empty)
+    implicit val testStatsMonoid: Monoid[TextStats] = TextStats.monoid(RawFeatureFilter.MaxCardinality)
+
+    val stringVals = values match {
+      case Left(stringSeq) => stringSeq
+      case Right(doubleSeq) => doubleSeq.map(_.toString)
+    }
+    stringVals.foldLeft(TextStats.empty)((acc, el) => acc + SmartTextVectorizer.computeTextStats(
+      Option(el), shouldCleanText = false, maxCardinality = RawFeatureFilter.MaxCardinality)
+    )
   }
 
   private def countStringValues[T](seq: Seq[T]): Map[String, Long] = {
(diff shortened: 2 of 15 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -203,6 +192,26 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
   }
 }
 
+object SmartTextMapVectorizer extends CleanTextFun {
+
+  private[op] def computeTextMapStats[T <:  OPMap[String]](
```
tovbinm: please add docs for the function

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -164,13 +158,31 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
+object SmartTextVectorizer extends CleanTextFun {
   val MaxCardinality: Int = 100
   val MinTextLengthStdDev: Double = 0
+
   private[op] def partition[T: ClassTag](input: Array[T], condition: Array[Boolean]): (Array[T], Array[T]) = {
     val all = input.zip(condition)
     (all.collect { case (item, true) => item }, all.collect { case (item, false) => item })
   }
+
+  private[op] def computeTextStats[T <: Text : TypeTag](
```
tovbinm: same here

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -189,8 +201,8 @@ private[op] case class TextStats
   val lengthMean: Double = lengthCounts.foldLeft(0.0)((acc, el) => acc + el._1 * el._2) / lengthSize
   val lengthVariance: Double = lengthCounts.foldLeft(0.0)(
     (acc, el) => acc + el._2 * (el._1 - lengthMean) * (el._1 - lengthMean)
-  )
-  val lengthStdDev: Double = math.sqrt(lengthVariance / lengthSize)
+  ) / lengthSize
```
tovbinm: can `lengthSize` be `0`?

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
```diff
@@ -162,30 +162,84 @@ object TextTokenizer {
...
+   * @param analyzer            text analyzer instance
+   * @param sentenceSplitter    sentence splitter instance
+   * @param autoDetectLanguage  whether to attempt language detection
+   * @param defaultLanguage     default language
+   * @param autoDetectThreshold language detection threshold
+   * @param toLowercase         whether to convert all characters to lowercase before tokenizing
+   * @param minTokenLength      minimum token length
+   * @return detected language and sentence tokens
+   */
+  def tokenize(
```
tovbinm: That's a bummer that you have to add so much boiler plate here. There are alternatives we can discuss.

utils/src/main/scala/com/salesforce/op/test/TestCommon.scala
```diff
@@ -130,4 +130,22 @@ trait TestCommon extends Matchers with Assertions {
...
+   * @param tol       Tolerance for comparison. Comparison succeeds if the difference (absolute or relative) is less
+   *                  than tol.
+   * @return          Assertion to check approximate equality between actual and expected
+   */
+  def compareWithTol(actual: Double, expected: Double, tol: Double): Assertion =
+    withClue(s"Real number comparison failure! Actual: $actual, Expected: $expected \n") {
+      if (expected != 0) {
+        math.abs((actual - expected) / expected) < tol shouldBe true
+      }
+      else math.abs(actual - expected) < tol shouldBe true
```
tovbinm: `math.abs(actual - expected) should be < tol`

utils/src/main/scala/com/salesforce/op/test/TestCommon.scala
```diff
@@ -130,4 +130,22 @@ trait TestCommon extends Matchers with Assertions {
...
+   * @param actual    Actual value produced
+   * @param expected  Expected value
+   * @param tol       Tolerance for comparison. Comparison succeeds if the difference (absolute or relative) is less
+   *                  than tol.
+   * @return          Assertion to check approximate equality between actual and expected
+   */
+  def compareWithTol(actual: Double, expected: Double, tol: Double): Assertion =
+    withClue(s"Real number comparison failure! Actual: $actual, Expected: $expected \n") {
+      if (expected != 0) {
+        math.abs((actual - expected) / expected) < tol shouldBe true
```
tovbinm: `math.abs((actual - expected) / expected) should be < tol`

utils/src/main/scala/com/salesforce/op/test/TestCommon.scala
```diff
@@ -130,4 +130,22 @@ trait TestCommon extends Matchers with Assertions {
...
+   * It uses relative differences unless the expected value is zero, in which case it uses absolution differences.
+   *
+   * @param actual    Actual value produced
+   * @param expected  Expected value
+   * @param tol       Tolerance for comparison. Comparison succeeds if the difference (absolute or relative) is less
+   *                  than tol.
+   * @return          Assertion to check approximate equality between actual and expected
+   */
+  def compareWithTol(actual: Double, expected: Double, tol: Double): Assertion =
+    withClue(s"Real number comparison failure! Actual: $actual, Expected: $expected \n") {
```
tovbinm: nice

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
Jauntbox: Yep, and I think this is the behavior we want (these derived quantities being Double.NaN in this case).

The aggregations are already on the maps themselves, so these results being NaN will not propagate through - they are just descriptives. In the unlikely case where an entire column consists of stop words or something that doesn't tokenize (so lengthSize = 0), then we won't trigger removal either since Double.NaN < minLengthStdDev will always be false.

I can add a test where we check the values when lengthSize is 0 to make it more explicit.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
Jauntbox: Yes, I was not too happy with this either, but could not think of a clean alternative.

utils/src/main/scala/com/salesforce/op/test/TestCommon.scala
Jauntbox: Ahh, I tried shouldBe < tol (which did not work) - that's better

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
```diff
@@ -151,8 +151,8 @@ object TextTokenizer {
    * @param minTokenLength      minimum token length
    * @return detected language and sentence tokens
    */
-  def tokenize(
-    text: Text,
+  def tokenizeString(
+    textString: String,
```
tovbinm: now this function can explode with NullPointerException is textString is null, while before it could have happened.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
tovbinm: Before this change everything had to be wrapped into `Text` to call the `tokenize`. Then for `String` value would have require additional wrapping into `Option`, which was safer but less efficient. 
I cant find where do you use other overloading for `tokenize` with `String` and `Option[String]`. Is it in this PR?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
tovbinm: Ah, good point on NaN.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
Jauntbox: The `computeTextStats` and `computeTextMapStats` use them. I wanted to avoid having to wrap/unwrap values here since these functions are called per-entry.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
Jauntbox: Hmmm, I'm not sure if it's possible for the `textString` argument to be `null` in practice though. When this function used for tokenizing the map entries, a value that was originally null there will just not show up as an entry in the map. When it's used for tokenizing a normal Text entry, then we should have already safely converted any nulls or missing elements into an `Option[String]`, right?

The actual tokenize call during vectorization is still `tokenize(v.toText)` where `v` is the value in a text map. I'd actually argue that that should be changed to `tokenizeText(v)` to save time converting it to Text and back again.

I agree it's technically less safe, but I don't think it's necessary to have null checking at this point in the flow. We should make sure the data gets created in a safe way, which I think we already do. Are there some specific edge cases I'm missing?

tovbinm: I think the simplest is to add a null check `tokenizeString` and return

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
tovbinm: Makes sense then.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
Jauntbox: Now that I think about it more, I'm pretty sure the old tokenize function would also give a NPE if you fed it a sneaky null value. The `SomeValue.unapply` function explicitly calls `v.isEmpty` which would also fail if v was `null`.

I put back the old tokenize function as `oldTokenize` and tried
```val sneakyStringOpt: Option[String] = null
val myText = Text(sneakyStringOpt)
val res = TextTokenizer.oldTokenize(myText)```
which did indeed throw a NPE.

We have tests all over the place (eg. our vectorizer tests and FeatureTypeSparkConverterTest) that make sure we can handle `null` values in dataframes and safely convert them into our types. I'm not aware of any explicit null checks in our functions elsewhere, so it just feels weird to put one here.

@leahmcguire any opinions on this?

Comment (tovbinm) [reply core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala]: SomeValue.unapply operates on value which is Option[String]. Null check is done during the construction of Text.

Comment (tovbinm) [reply core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala]: Your example you provided is not currently possible and also not a fair one :)

Comment (leahmcguire) [reply core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala]: @Jauntbox is this only called from the Option[String] version below? if so make it private and it is fine.

Comment (leahmcguire) [reply core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala]: In fact make them both private please

File: core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala [hunk 9]
```diff
@@ -282,7 +283,15 @@ object FeatureDistribution {
...
   private def cardinalityValues(values: ProcessedSeq): TextStats = {
-    TextStats(countStringValues(values.left.getOrElse(values.right.get)), Map.empty)
+    implicit val testStatsMonoid: Monoid[TextStats] = TextStats.monoid(RawFeatureFilter.MaxCardinality)
+
+    val stringVals = values match {
+      case Left(stringSeq) => stringSeq
+      case Right(doubleSeq) => doubleSeq.map(_.toString)
+    }
+    stringVals.foldLeft(TextStats.empty)((acc, el) => acc + SmartTextVectorizer.computeTextStats(
+      Option(el), shouldCleanText = false, maxCardinality = RawFeatureFilter.MaxCardinality)
```
TuanNguyen27: Should `shouldCleanText = true` instead ?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
```diff
@@ -164,13 +158,40 @@ class SmartTextVectorizer[T <: Text](uid: String = UID[SmartTextVectorizer[T]])(
...
+    text: T#Value,
+    shouldCleanText: Boolean,
+    maxCardinality: Int
+  ): TextStats = {
+    // Go through each token in text and start appending it to a TextStats instance
+    val lengthsMap: Map[Int, Long] = TextTokenizer.tokenizeStringOpt(text).tokens.value
+      .foldLeft(Map.empty[Int, Long])(
+        (acc, el) => TextStats.additionHelper(acc, Map(el.length -> 1L), maxCardinality)
+      )
+    val (valueCounts, lengthCounts) = text match {
```
TuanNguyen27: when we reach `RawFeatureFilter.MaxCardinality` for `valueCounts`, will `lengthCounts` also stop accumulating ?

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
```diff
@@ -162,30 +162,84 @@ object TextTokenizer {
...
-            tokens.filter(_.length >= minTokenLength).toTextList
-          }
-        TextTokenizerResult(language, sentences)
-      case _ =>
-        TextTokenizerResult(defaultLanguage, Seq(TextList.empty))
+    val sentences = sentenceSplitter.map(_.getSentences(lowerTxt, language))
+      .getOrElse(Seq(lowerTxt))
+      .map { sentence =>
+        val tokens = analyzer.analyze(sentence, language)
+        tokens.filter(_.length >= minTokenLength).toTextList
```
TuanNguyen27: Why are we only keeping tokens with length > `minTokenLength ` ?

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextVectorizer.scala
TuanNguyen27: I think in my earlier attempt to track token length distribution i also ran into this issue...

TuanNguyen27: nvm, this is taken care of by `val newLengthCounts = additionHelper(l.lengthCounts, r.lengthCounts, maxCardinality)` , pls disregard this comment :D

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
Jauntbox: I can change it, but I don't think it matters much here. These values aren't used in SmartTextVectorizer. They're the ones that show up in the ModelInsights.

core/src/main/scala/com/salesforce/op/stages/impl/feature/TextTokenizer.scala
Jauntbox: This was existing behavior. It's a configurable parameter (defaulting to 1), so is not required.

Jauntbox:
Just a heads up on a few more commits - adding a toggle for tokenization in text lengths. Will cause a problems with Chinese/Korean text based on our current tokenizers.

Jauntbox:
Ok - ready for a final look. Sorry for the last-minute refactoring, but realized we needed this toggle exposed for experiments.

Final refactoring:
- Removed tokenizeStringOpt method since we can get by without it (should be more readable too)
- Added toggle to SmartTextVectorizer and SmartTextMapVectorizer to enable/disable tokenization when calculating length distributions in TextStats
- Added tests to check that this toggle does what we want
- Added logging of derived quantities and vectorization method used in SmartTextVectorizer and SmartTextMapVectorizer

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
```diff
@@ -268,21 +271,36 @@ object FeatureDistribution {
...
+    }
+    stringVals.foldLeft(TextStats.empty)((acc, el) => acc + TextStats.textStatsFromString(
+      textString = el,
+      shouldCleanText = shouldCleanText,
+      shouldTokenize = tokenizeForLengths,
+      maxCardinality = RawFeatureFilter.MaxCardinality)
+    )
   }
 
   private def countStringValues[T](seq: Seq[T]): Map[String, Long] = {
```
TuanNguyen27: not relevant to this PR but i think `countStringValues` is no longer used.

core/src/main/scala/com/salesforce/op/filters/PreparedFeatures.scala
```diff
@@ -169,7 +169,8 @@ private[filters] object PreparedFeatures {
       case SomeValue(v: DenseVector) => Map((name, None) -> Right(v.toArray.toSeq))
       case SomeValue(v: SparseVector) => Map((name, None) -> Right(v.indices.map(_.toDouble).toSeq))
       case ft@SomeValue(_) => ft match {
-        case v: Text => Map((name, None) -> Left(v.value.toSeq.flatMap(tokenize)))
+        // case v: Text => Map((name, None) -> Left(v.value.toSeq.flatMap(tokenize)))
```
TuanNguyen27: We are no longer tokenzing text during data prep?

Jauntbox: Whoops, that was for testing - forgot to take it out.

core/src/main/scala/com/salesforce/op/filters/FeatureDistribution.scala
Jauntbox: ok, I can remove it then

TuanNguyen27:
lgtm !

core/src/main/scala/com/salesforce/op/dsl/RichMapFeature.scala
```diff
@@ -322,6 +328,8 @@ trait RichMapFeature {
         .setHashSpaceStrategy(hashSpaceStrategy)
         .setHashAlgorithm(hashAlgorithm)
         .setBinaryFreq(binaryFreq)
+        .setTokenizeForLengths(tokenizeForLengths)
```
leahmcguire: can we make this an enum rather than a boolean? then we have room to expend in the future

PR #467, opened 2020-03-18
Use Spark job grouping to distinguish steps of the machine learning flow

nicodv:
**Related issues**
N/A

Leverages Spark's ability to set a "job group" ID to distinguish certain steps of the machine learning. Examples: data IO, model IO, feature engineering, cross-validation.

`OpSparkListener` is extended to capture which job group is currently active. Also, the new `OpStep` enum's entry names automatically show up in the Spark UI so that the function of stages can be more easily interpreted.

To this end:
- An `OpStep` enum is introduced
- A utility is introduced to set the job group for a code block, or indefinitely

Because a main goal is to get the current step into the real-time `SparkListener` framework, the latter's ability to get hold of the Spark job group was an easy way to accomplish this.
Considered but not feasible:
- Spark listeners per step by adding/removing Spark listeners to/from the listener bus. More complex than using the job grouping, without added benefit.
- Spark's local properties, but it seems they are not shared with Spark listeners

Also considered were the addition of other steps, such as "sanity checker", "scoring" or "metrics". However, these are not included here as this would:
- Needed to be done in places inconsistent with other steps (most of the grouping is contained to `OpWorkflow`/`OpWorkflowModel`)
- Lead to additional overhead/complexity, e.g. because scoring and metrics calculation is part of a single Spark stage and therefore could not be split up by Spark job group.

**Additional context**
The extension to `OpSparkListener` allows for more advanced handling of the metrics that are collected by it, e.g. the metrics can be grouped by `OpStep`.


core/src/main/scala/com/salesforce/op/utils/spark/JobGroupUtil.scala
```diff
@@ -0,0 +1,52 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.salesforce.op.utils.spark
+
+import org.apache.spark.sql.SparkSession
+
+object JobGroupUtil {
```
tovbinm: please add docs to the object and it method arguments

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
```diff
@@ -143,13 +144,14 @@ E <: Estimator[_] with OpPipelineStage2[RealNN, OPVector, Prediction]]
   final override def fit(dataset: Dataset[_]): SelectedModel = {
 
     implicit val spark = dataset.sparkSession
+    JobGroupUtil.setJobGroup(OpStep.CrossValidation)
```
tovbinm: can `setJobGroup` return the previous job group? that way you can reset it back:
```scala
val prevJobGroup = JobGroupUtil.setJobGroup(OpStep.CrossValidation)
// a lot of code
JobGroupUtil.setJobGroup(prevJobGroup)
```

core/src/main/scala/com/salesforce/op/OpWorkflowCore.scala
```diff
@@ -306,7 +307,9 @@ private[op] trait OpWorkflowCore {
   def computeDataUpTo(feature: OPFeature, path: String)
     (implicit spark: SparkSession): Unit = {
     val df = computeDataUpTo(feature)
-    df.saveAvro(path)
+    JobGroupUtil.withJobGroup(OpStep.SavingScores) {
```
tovbinm: These are not necessarily scores, but ok.

core/src/test/scala/com/salesforce/op/utils/spark/JobGroupUtilTest.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.junit.JUnitRunner
+import com.salesforce.op.test.{TestCommon, TestSparkContext}
+
+@RunWith(classOf[JUnitRunner])
+class JobGroupUtilTest extends FlatSpec with TestCommon with TestSparkContext {
+
+  "JobGroupUtil" should "be able to set a job group ID around a code block" in {
```
tovbinm: `Spec[JobGroupUtil]` or `Spec(JobGroupUtil.getClass)`

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
```diff
@@ -143,6 +145,11 @@ class OpSparkListener
 
 }
 
+object OpSparkListener {
+  private val SPARK_JOB_GROUP_ID = "spark.jobGroup.id"
+  val DEFAULT_GROUP_ID: String = "other"
```
tovbinm: Why not add this to `OpStep` enum? e.g. `case object Other(name: String = "other") extends OpStep(name)`

core/src/main/scala/com/salesforce/op/utils/spark/JobGroupUtil.scala
nicodv: done

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
nicodv: It can't, nor is there a 'getJobGroup`.

I could implement it myself, with `JobGroupUtil.setTemporaryJobGroup()`, but this is the only place and cross-validation is kind of a special case anyway.

core/src/main/scala/com/salesforce/op/OpWorkflowCore.scala
nicodv: renamed

core/src/test/scala/com/salesforce/op/utils/spark/JobGroupUtilTest.scala
nicodv: done

utils/src/main/scala/com/salesforce/op/utils/spark/OpSparkListener.scala
nicodv: Makes sense, done. It did force me to move `OpStep` from `core` to the `utils` package.

PR #468, opened 2020-03-26
Avoid having to have an implicit SparkSession in OpWorkFlow(Model)

nicodv:
**Related issues**
N/A

Avoid having to provide a `SparkSession` implicit to `OpWorkflow.loadModel`/`OpWorkflowModel.save` to keep backward compatibility.

N/A

**Additional context**
Small follow-up on https://github.com/salesforce/TransmogrifAI/pull/467

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
@@ -33,10 +33,12 @@ package com.salesforce.op
 import com.salesforce.op.features.FeatureJsonHelper
 import com.salesforce.op.filters.RawFeatureFilterResults
 import com.salesforce.op.stages.{OPStage, OpPipelineStageWriter}
+import com.salesforce.op.utils.spark.{JobGroupUtil, OpStep}
 import enumeratum._
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.io.compress.GzipCodec
 import org.apache.spark.ml.util.MLWriter
+import org.apache.spark.sql.SparkSession
 import org.json4s.JsonAST.{JArray, JObject, JString}
 import org.json4s.JsonDSL._
 import org.json4s.jackson.JsonMethods._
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
```diff
@@ -180,19 +197,6 @@ private[op] object OpWorkflowModelReadWriteShared {
...
 object OpWorkflowModelWriter {
 
-  /**
-   * Save [[OpWorkflowModel]] to path
-   *
-   * @param model     workflow model instance
-   * @param path      path to save the model and its stages
-   * @param overwrite should overwrite the destination
-   */
-  def save(model: OpWorkflowModel, path: String, overwrite: Boolean = true): Unit = {
```
gerashegalov: the scope of this PR did not require it, we should not do breaking changes such as removing public methods.

tovbinm: Oh, I missed that one! good catch @gerashegalov

tovbinm: @nicodv please revert this part of the change.

nicodv: @gerashegalov / @tovbinm So we are OK with adding an implicit SparkSession to `OpWorkflowModel.save()`?

gerashegalov: @nicodv this is not the choice we have to make. We can make it consistent by just changing `saveImpl` in `class OpWorkflowModelWriter`
```scala
  override protected def saveImpl(path: String): Unit = {
    JobGroupUtil.withJobGroup(OpStep.ModelIO) {
      sc.parallelize(Seq(toJsonString(path)), 1)
        .saveAsTextFile(OpWorkflowModelReadWriteShared.jsonPath(path), classOf[GzipCodec])
    }(sparkSession)
  }
```

nicodv: Ah, I overlooked the `saveImpl` method. New PR that reverts this: https://github.com/salesforce/TransmogrifAI/pull/470

PR #471, opened 2020-04-07
Fix Days.daysBetween int overflow

TuanNguyen27:
`Days.daysBetween(datetime_1, datetime_2)` could throw `java.lang.ArithmeticException: Value cannot fit in an int:` if the duration between two dates in milliseconds cannot fit a 32-bit int representation. See related Stack Overflow issue [here](https://stackoverflow.com/questions/25204043/joda-time-calculation-of-seconds-between-two-dates-throws-an-exception)

Switch to using joda's `Duration`. 

N/A 

**Additional context**
N/A

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/feature/OPMapVectorizer.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/feature/OPMapVectorizer.scala
@@ -40,7 +40,7 @@ import org.apache.spark.ml.linalg.Vectors
 import org.apache.spark.ml.param._
 import org.apache.spark.sql.types.MetadataBuilder
 import org.apache.spark.sql.{Dataset, Encoders}
-import org.joda.time.{DateTime, DateTimeZone, Days}
+import org.joda.time.{DateTime, DateTimeZone, Days, Duration}
 
 import scala.reflect.runtime.universe.TypeTag
 
@@ -397,7 +397,7 @@ final class DateMapVectorizerModel[T <: OPMap[Long]] private[op]
   val timeZone: DateTimeZone = DateTimeUtils.DefaultTimeZone
 
   def convertFn: DateMap#Value => RealMap#Value = (dt: DateMap#Value) =>
-    dt.mapValues(v => Days.daysBetween(new DateTime(v, timeZone), referenceDate).getDays.toDouble)
+    dt.mapValues(v => new Duration(new DateTime(v, timeZone), referenceDate).getStandardDays.toDouble)
 }
 
 final class RealMapVectorizerModel[T <: OPMap[Double]] private[op]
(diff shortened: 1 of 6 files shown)


core/src/test/scala/com/salesforce/op/stages/impl/feature/DateMapVectorizerTest.scala
```diff
@@ -122,7 +122,7 @@ class DateMapVectorizerTest extends FlatSpec with TestSparkContext with Attribut
   }
 
   it should "vectorize dates correctly on test date" in {
-    checkAt(new JDateTime(2017, 9, 28, 15, 45, 39, DateTimeUtils.DefaultTimeZone))
+    checkAt(new JDateTime(1901, 1, 1, 0, 0, 0, DateTimeUtils.DefaultTimeZone))
```
tovbinm, line 125: better add a new test case

core/src/main/scala/com/salesforce/op/stages/impl/feature/OPMapVectorizer.scala
```diff
@@ -397,7 +397,7 @@ final class DateMapVectorizerModel[T <: OPMap[Long]] private[op]
   val timeZone: DateTimeZone = DateTimeUtils.DefaultTimeZone
 
   def convertFn: DateMap#Value => RealMap#Value = (dt: DateMap#Value) =>
-    dt.mapValues(v => Days.daysBetween(new DateTime(v, timeZone), referenceDate).getDays.toDouble)
+    dt.mapValues(v => new Duration(new DateTime(v, timeZone), referenceDate).getStandardDays.toDouble)
```
tovbinm, line 400: also, there are more places like this in the code. try grepping.

TuanNguyen27: done!

gerashegalov: there is at least one more in `DateTimeVectorizerTest`

core/src/test/scala/com/salesforce/op/stages/impl/feature/DateMapVectorizerTest.scala
gerashegalov: @TuanNguyen27 please add test cases for ArithmeticExceptions you  are trying to solve

gerashegalov:
please add a real PR description

core/src/test/scala/com/salesforce/op/stages/impl/feature/DateMapVectorizerTest.scala
TuanNguyen27: @gerashegalov the new test case is `checkAt(new JDateTime(1901, 1, 1, 0, 0, 0, DateTimeUtils.DefaultTimeZone))`, which exceed the limit if we use `Days.daysBetween`. Does that work ?

PR #473, opened 2020-04-21
Add categorical detection to be coverage based in addition to unique count based

michaelweilsalesforce:
**Related issues**
Currently SmartTextVectorizer and SmartTextMapVectorizer will count the number of unique entries in a text field (up to a threshold, currently 50) and treat the feature as categorical if it has < 50 unique entries.
You can still run into features that are effectively categorical, but may have a long tail of low-frequency entries. We would get better signal extraction if we treated these as categorical instead of hashing them.

Adding an extra check for Text(Map) features in order to become categoricals. This only applies to features that have a cardinality higher than the threshold and therefore would be hashed.

A better approach to detecting text features that are really categorical would be to use a coverage criteria. For example, the topK entries with minimum support cover at least 90% of the entries, then this would be a good feature to pivot by entry instead of hash by token. The value of 90% can be tuned by the user thanks to a param.

Extra checks need to be passed : 
- Cardinality must be greater than maxCard (already mentioned above).
- Cardinality must also be greater than topK. 
- Finally, the computed coverage of the topK with minimum support must be > 0.

If there is m < TopK elements with the required minimum support, then we are looking at the coverage of these m elements.

I've considered using Algebird Count Min Sketch in order to compute the current `TextStats`. 
However I ran into multiple issue : 
- Lack of transparency: `TopNCMS` only returns the "HeavyHitters" however you need much more than that(e.g. cardinality)  in order to use the coverage method.
- Serialization issues when writing to JSON

A branch still exists : [mw/coverage](https://github.com/salesforce/TransmogrifAI/tree/mw/coverage), but it is in shambles.

**Additional context**
Some criticism regarding `TextStats`. It seems not to be a semi group as it is not associative. Was it intended?

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
@@ -141,11 +141,17 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
     val allFeatureInfo = aggregatedStats.toSeq.map { textMapStats =>
       textMapStats.keyValueCounts.toSeq.map { case (k, textStats) =>
         val totalCount = textStats.valueCounts.values.sum
-        val top100 = textStats.valueCounts.toSeq.sortBy(- _._2).take(100).toMap
+        val sorted = textStats.valueCounts.toSeq.sortBy(- _._2)
+        val sortedValues = sorted.map(_._2)
+        val cumSum = sortedValues.tail.scanLeft(sortedValues.head)(_ + _)
+        println(sorted)
+        println(sorted.map(_._1).zip(cumSum))
+        println(cumSum.filter(_ <= 0.9 * totalCount).last)
 
         val vecMethod: TextVectorizationMethod = textStats match {
-          case _ if top100.size < 100 && textStats.valueCounts.size <= maxCard => TextVectorizationMethod.Pivot
-          case _ if top100.size >= 100 && top100.values.sum * 1.0 / totalCount >= 0.9 => TextVectorizationMethod.Pivot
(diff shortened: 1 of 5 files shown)


leahmcguire:
So is this a WIP or something? :-P

michaelweilsalesforce:
@leahmcguire @Jauntbox No longer WIPWIPWIPWIPWIPWIPWIPWIPWIP

core/src/main/scala/com/salesforce/op/stages/impl/feature/SmartTextMapVectorizer.scala
```diff
@@ -138,9 +137,23 @@ class SmartTextMapVectorizer[T <: OPMap[String]]
...
+        val filteredStats = textStats.valueCounts.filter { case (_, count) => count >= $(minSupport) }
+        val sorted = filteredStats.toSeq.sortBy(- _._2)
+        val sortedValues = sorted.map(_._2)
+        // Cumulative Count
+        val cumCount = sortedValues.headOption.map(_ => sortedValues.tail.scanLeft(sortedValues.head)(_ + _))
+          .getOrElse(Seq.empty)
+        val coverage = cumCount.lift(math.min(topKValue, cumCount.length) - 1).getOrElse(0L) * 1.0 / totalCount
         val vecMethod: TextVectorizationMethod = textStats match {
+          // If cardinality not respect, but coverage is, then pivot the feature
+          case _ if textStats.valueCounts.size > maxCard && textStats.valueCounts.size > topKValue && coverage > 0 &&
```
leahmcguire, line 155: Even with the comment I am having trouble parsing which conditions you are looking for - can you add a but more?

leahmcguire: can you just change it to``` coverage > $(coveragePct)``` and get rid of the additional  ```coverage > 0``` check?

PR #474, opened 2020-05-04
ignoring cardinality in modelinsights serialization

winterslu:
Related issues
Tokens of text field are printing out while internal usage.

Describe the proposed solution
Ignore cardinality field in modelInsights serialization, addition to featureDistribution

Additional context
Previous PR #420 & #438 & #447

### DIFF
--- a/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
+++ b/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
@@ -970,5 +970,4 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
     )
   }
 
-  it should ""
 }
--- a/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
+++ b/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
@@ -456,6 +456,7 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
             i.featureName shouldEqual o.featureName
             i.featureType shouldEqual o.featureType
             i.derivedFeatures.zip(o.derivedFeatures).foreach { case (ii, io) => ii.corr shouldEqual io.corr }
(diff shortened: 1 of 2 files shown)


core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -968,4 +969,6 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
       "second" -> classOf[SingleMetric]
     )
   }
+
+  it should ""
```
nicodv, line 973: Should it?

WintersLu:
@gerashegalov this is fixing the final missing piece of very old bug, need to get this merged so we can move forward

leahmcguire (approved):

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
```diff
@@ -456,6 +456,7 @@ class ModelInsightsTest extends FlatSpec with PassengerSparkFixtureTest with Dou
             i.featureName shouldEqual o.featureName
             i.featureType shouldEqual o.featureType
             i.derivedFeatures.zip(o.derivedFeatures).foreach { case (ii, io) => ii.corr shouldEqual io.corr }
+            o.distributions.foreach { o => o.cardEstimate shouldEqual None}
```
gerashegalov, line 459: the test could be more obvious asserting that the pre-serialization object has cardEstimate. And that it loses cardEstimates per requirement after serialization. As opposed to just checking the post-serialization state. 

However, I verified that the test fails without your main change, so we should be ok to merge it this way.

gerashegalov (approved):
LGTM

core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala
WintersLu: @gerashegalov I updated the counter part of this test case

Scutua (approved):

PR #475, opened 2020-05-05
Write and read Spark stages to/from MLeap instead of Spark classes

leahmcguire:
**Related issues**
Currently, Spark save method is used to serialize and deserialize Spark wrapped stages. This PR changes the underlying serialization to write and read from MLeap bundles.

Writes to MLeap and reads from MLeap with fallback to trying to read from Spark save.

N/A

**Additional context**
Next steps will be PR's to read the stages directly with the MLeap context rather than the Spark context for local scoring (and possibly all scoring - to better optimize the DAG)


leahmcguire:
@TuanNguyen27 the test that you put in that should have failed on the local XGboost is (correctly) failing in this PR.

core/src/main/scala/com/salesforce/op/stages/sparkwrappers/specific/OpPredictorWrapper.scala
```diff
@@ -98,12 +100,19 @@ class OpPredictorWrapper[E <: Predictor[Vector, E, M], M <: PredictionModel[Vect
...
 
-    SparkModelConverter.toOP(model, uid)
+    val wrappedModel = SparkModelConverter.toOP(model, uid)
       .setParent(this)
       .setInput(in1.asFeatureLike[RealNN], in2.asFeatureLike[OPVector])
       .setMetadata(getMetadata())
       .setOutputFeatureName(getOutputFeatureName)
+
+    if (model.isInstanceOf[XGBoostClassificationModel] || model.isInstanceOf[XGBoostRegressionModel]) {
+      wrappedModel.setOutputDF(model.transform(dataset.limit(1)))
```
TuanNguyen27, line 110: just curious why do we have `.limit(1)` here ?

leahmcguire: we only need one example for the xgboost mleap save (it has a step that calls .first() to get the vector size)

core/src/main/scala/com/salesforce/op/stages/sparkwrappers/specific/OpPredictorWrapper.scala
```diff
@@ -119,4 +131,11 @@ abstract class OpPredictorWrapperModel[M <: PredictionModel[Vector, M]]
 ) extends Model[OpPredictorWrapperModel[M]] with SparkWrapperParams[M]
   with OpTransformer2[RealNN, OPVector, Prediction] {
   setDefault(sparkMlStage, Option(sparkModel))
+
+  protected def getSparkOrLocalMethod(sparkMethodName: String, localMethodName: String,
```
tovbinm, line 135: Let's add two helper methods and use them instead:

```scala
def getRawToProbabilityMethod(): MethodMirror = getSparkOrLocalMethod("raw2probability", "rawToProbability")

def getProbabilityToPredictionMethod(): MethodMirror = getSparkOrLocalMethod("probability2prediction", "probabilityToPrediction")
```

core/src/main/scala/com/salesforce/op/stages/impl/regression/OpRandomForestRegressor.scala
```diff
@@ -125,4 +127,9 @@ class OpRandomForestRegressionModel
   ttov: TypeTag[Prediction#Value]
 ) extends OpPredictionModel[RandomForestRegressionModel](
   sparkModel = sparkModel, uid = uid, operationName = operationName
-)
+){
+  @transient lazy protected val predict: Vector => Double = getSparkMlStage().map(s => s.predict(_))
```
tovbinm, line 131: This also seems to be a very repetitive pattern. We can add a helper method for it as well.

core/src/main/scala/com/salesforce/op/stages/sparkwrappers/specific/OpPredictorWrapper.scala
tovbinm: Let's add a comment for it. Looks like this is the only such exception so far.

core/src/main/scala/com/salesforce/op/stages/sparkwrappers/specific/OpPredictorWrapper.scala
```diff
@@ -98,15 +103,22 @@ class OpPredictorWrapper[E <: Predictor[Vector, E, M], M <: PredictionModel[Vect
...
+      wrappedModel.setOutputDF(model.transform(dataset.limit(1)))
+    }
+
+    wrappedModel
   }
+
 }
 
-abstract class OpPredictorWrapperModel[M <: PredictionModel[Vector, M]]
+abstract class OpPredictorWrapperModel[M <: PredictionModel[Vector, M] : ClassTag]
```
tovbinm, line 121: If you need the `ClassTag` here then please add it to the implicit arguments list and name it, e.g. `implicit ctag: ClassTag`.

core/src/main/scala/com/salesforce/op/stages/sparkwrappers/specific/OpProbabilisticClassifierModel.scala
```diff
@@ -45,7 +46,7 @@ import scala.reflect.runtime.universe._
  * @param operationName unique name of the operation this stage performs
  * @tparam T type of the model to wrap
  */
-abstract class OpProbabilisticClassifierModel[T <: ProbabilisticClassificationModel[Vector, T]]
+abstract class OpProbabilisticClassifierModel[T <: ProbabilisticClassificationModel[Vector, T] : ClassTag]
```
tovbinm, line 49: same here - add `ctag: ClassTag` to the implicit arguments list.

core/src/main/scala/com/salesforce/op/utils/stages/FitStagesUtil.scala
```diff
@@ -228,7 +226,7 @@ private[op] case object FitStagesUtil {
           train = currTrain,
           test = currTest,
           hasTest = hasTest,
-          transformData = indexOfLastEstimator.exists(_ < index), // only need to update for fit before last estimator
+          transformData = true, // even transformers need to be fit because may need metadata from training
```
tovbinm, line 229: 👍

features/build.gradle
```diff
@@ -17,4 +17,11 @@ dependencies {
 
     // Json4s extensions (needed for Joda time)
     compile "org.json4s:json4s-ext_$scalaVersion:$json4sVersion"
+
+    // MLeap serialization & runtime for Spark models
+    compile "ml.combust.mleap:mleap-spark_$scalaVersion:$mleapVersion"
+    compile "ml.combust.mleap:mleap-runtime_$scalaVersion:$mleapVersion"
+    compile "ml.combust.mleap:mleap-xgboost-spark_$scalaVersion:$mleapVersion"
+    compile "ml.combust.mleap:mleap-xgboost-runtime_$scalaVersion:$mleapVersion"
+    // compile "com.jsuereth:scala-arm_$scalaVersion:$mleapVersion"
```
tovbinm, line 26: What is this `com.jsuereth:scala-arm_`? Please remove if not in use.

features/src/main/scala/com/salesforce/op/stages/SparkStageParam.scala
```diff
@@ -55,9 +66,11 @@ class SparkStageParam[S <: PipelineStage with Params]
    * Spark stage saving path
    */
   var savePath: Option[String] = None
+  @transient var sbc: Option[SparkBundleContext] = None
```
tovbinm, line 69: should `sbc` and `localTransformer` be private and instead exposed over a getter?

local/src/main/scala/com/salesforce/op/local/OpWorkflowModelLocal.scala
```diff
@@ -85,29 +74,26 @@ trait OpWorkflowModelLocal extends Serializable {
...
       val opStages = stagesWithIndex.collect { case (s: OpTransformer, i) => OPModel(s.getOutputFeatureName, s) -> i }
 
       // Collect all Spark wrapped stages
-      val sparkStages = stagesWithIndex.filterNot(_._1.isInstanceOf[OpTransformer]).collect {
-        case (opStage: OPStage with SparkWrapperParams[_], i) if opStage.getSparkMlStage().isDefined =>
-          val sparkStage = opStage.getSparkMlStage().get.asInstanceOf[Transformer]
-          (opStage, sparkStage, i)
+      val mleapStages = stagesWithIndex.filterNot(_._1.isInstanceOf[OpTransformer]).collect {
+        case (opStage: OPStage with SparkWrapperParams[_], i) if opStage.getLocalMlStage().isDefined =>
+          val model = opStage.getLocalMlStage().get
```
tovbinm, line 90: Better pattern match and error gracefully when local model is missing. For example:
```scala
opStage.getLocalMlStage() match {
  case None => throw new RuntimeException("Local model not found for stage ${opStage.uid} of type ${opStage.getClass}")
  case Some(model) =>
     // Apply model 
}


```

features/src/main/scala/com/salesforce/op/stages/SparkStageParam.scala
leahmcguire: I am following the convention of savePath. This is the parameter itself - there are getters and setters for the path and transformers on the spark wrapped stage itself but not on the parameter...

core/src/main/scala/com/salesforce/op/stages/impl/regression/OpRandomForestRegressor.scala
leahmcguire: so the problem with a helper function in this one is there is no shared class for the mleap regressors that contains the predict function. they all implement it but you have to cast them to their specific type to get the predict. Thus a helper function only saves one map and makes it hard to read. I suppose I could use reflection in a shared helper...do you think that is better?

tovbinm:
🥳 🥳 🥳

koertkuipers:
this seems to have broken some of our inhouse unit tests. in some cases it was because we wrote to relative paths i think. those were easily fixed by making paths absolute. in other situations the paths were absolute and i am unsure why it broke at this point...

stacktraces all have to do with mleap BundleFile on reading and writing. always the same NPE in `UnixPath.normalizeAndCheck`. for example:

```
[info]   Cause: java.lang.NullPointerException:
[info]   at sun.nio.fs.UnixPath.normalizeAndCheck(UnixPath.java:77)
[info]   at sun.nio.fs.UnixPath.<init>(UnixPath.java:71)
[info]   at sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:281)
[info]   at ml.combust.bundle.BundleFile$.apply(BundleFile.scala:59)
[info]   at ml.combust.bundle.BundleFile$.apply(BundleFile.scala:40)
[info]   at com.salesforce.op.stages.SparkStageParam.$anonfun$jsonDecodeMleap$1(SparkStageParam.scala:164)
[info]   at resource.DefaultManagedResource.open(AbstractManagedResource.scala:110)
[info]   at resource.AbstractManagedResource.acquireFor(AbstractManagedResource.scala:87)
[info]   at resource.DeferredExtractableManagedResource.either(AbstractManagedResource.scala:29)
[info]   at resource.DeferredExtractableManagedResource.opt(AbstractManagedResource.scala:31)
[info]   at com.salesforce.op.stages.SparkStageParam.jsonDecodeMleap(SparkStageParam.scala:173)
[info]   at com.salesforce.op.stages.SparkStageParam.jsonDecode(SparkStageParam.scala:123)
[info]   at com.salesforce.op.stages.SparkStageParam.jsonDecode(SparkStageParam.scala:55)
[info]   at org.apache.spark.ml.util.DefaultParamsReader$Metadata.$anonfun$setParams$1(ReadWrite.scala:564)
[info]   at scala.collection.immutable.List.foreach(List.scala:392)
[info]   at org.apache.spark.ml.util.DefaultParamsReader$Metadata.setParams(ReadWrite.scala:561)
[info]   at org.apache.spark.ml.util.DefaultParamsReader$Metadata.getAndSetParams(ReadWrite.scala:549)
[info]   at org.apache.spark.ml.SparkDefaultParamsReadWrite$.getAndSetParams(SparkDefaultParamsReadWrite.scala:126)
```

koertkuipers:
is protobuf 3 going to be an issue on spark/hadoop?

tovbinm:
@koertkuipers Can you please open an issue to track this? Can you also share which transformer / estimator are you using in your workflow?

koertkuipers:
https://github.com/salesforce/TransmogrifAI/issues/514

PR #476, opened 2020-05-11
Remove duplicate features using sanity checker feature to feature correlations

leahmcguire:
**Related issues**

When features are exact duplicates of each other we would like to remove one before modeling

**Additional context**


core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
```diff
@@ -319,29 +323,45 @@ case class Correlations
...
   private def correlationsFromMetadata(meta: Metadata): Correlations = {
     val wrapped = meta.wrapped
-    Correlations(
-      featuresIn = wrapped.getArray[String](SanityCheckerNames.FeaturesIn).toSeq,
-      values = wrapped.getArray[Double](SanityCheckerNames.Values).toSeq,
-      nanCorrs = wrapped.getArray[String](SanityCheckerNames.CorrelationsWLabelIsNaN).toSeq,
-      corrType = CorrelationType.withNameInsensitive(wrapped.get[String](SanityCheckerNames.CorrelationType))
-    )
+    val features = wrapped.getArray[String](SanityCheckerNames.FeaturesIn).toSeq
+    if (wrapped.underlyingMap.keySet.contains("values")) { // old sanity checker meta
```
nicodv, line 345: nice backwards compatibility hack! maybe use "correlationsWithLabel" instead of "values", as its speaks more to the type of change that occurred?

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
```diff
@@ -274,23 +273,24 @@ case class CategoricalStats
...
 (
   featuresIn: Seq[String],
-  values: Seq[Double],
-  nanCorrs: Seq[String],
+  valuesWithLabel: Seq[Double],
+  valuesWithFeatures: Seq[Seq[Double]],
   corrType: CorrelationType
 ) extends MetadataLike {
-  require(featuresIn.length == values.length, "Feature names and correlation values arrays must have the same length")
+  require(featuresIn.length == valuesWithLabel.length,
```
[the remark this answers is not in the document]
leahmcguire: correlation matrix may be empty so not sure it makes sense

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
```diff
@@ -94,15 +94,24 @@ trait SanityCheckerParams extends DerivedFeatureFilterParams {
   final val maxCorrelation = new DoubleParam(
     parent = this, name = "maxCorrelation",
     doc = "Maximum correlation (absolute value) allowed between a feature in the feature vector and the label",
-    isValid = ParamValidators.inRange(lowerBound = 0.0, upperBound = 1.0, lowerInclusive = true, upperInclusive = true)
+    isValid = ParamValidators.inRange(lowerBound = -0.1, upperBound = 1.1, lowerInclusive = true, upperInclusive = true)
```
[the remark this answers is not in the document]
leahmcguire: Floating point - I found in some tests that if you just have a check that the corr be > 1.0 doubles mean that some perfectly correlated features are excluded because they have a corr of 1.00000000004 an I wanted to be able to turn the check off effectively without adding another param

core/src/main/scala/com/salesforce/op/stages/impl/preparators/DerivedFeatureFilterUtils.scala
```diff
@@ -174,6 +176,10 @@ object DerivedFeatureFilterUtils {
             case ind => Option(corrsWithLabel(ind))
           },
           cramersV = cramersVMap.get(name),
+          featureCorrs = corrIndices.indexOf(i) match {
```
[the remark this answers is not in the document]
leahmcguire: So this line is part of creating the exclusion criteria by row it cannot go outside the iteration over i

nicodv:
LGTM

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
Jauntbox, line 97: Why did these ranges need to be changed. We shouldn't get correlations outside of [0, 1], right?

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
```diff
@@ -94,15 +94,24 @@ trait SanityCheckerParams extends DerivedFeatureFilterParams {
...
+  )
+  def setMaxFeatureCorr(value: Double): this.type = set(maxFeatureCorr, value)
+  def getMaxFeatureCorr: Double = $(maxFeatureCorr)
+
   final val minCorrelation = new DoubleParam(
     parent = this, name = "minCorrelation",
     doc = "Minimum correlation (absolute value) allowed between a feature in the feature vector and the label",
-    isValid = ParamValidators.inRange(lowerBound = 0.0, upperBound = 1.0, lowerInclusive = true, upperInclusive = true)
+    isValid = ParamValidators.inRange(lowerBound = -0.1, upperBound = 1.1, lowerInclusive = true, upperInclusive = true)
```
Jauntbox, line 114: Same here - why are these changed?

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
```diff
@@ -92,9 +92,8 @@ case class SanityCheckerSummary
     sample: Double
   ) {
     this(
-      correlationsWLabel = new Correlations(
-        stats.filter(s => s.corrLabel.isDefined && !s.corrLabel.get.isNaN).map(s => s.name -> s.corrLabel.get),
-        stats.filter(s => s.corrLabel.isDefined && s.corrLabel.get.isNaN).map(_.name),
+      correlations = new Correlations(
+        stats.filter(s => s.corrLabel.isDefined).map(s => (s.name, s.corrLabel.get, s.featureCorrs)),
```
Jauntbox, line 96: Do we not have to check for NaNs anymore? What if the feature is constant and we turn off the min variance threshold?

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
```diff
@@ -274,23 +271,24 @@ case class CategoricalStats
  * Correlations between features and the label from [[SanityChecker]]
  *
  * @param featuresIn names of features
- * @param values     correlation of feature with label
- * @param nanCorrs   nan correlation features
+ * @param valuesWithLabel     correlation of feature with label
+ * @param valuesWithFeatures   correlations between features features
```
Jauntbox, line 275: "correlations between features ~~features~~"

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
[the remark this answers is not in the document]
nicodv: @Jauntbox ^^^

core/src/test/scala/com/salesforce/op/stages/impl/preparators/BadFeatureZooTest.scala
```diff
@@ -88,6 +88,7 @@ class BadFeatureZooTest extends FlatSpec with TestSparkContext with Logging {
     val genFeatureVector = Seq(rawCity, rawCountry, rawPickList, rawCurrency).transmogrify()
     val checkedFeatures = new SanityChecker()
       .setCheckSample(1.0)
+      .setMaxFeatureCorr(1.1)
```
Jauntbox, line 91: Oh, I see - you're using values > 1.0 to turn it off. I think it may be clearer if we keep the bounds in [0, 1.0], but make it a < check instead of a <= check (so that setting it to 1.0 would turn it off).

leahmcguire: So I did that! Turns out that with doubles you can get a correlation > 1.0 with perfectly correlated things :-P

Eg in a couple of test 1 had it set to 1.0 with corr > and it failed because the correlation was 1.00000000004

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerMetadata.scala
leahmcguire: So I fixed it - in that when it goes to metadata it turns it into strings and then when it gets out back into the class it is turned into doubles. Now nans work! The json serializer would not have helped because it was a problem with the spark metadata.

core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala
```diff
@@ -94,15 +94,24 @@ trait SanityCheckerParams extends DerivedFeatureFilterParams {
   final val maxCorrelation = new DoubleParam(
     parent = this, name = "maxCorrelation",
     doc = "Maximum correlation (absolute value) allowed between a feature in the feature vector and the label",
-    isValid = ParamValidators.inRange(lowerBound = 0.0, upperBound = 1.0, lowerInclusive = true, upperInclusive = true)
+    isValid = ParamValidators.inRange(lowerBound = -0.1, upperBound = 1.1, lowerInclusive = true, upperInclusive = true)
   )
   def setMaxCorrelation(value: Double): this.type = set(maxCorrelation, value)
   def getMaxCorrelation: Double = $(maxCorrelation)
 
+  final val maxFeatureCorr = new DoubleParam(
```
gerashegalov, line 102: better spell out Correlation

PR #484, opened 2020-06-17
Expose derived feature value in model insights

TuanNguyen27:
Some of the derived features (or engineered features) have human-readable values that could be exposed inside modelInsights for downstream consumption.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/ModelInsights.scala
+++ b/core/src/main/scala/com/salesforce/op/ModelInsights.scala
@@ -636,7 +636,10 @@ case object ModelInsights {
               derivedFeatureName = h.columnName,
               stagesApplied = h.parentFeatureStages,
               derivedFeatureGroup = h.grouping,
-              derivedFeatureValue = h.indicatorValue,
+              derivedFeatureValue = (h.indicatorValue, h.descriptorValue) match {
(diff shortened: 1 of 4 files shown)


TuanNguyen27:
> LGTM, are there other places where the descriptor value should be passed on?

I did a spot check, i think only `OpMapVec` was missing this.

TuanNguyen27:
> Yes, what happened with the OpMapVec? Did we miss adding descriptorValue?
> Also, maybe add one test in ModelInsight when a descriptor value is being added.

There's [this line](https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala#L648) and [this line](https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/ModelInsightsTest.scala#L665) that check for non-empty `derivedFeatureValue`. However we don't have a test for `derivedFeatureValue` when it is derived from a Map feature. Is it okay if i just add a check for `derivedFeatureValue` when it comes from a Map feature ?

TuanNguyen27:
@leahmcguire @michaelweilsalesforce It's actually tricky to write a test for this change, because the existing tests drop `numericMap` as a feature...Do you have any suggestions ?

Issue #483, opened 2020-06-12
Make AvroInOut#selectExistingPaths more robust and readable


gerashegalov:
**Describe the bug**
The logic for instantiating the filesystem object is [unnecessarily complex](https://github.com/gerashegalov/TransmogrifAI/blob/04cd7ff35d2ecab46180cf9689a7a53148957319/utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala#L90). `selectExistingPaths` relies on the side effect of Path creation to detect that no paths were passed to begin with but after actually creating the filesytem object on the empty path.

**Expected behavior**
Convert array of strings to array of Hadooo Path objects right away. This will already bail on the empty paths. Obtaining FS is just `val fs = firstFile.getFileSystem(sc.hadoopConfiguration)`. Then filter as is.

sakhuja:
@gerashegalov I can take a look at this. Thanks,

gerashegalov:
resolved by #486

PR #486, opened 2020-06-21
Simplified selectExistingPaths

sakhuja:
1. selectExistingPaths method simplification. 
2. handling for order free path validity in the path argument.

**Related issues**
[Issues]https://github.com/salesforce/TransmogrifAI/issues/483

In selectExistingPaths method, we check for the component paths validity using FileSystem object's read outcome. Also, make use of scala's Try method to bucketize valid and invalid paths.
as part of this change we also make sure that the order of the component paths is not critical to the success of the selectExistingPaths method, specially when the first path component is invalid. Added a test as well to verify that we can handle order independent paths.

### DIFF
--- a/utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
+++ b/utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
@@ -30,8 +30,6 @@
 
 package com.salesforce.op.utils.io.avro
 
-import java.net.URI
-
 import com.salesforce.op.utils.spark.RichRDD._
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
@@ -45,7 +43,7 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.SparkSession
 
 import scala.reflect.ClassTag
-import scala.util.{Failure, Success, Try}
+import scala.util.Try
 
 
 /**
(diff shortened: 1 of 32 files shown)


sakhuja:
@gerashegalov This change is ready for review. 

please note: I've added my salesforce email id to my account to confirm with salesforce-cla requirement.

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
```diff
@@ -86,18 +86,26 @@ object AvroInOut {
...
-      val separator = FileSystems.getDefault.getSeparator
-      val firstPath = paths.head.split(separator)
-      val bucket = firstPath.slice(0, math.min(4, firstPath.length)).mkString(separator)
-      FileSystem.get(new URI(bucket), sc.sparkContext.hadoopConfiguration)
-    } catch {
-      case ex: Exception => throw new IllegalArgumentException(s"Bad path '$path': ${ex.getMessage}")
+    val paths = path.split(',')
+    val fsSeq = paths.map(path => Try(FileSystem.get(new URI(path), sc.sparkContext.hadoopConfiguration)))
+    val validFS = fsSeq.filter(p => p.isSuccess)
+    if (validFS.nonEmpty) {
```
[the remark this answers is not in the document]
gerashegalov: I was hoping to keep the code short to the tune of 
```scala
    val paths = Option(path).map(_.split(',')).getOrElse(Array.empty).map(new Path(_))
    val found = paths.headOption.flatMap(firstPath => 
      Try(firstPath.getFileSystem(sc.sparkContext.hadoopConfiguration)).recover { case t => 
        throw new IllegalArgumentException(s"Bad path '$path': ${t.getMessage}", t)
      }.toOption)
      .map(fs => paths.filter(fs.exists)).getOrElse(Array.empty)
```

utils/src/main/scala/com/salesforce/op/utils/io/avro/AvroInOut.scala
```diff
@@ -86,17 +84,13 @@ object AvroInOut {
...
-    }
-    val found = paths.filter(p => fs.exists(new Path(p)))
-    if (found.isEmpty) throw new IllegalArgumentException(s"No valid directory found in path '$path'")
+    val paths = Option(path).map(_.split(',')).getOrElse(Array.empty).map(new Path(_))
+    val found = paths.headOption.flatMap(firstPath =>
+      Try(firstPath.getFileSystem(sc.sparkContext.hadoopConfiguration)).recover { case t =>
+        throw new IllegalArgumentException(s"Bad path '$path': ${t.getMessage}", t)
+      }.toOption)
+      .map(fs => paths.filter(fs.exists)).getOrElse(Array.empty)
+    if(found.isEmpty) throw new IllegalArgumentException(s"No valid directory found in path '$path'")
```
gerashegalov, line 93: nit: whitespace disappeared in `if (found`

utils/src/test/scala/com/salesforce/op/utils/io/avro/AvroInOutTest.scala
```diff
@@ -109,13 +109,22 @@ class AvroInOutTest extends FlatSpec with TestSparkContext {
     f3.delete()
     assume(f1.exists && !f2.exists && !f3.exists)
 
+    // check for one dir being invalid in the path amongst two
```
gerashegalov, line 112: thanks for comments

sakhuja:
Thanks for the review @gerashegalov. Could you help merge? I don't have write-access.

PR #487, opened 2020-06-25
Replace empty XGBoost feature importance with a vector of zeros

TuanNguyen27:
**Related issues**
A successfully trained XGBoost model could return an empty feature importance vector when the features have zero signal w.r.t the label. This behavior will fail `getModelContributions` via [this line](https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostParams.scala#L94). 

```
require(featureScore.nonEmpty, "Feature score map is empty")
```
In this case, we will return a feature contribution vector of 0's so that it matches the behavior of other models.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/ModelInsights.scala
+++ b/core/src/main/scala/com/salesforce/op/ModelInsights.scala
@@ -802,12 +802,18 @@ case object ModelInsights {
       case m: XGBoostRegressionModel =>
         Try(Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)) match {
           case Success(contrib) => contrib
-          case _ => Seq.empty
+          case _ => featureVectorSize match {
+            case Some(n) => Seq(Seq.fill(n)(0.0))
+            case _ => Seq(Seq.empty)
+          }
         }
       case m: XGBoostClassificationModel =>
         Try(Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)) match {
           case Success(contrib) => contrib
-          case _ => Seq.empty
+          case _ => featureVectorSize match {
+            case Some(n) => Seq(Seq.fill(n)(0.0))
+            case _ => Seq(Seq.empty)
+          }
         }
     }
     contributions.getOrElse(Seq.empty)
--- a/core/src/main/scala/com/salesforce/op/ModelInsights.scala
+++ b/core/src/main/scala/com/salesforce/op/ModelInsights.scala
@@ -803,7 +803,7 @@ case object ModelInsights {
         Try(Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)) match {
           case Success(contrib) => contrib
           case _ => featureVectorSize match {
-            case Some(n) => Seq(Seq.fill(n)(0.0))
+            case Some(n) => Seq(Array.ofDim[Double](n).toIndexedSeq)
             case _ => Seq(Seq.empty)
           }
         }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -799,8 +799,16 @@ case object ModelInsights {
...
       case m: GeneralizedLinearRegressionModel => Seq(m.coefficients.toArray.toSeq)
-      case m: XGBoostRegressionModel => Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)
-      case m: XGBoostClassificationModel => Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)
+      case m: XGBoostRegressionModel =>
+        Try(Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)) match {
+          case Success(contrib) => contrib
+          case _ => Seq.empty
+        }
+      case m: XGBoostClassificationModel =>
+        Try(Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)) match {
```
leahmcguire, line 808: Why does this need to be wrapped in try? Has this changed location or name? Please track down what is failing rather than allowing it to silently fail

TuanNguyen27: No, there's no change in location or name. However, there's something flaky about extracting model importance from a trained XGBoost model. In my downstream integration test, `nativeBooster.getFeatureScoreVector` occasionally failed due to `require(featureScore.nonEmpty, "Feature score map is empty")`. I can't reproduce this error in my experiments on real datasets (they don't fail at all and modelInsights was extracted just fine) so i'm not sure if it's a real issue...

Jauntbox: Yeah, I'm not sure we'd want to allow this. I think requiring feature importances is a good idea, and it would be better to understand why this is happening. Is something going wrong with the model training entirely, and this is the first symptom of a larger problem?

I'd start with adding some logging to the failing test and see if that helps pinpoint the issue.

TuanNguyen27: @Jauntbox @leahmcguire turns out if there's no useful feature, XGBoost would return an empty vector here, would that justify having a `Try` ?

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -799,8 +799,22 @@ case object ModelInsights {
       case m: RandomForestRegressionModel => Seq(m.featureImportances.toArray.toSeq)
       case m: GBTRegressionModel => Seq(m.featureImportances.toArray.toSeq)
       case m: GeneralizedLinearRegressionModel => Seq(m.coefficients.toArray.toSeq)
-      case m: XGBoostRegressionModel => Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)
-      case m: XGBoostClassificationModel => Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)
+      case m: XGBoostRegressionModel =>
+        Try(Seq(m.nativeBooster.getFeatureScoreVector(featureVectorSize).toArray.toSeq)) match {
+          case Success(contrib) => contrib
+          case _ => featureVectorSize match {
```
Jauntbox, line 805: Can anyone comment on why `featureVectorSize` is an Option in the first place? I couldn't find a place where this function is ever called without it being explicitly defined.

You may be able to just use `Seq(Seq.fill(n)(0.0))`, if so.

TuanNguyen27: So i can drop `case _ => Seq(Seq.empty)` with no consequences ? :D

core/src/main/scala/com/salesforce/op/ModelInsights.scala
```diff
@@ -799,8 +799,22 @@ case object ModelInsights {
...
+            case Some(n) => Seq(Seq.fill(n)(0.0))
```
tovbinm, line 806: Better do `Array.ofDim[Int](100).toIndexedSeq`

tovbinm: @TuanNguyen27 ^ FYI

PR #489, opened 2020-07-08
Train a XGBoost model and use it to score to test MLeap compatability

TuanNguyen27:
Refactor & add a test in OpWorkflowModelLocalTest.scala to make sure that we can score new records with a XGBoost classifier.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
@@ -126,6 +126,10 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
         .addGrid(xgb.eta, DefaultSelectorParams.Eta)
         .addGrid(xgb.maxDepth, DefaultSelectorParams.MaxDepthXGBClassifier)
         .addGrid(xgb.minChildWeight, DefaultSelectorParams.MinChildWeight)
+        .addGrid(xgb.missing, DefaultSelectorParams.MissingValPad)
+        .addGrid(xgb.maximizeEvaluationMetrics, Array(true))
+        .addGrid(xgb.evalMetric, Array("aucpr"))
+        .addGrid(xgb.objective, Array("binary:logistic"))
 
       Seq(lr -> lrParams, rf -> rfParams, gbt -> gbtParams, svc -> svcParams,
         nb -> nbParams, dt -> dtParams, xgb -> xgbParams)
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
@@ -60,7 +60,8 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
      * off by default
      */
     val modelTypesToUse: Seq[BinaryClassificationModelsToTry] = Seq(
-      MTT.OpLogisticRegression, MTT.OpRandomForestClassifier, MTT.OpGBTClassifier
+      MTT.OpLogisticRegression, MTT.OpRandomForestClassifier,
+      MTT.OpGBTClassifier, MTT.OpXGBoostClassifier
     )
 
     /**
@@ -123,7 +124,7 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
       val xgbParams = new ParamGridBuilder()
         .addGrid(xgb.numRound, DefaultSelectorParams.NumRound)
         .addGrid(xgb.eta, DefaultSelectorParams.Eta)
-        .addGrid(xgb.maxDepth, DefaultSelectorParams.MaxDepth)
+        .addGrid(xgb.maxDepth, DefaultSelectorParams.MaxDepthXGBClassifier)
         .addGrid(xgb.minChildWeight, DefaultSelectorParams.MinChildWeight)
 
       Seq(lr -> lrParams, rf -> rfParams, gbt -> gbtParams, svc -> svcParams,
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
@@ -123,13 +123,15 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
       val xgb = new OpXGBoostClassifier()
       val xgbParams = new ParamGridBuilder()
         .addGrid(xgb.numRound, DefaultSelectorParams.NumRound)
+        .addGrid(xgb.numEarlyStoppingRounds, DefaultSelectorParams.EarlyStopping)
         .addGrid(xgb.eta, DefaultSelectorParams.Eta)
+        .addGrid(xgb.gamma, DefaultSelectorParams.BinaryClassXGBGamma)
         .addGrid(xgb.maxDepth, DefaultSelectorParams.MaxDepthXGBClassifier)
         .addGrid(xgb.minChildWeight, DefaultSelectorParams.MinChildWeight)
         .addGrid(xgb.missing, DefaultSelectorParams.MissingValPad)
-        .addGrid(xgb.maximizeEvaluationMetrics, Array(true))
-        .addGrid(xgb.evalMetric, Array("aucpr"))
-        .addGrid(xgb.objective, Array("binary:logistic"))
+        .addGrid(xgb.maximizeEvaluationMetrics, DefaultSelectorParams.MaximizeEvaluationMetrics)
+        .addGrid(xgb.evalMetric, DefaultSelectorParams.BinaryClassXGBEvaluationMetric)
+        .addGrid(xgb.objective, DefaultSelectorParams.BinaryClassXGBObjective)
 
       Seq(lr -> lrParams, rf -> rfParams, gbt -> gbtParams, svc -> svcParams,
         nb -> nbParams, dt -> dtParams, xgb -> xgbParams)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -123,8 +124,12 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
       val xgbParams = new ParamGridBuilder()
         .addGrid(xgb.numRound, DefaultSelectorParams.NumRound)
         .addGrid(xgb.eta, DefaultSelectorParams.Eta)
-        .addGrid(xgb.maxDepth, DefaultSelectorParams.MaxDepth)
+        .addGrid(xgb.maxDepth, DefaultSelectorParams.MaxDepthXGBClassifier)
         .addGrid(xgb.minChildWeight, DefaultSelectorParams.MinChildWeight)
+        .addGrid(xgb.missing, DefaultSelectorParams.MissingValPad)
+        .addGrid(xgb.maximizeEvaluationMetrics, Array(true))
+        .addGrid(xgb.evalMetric, Array("aucpr"))
```
tovbinm, line 131: please define these in `DefaultSelectorParams`

TuanNguyen27:
When i load the XGBoost model to score, i got this error 

```
    org.scalatest.exceptions.TestFailedException: Record index 0: Map("survived_1-stagesApplied_RealNN_000000000014" -> 0.0, "age-description-gender-height-survived-weight_7-stagesApplied_Prediction_00000000002d" -> Map("probability_1" -> 0.375, "probability_0" -> 0.625, "rawPrediction_0" -> -0.375, "prediction" -> 0.0, "rawPrediction_1" -> 0.375), "description_1-stagesApplied_RealNN_000000000015" -> 3, "description_2-stagesApplied_Text_000000000016" -> "__unknown") was not equal to Map("age-description-gender-height-survived-weight_7-stagesApplied_Prediction_000000000027" -> Map("probability_1" -> 0.375, "probability_0" -> 0.625, "rawPrediction_0" -> -0.375, "prediction" -> 0.0, "rawPrediction_1" -> 0.375), "survived_1-stagesApplied_RealNN_000000000014" -> 0.0, "description_1-stagesApplied_RealNN_000000000015" -> 3.0, "description_2-stagesApplied_Text_000000000016" -> "__unknown")
```

which is due to the difference between `age-description-gender-height-survived-weight_7-stagesApplied_Prediction_00000000002d` and `age-description-gender-height-survived-weight_7-stagesApplied_Prediction_000000000027`. Do you have any idea why this is happening @Jauntbox @leahmcguire ?

core/src/main/scala/com/salesforce/op/stages/impl/selector/DefaultSelectorParams.scala
```diff
@@ -63,7 +63,14 @@ object DefaultSelectorParams {
...
-  val NumRound = Array(100) // number of rounds for xgboost (default 1)
-  val Eta = Array(0.1 , 0.3) // step size shrinkage for xgboost (default 0.3)
-  val MinChildWeight = Array(1.0, 5.0, 10.0) // minimum sum of instance weight needed in a child for xgboost (default 1)
+  val NumRound = Array(200) // number of rounds for xgboost (default 1)
+  val Eta = Array(0.02) // step size shrinkage for xgboost (default 0.3)
+  val MinChildWeight = Array(1.0, 10.0) // minimum sum of instance weight needed in a child for xgboost (default 1)
+  val MaxDepthXGBClassifier = Array(10) // max depth of each tree for xgboost (default is 6)
+  val MissingValPad = Array(0.toFloat) // treat 0 as missing value, (default is Float.NaN)
+  val BinaryClassXGBEvaluationMetric = Array("aucpr") // XGBoost evaluation metrics for validation data (default: error)
+  val BinaryClassXGBObjective = Array("binary:logistic") // XGBoost learning objective (default: reg:squarederror)
```
tovbinm, line 72: can we add an enum for `BinaryClassXGBEvaluationMetric` and `BinaryClassXGBObjective` values?

TuanNguyen27: the grid will need to take in a `string` still, how does that work with an enum ?

TuanNguyen27: We have similar `Array[String]` in our production code for GLM hyper-parameter grid, so I think it's okay to leave this as a string instead of an enum. What do you think ?

tovbinm: Ah, stupid grid. ok.

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -60,7 +60,8 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
      * off by default
      */
     val modelTypesToUse: Seq[BinaryClassificationModelsToTry] = Seq(
-      MTT.OpLogisticRegression, MTT.OpRandomForestClassifier, MTT.OpGBTClassifier
+      MTT.OpLogisticRegression, MTT.OpRandomForestClassifier,
```
Jauntbox, line 63: I think it makes more sense to replace GBTs with xgboost, rather than having both.

core/src/main/scala/com/salesforce/op/stages/impl/selector/DefaultSelectorParams.scala
```diff
@@ -63,7 +63,14 @@ object DefaultSelectorParams {
   //    - "gamma"    : "inverse", "identity", "log"
   //    - "tweedie"  : power link function specified through "linkPower". The default link power in
   //    the tweedie family is 1 - variancePower.
-  val NumRound = Array(100) // number of rounds for xgboost (default 1)
-  val Eta = Array(0.1 , 0.3) // step size shrinkage for xgboost (default 0.3)
-  val MinChildWeight = Array(1.0, 5.0, 10.0) // minimum sum of instance weight needed in a child for xgboost (default 1)
+  val NumRound = Array(200) // number of rounds for xgboost (default 1)
+  val Eta = Array(0.02) // step size shrinkage for xgboost (default 0.3)
+  val MinChildWeight = Array(1.0, 10.0) // minimum sum of instance weight needed in a child for xgboost (default 1)
+  val MaxDepthXGBClassifier = Array(10) // max depth of each tree for xgboost (default is 6)
```
Jauntbox, line 69: Can you make the variable names standard? Right now, some have "Classifier" at the end and some have "BinaryClass" and the beginning.

local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
```diff
@@ -153,7 +148,8 @@ class OpWorkflowModelLocalTest extends FlatSpec with PassengerSparkFixtureTest w
 
   private def assert(
     scores: Array[Map[String, Any]],
-    expectedScores: Array[(Prediction, RealNN, RealNN, Text)]
+    expectedScores: Array[(Prediction, RealNN, RealNN, Text)],
+    prediction: FeatureLike[Prediction]
```
Jauntbox, line 152: It doesn't look like you used this. Is it needed?

local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
```diff
@@ -169,6 +165,36 @@ class OpWorkflowModelLocalTest extends FlatSpec with PassengerSparkFixtureTest w
...
+  ): Unit = {
+    val scoreFn = OpWorkflowModel.load(modelLocation).scoreFunction
+    scoreFn shouldBe a[ScoreFunction]
+    val scores = rawData.map(scoreFn)
+    assert(scores, expectedScores, prediction)
+  }
+
+  private def buildAndSaveModel(modelsAndParams: Seq[(EstimatorType, Array[ParamMap])]) = {
```
Jauntbox, line 180: Fix the indenting in the body here (Code -> Reformat Code in IntelliJ)

local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
```diff
@@ -169,6 +165,36 @@ class OpWorkflowModelLocalTest extends FlatSpec with PassengerSparkFixtureTest w
...
+    val workflow = new OpWorkflow().setReader(dataReader)
+    .setResultFeatures(prediction, survivedNum, indexed, deindexed)
+    lazy val model = workflow.train()
+    val path = Paths.get(tempDir.toString, "op-runner-local-test-model").toFile.getCanonicalFile.toString
+    model.save(path)
+    (path, model, prediction)
+  }
+
+  private def genRawDataAndScore(model: OpWorkflowModel, prediction: FeatureLike[Prediction]) = {
+    lazy val rawData = dataReader.generateDataFrame(model.getRawFeatures()).sort(KeyFieldName).collect().map(_.toMap)
```
Jauntbox, line 193: I don't actually know, but do these local variables need to be lazy if the method itself is assigned to a lazy val?

TuanNguyen27: good point, let me try removing lazy val inside function def :)

local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
TuanNguyen27: it's used under `[prediction.name -> predV.value](https://github.com/salesforce/TransmogrifAI/blob/c4919cfb17824750622e33fe537e4d4f45463940/local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala#L158)`

PR #490, opened 2020-07-09
Let user specify feature importance type for XGBoost

TuanNguyen27:
XGBoost default feature importance is now `gain` instead of `weight`. Switching to `gain` will make XGBoost feature importance more consistent with [Spark's Random Forest](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/tree/treeModels.scala#L168-L197), which uses average of single tree importances across all trees in the ensemble.

Description of the available options for `importanceType`

```
‘gain’: the average gain across all splits the feature is used in.

‘cover’: the average coverage across all splits the feature is used in.

‘total_gain’: the total gain across all splits the feature is used in.

‘total_cover’: the total coverage across all splits the feature is used in.
```

### DIFF
--- a/core/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostParams.scala
+++ b/core/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostParams.scala
@@ -88,8 +88,8 @@ case object OpXGBoost {
     /**
      * Converts feature score map into a vector
      *
-     * @param featureVectorSize   size of feature vectors the xgboost model is trained on
-     * @param importanceType      type of feature importance to calculate ["gain, "cover", "total_gain", "total_cover"]
+     * @param featureVectorSize size of feature vectors the xgboost model is trained on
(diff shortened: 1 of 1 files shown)


core/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostParams.scala
```diff
@@ -87,10 +87,13 @@ case object OpXGBoost {
      * Converts feature score map into a vector
      *
      * @param featureVectorSize   size of feature vectors the xgboost model is trained on
+     * @param importanceType      type of feature importance to calculate ["gain, "cover", "total_gain", "total_cover"]
      * @return vector containing feature scores
      */
-    def getFeatureScoreVector(featureVectorSize: Option[Int] = None): Vector = {
-      val featureScore = booster.getFeatureScore()
+    def getFeatureScoreVector(
+      featureVectorSize: Option[Int] = None, importanceType: String = "gain"
```
tovbinm, line 94: let's add a enum for this

PR #491, opened 2020-07-14
Rename whitelist/blacklist to allowlist/blocklist

nicodv:
**Related issues**
N/A

N/A

N/A

**Additional context**
This lets Transmogrifai adhere to Salesforce's Inclusive Content initiative. Backwards compatible for models that still use old terms.


core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
```diff
@@ -165,9 +165,9 @@ private[op] object OpWorkflowModelReadWriteShared {
     val values = findValues
     case object Uid extends FieldNames("uid")
     case object ResultFeaturesUids extends FieldNames("resultFeaturesUids")
-    case object BlacklistedFeaturesUids extends FieldNames("blacklistedFeaturesUids")
-    case object BlacklistedMapKeys extends FieldNames("blacklistedMapKeys")
-    case object BlacklistedStages extends FieldNames("blacklistedStages")
+    case object BlocklistedFeaturesUids extends FieldNames("blocklistedFeaturesUids")
```
tovbinm, line 168: This change would any existing models with a blacklist defined behave incorrectly. I would recommend to keep backwards compatibility of `resolveBlocklist` (fka `resolveBlacklist`) at least until the next major release, i.e. 0.8.0.

nicodv: Good point. Let me try to keep backwards compatibility then. Marking as WIP for now.

nicodv: I've added backwards compatibility and a test for it, @tovbinm

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -166,31 +167,44 @@ class OpWorkflowModelReader(val workflowOpt: Option[OpWorkflow]) extends MLReade
...
+    }.maxBy(_.getOrElse(Array()).length)
   }
 
-  private def resolveBlacklistMapKeys(json: JValue): Try[Map[String, Set[String]]] = Try {
-    (json \ BlacklistedMapKeys.entryName).extractOpt[Map[String, List[String]]] match {
-      case Some(blackMapKeys) => blackMapKeys.map { case (k, vs) => k -> vs.toSet }
-      case None => Map.empty
-    }
+  private def resolveBlocklistMapKeys(json: JValue): Try[Map[String, Set[String]]] = Try {
+    // For backward compatibility. The relevant field name is determined
```
tovbinm, line 199: 5c: you have `extractOpt`, does it not return `Option[Map[...]]`?

nicodv: Indeed. Using `flatMap` now.

helloworld/src/main/resources/BostonDataset/housing.names
```diff
@@ -41,12 +41,9 @@
     9. RAD       index of accessibility to radial highways
     10. TAX      full-value property-tax rate per $10,000
     11. PTRATIO  pupil-teacher ratio by town
-    12. B        1000(Bk - 0.63)^2 where Bk is the proportion of blacks 
-                 by town
+    12. AA       1000(x - 0.63)^2 where x is the proportion of
```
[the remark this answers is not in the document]
nicodv: The fact that this variable is available is perhaps a bigger issue than the complicated question of terminology. Let me take this out of the PR and we can tackle the Boston data set separately.

For reference, Scikit-learn decided to remove this data set altogether: https://github.com/scikit-learn/scikit-learn/issues/16155

PR #492, opened 2020-07-14
Add thresholded confusion matrix elements to binary classification metrics

Jauntbox:
**Related issues**
N/A

~~Spark's threshold metrics only compute precision, recall, and false positive rate. It would be helpful to also have the confusion matrix components (TP, FP, TN, FN). Fortunately, the three quantities already calculated along with the equation TP + FP + TN + FN = Count provide a system of 4 independent equations in 4 unknowns, so the confusion matrix components can be calculated with closed-form formulae.~~

It turns out these equations become underdetermined whenever TP = 0 since both the precision and recall equations become 0 = 0. The actual calculation needed here (confusion matrices by threshold) is sitting in Spark's code, but not exposed (yet). Even worse, it's private instead of package private, so we can't get around it with an implicit class extension. For the moment, I'm doing a gross copy/paste to expose that info to our evaluators. The longer-term solution is to make a PR to Spark to expose this, so we can get rid of `RichBinaryClassificationMetrics`.

This is now plan B!

**Additional context**
A separate PR will add thresholded, discretized metrics for regression as well


core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
```diff
@@ -112,10 +114,29 @@ private[op] class OpBinaryClassificationEvaluator
...
+
+      // Solutions can be checked with any system of equations solver. For Wolfram Alpha, input:
+      // inv {{(1-P), -P, 0, 0}, {(1-R), 0, 0, -R}, {0, 1-"Fpr", -"Fpr", 0}, {1,1,1,1}} * {0, 0, 0, "Count"}
+      val (tupPos, tupNeg) = thresholds.zipWithIndex.map{ case (_, i) => {
+        val denom = 1.0 / recallByThreshold(i) + (1.0 / precisionByThreshold(i) - 1.0) / falsePositiveRateByThreshold(i)
+        val prefactor = count / denom
+        val tpByThreshold = math.round(prefactor).toDouble
+        val fpByThreshold = math.round(prefactor * (1.0 / precisionByThreshold(i) - 1.0)).toDouble
+        val tnByThreshold = math.round(prefactor * (1.0 / falsePositiveRateByThreshold(i) - 1.0) *
+          (1.0 / precisionByThreshold(i) - 1.0))toDouble
```
nicodv, line 126: missing `.` for `.toDouble`, although that's not a syntax error apparently. 😮

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
```diff
@@ -112,10 +114,29 @@ private[op] class OpBinaryClassificationEvaluator
       val falsePositiveRateByThreshold = sparkMLMetrics.roc().collect().map(_._1).slice(1, thresholds.length + 1)
       val aUROC = sparkMLMetrics.areaUnderROC()
       val aUPR = sparkMLMetrics.areaUnderPR()
+
+      // Solutions can be checked with any system of equations solver. For Wolfram Alpha, input:
+      // inv {{(1-P), -P, 0, 0}, {(1-R), 0, 0, -R}, {0, 1-"Fpr", -"Fpr", 0}, {1,1,1,1}} * {0, 0, 0, "Count"}
+      val (tupPos, tupNeg) = thresholds.zipWithIndex.map{ case (_, i) => {
```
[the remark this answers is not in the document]
Jauntbox: Ooo, good point - if any of the spark quantities are 0 that will cause problems. I can rewrite things so that there's less division, but I may need to put in a regularization param for some of them

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
Jauntbox: Oh wow, how did I miss that?

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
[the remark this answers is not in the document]
leahmcguire: yeah we dont want to fail a flow for a divide by zero here

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
Jauntbox: Removed!

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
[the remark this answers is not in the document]
Jauntbox: Removed!

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
```diff
@@ -195,7 +214,15 @@ case class BinaryClassificationMetrics
...
-  falsePositiveRateByThreshold: Seq[Double]
+  falsePositiveRateByThreshold: Seq[Double],
+  @JsonDeserialize(contentAs = classOf[java.lang.Double])
+  truePostitivesByThreshold: Seq[Double],
+  @JsonDeserialize(contentAs = classOf[java.lang.Double])
+  falsePositivesByThreshold: Seq[Double],
+  @JsonDeserialize(contentAs = classOf[java.lang.Double])
+  trueNegativesByThreshold: Seq[Double],
+  @JsonDeserialize(contentAs = classOf[java.lang.Double])
+  falseNegativesByThreshold: Seq[Double]
```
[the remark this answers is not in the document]
leahmcguire: agreed lets try to make it look as much like multiclass as possible

core/src/main/scala/org/apache/spark/mllib/evaluation/RichBinaryClassificationMetrics.scala
```diff
@@ -0,0 +1,285 @@
+/*
+ * Modified on 7/20/2020 to expose confusion matrices by threshold.
```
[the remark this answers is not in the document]
leahmcguire: like this and then describe the modifications 
`// scalastyle:off header.matches
/*
 * Modifications: (c) 2017, Salesforce.com, Inc.`

core/src/main/scala/org/apache/spark/mllib/evaluation/RichBinaryClassificationMetrics.scala
```diff
@@ -0,0 +1,285 @@
...
+ *                the curve. After down-sampling, the curves will instead be made of approximately
+ *                `numBins` points instead. Points are made from bins of equal numbers of
+ *                consecutive points. The size of each bin is
+ *                `floor(scoreAndLabels.count() / numBins)`, which means the resulting number
+ *                of bins may not exactly equal numBins. The last bin in each partition may
+ *                be smaller as a result, meaning there may be an extra sample at
+ *                partition boundaries.
+ */
+@Since("1.0.0")
+class RichBinaryClassificationMetrics @Since("3.0.0") (
```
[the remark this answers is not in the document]
leahmcguire: also you should get rid of the spark anotations

core/src/main/scala/org/apache/spark/mllib/evaluation/RichBinaryClassificationMetrics.scala
```diff
@@ -0,0 +1,285 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.mllib.evaluation
+
```
leahmcguire, line 23: why is this in the spark namespace? 

I am really unhappy with just copy pasting the whole class. It is an incredibly bad idea as it gives up a huge amount of code that we are responsible for maintaining feature parity on. 

It would be better to only copy the part that we need even if it means the calculations for that will be done twice when we use the original class

Jauntbox: It's in the Spark namespace because it uses other classes that are package private to `evaluation`. I couldn't copy out just the piece of code that does the confusion matrix calculation into our namespace since that piece relies on Spark's `BinaryLabelCounter` and `BinaryConfusionMatrix` classes. 

You'd really rather have me copy _just_ the `confusions` (which still needs to be in `org.apache.spark.mllib.evaluation`), and do the entire calculation twice? It seems wasteful to me to just get the benefit of avoiding a few extra pieces of the Spark class, but ok.

leahmcguire: There is a tradeoff between computational efficiency and code that will need to be maintained. I would prefer to have less copy pasted code over. I am fine with doing the calculations twice to get that

Jauntbox: There's an additional problem with partially copying this class. If we still get the rest of the binary classification metrics from a call to Spark's class, then we now rely on two implementations of the threshold calculations and these could diverge.

If we copy over everything that contributes to any of the threshold metrics we use so that all our metrics are consistent, then that's basically copying the entire class! Even the AuROC and AuPR calculations we use depend on the confusion matrices by threshold. 

I share the desire to minimize any copy/pasting from other libraries, but I don't see a way to partially copy this class without introducing problems like this.

Jauntbox: I changed it to extend Spark's class and just reimplement explicit confusion matrix piece, but I prefer copying the whole class. This approach still has the problems I mentioned in my comment above. I'll go with it if you and @nicodv think this is better, though.

nicodv: Thanks for clearly laying out the pros and cons, @Jauntbox . I do think this better, all things considered.

Your tests that check the reconstruction of the Spark-calculated thresholds should cover the divergence scenario, right?

nicodv:
LGTM

core/src/main/scala/org/apache/spark/mllib/evaluation/RichBinaryClassificationMetrics.scala
Jauntbox: The tests don't directly check the two threshold arrays against each other since I think it's too confusing to expose both of them. They do check for the situations we want to avoid the most (where the thresholds are at "different" points in the scores in the sense of model performance metrics). You can still have cases where the thresholds would shift up/down slightly and the precision/recall/fpr tests could still pass.

core/src/main/scala/com/salesforce/op/evaluators/OpBinaryClassificationEvaluator.scala
```diff
@@ -105,17 +107,27 @@ private[op] class OpBinaryClassificationEvaluator
...
+      val sparkMLMetrics = new RichBinaryClassificationMetrics(scoreAndLabels = scoreAndLabels, numBins = numBins)
       val thresholds = sparkMLMetrics.thresholds().collect()
       val precisionByThreshold = sparkMLMetrics.precisionByThreshold().collect().map(_._2)
       val recallByThreshold = sparkMLMetrics.recallByThreshold().collect().map(_._2)
       val falsePositiveRateByThreshold = sparkMLMetrics.roc().collect().map(_._1).slice(1, thresholds.length + 1)
       val aUROC = sparkMLMetrics.areaUnderROC()
       val aUPR = sparkMLMetrics.areaUnderPR()
+
+      val confusionMatrixByThreshold = sparkMLMetrics.confusionMatrixByThreshold().collect()
+      val (copiedTupPos, copiedTupNeg) = confusionMatrixByThreshold.map(c => {
```
leahmcguire, line 119: map.{ case (_, descriptiveName) => ((descriptiveName.numTruePositives, descriptiveName.numFalsePositives),

PR #494, opened 2020-07-20
Fix OpWorkflowModelLocalTest due to flaky XGBoost training

TuanNguyen27:
Source of flakiness: default `BinaryClassificationModelSelector.withTrainValidationSplit` sometimes makes the training set contain only positive or negative labels, which fails the training for xgboost.

We address this flakiness by fixing the seed in the `DataSplitter` for `withTrainValidationSplit`, which will result in the same train-test split every time the test is run.

```
ml.dmlc.xgboost4j.java.XGBoostError: [16:55:13] /xgboost/src/metric/rank_metric.cc:515: Check failed: !auc_error: AUC-PR: the dataset only contains pos or neg samples
```


gerashegalov:
Add a description of flakiness, and how your fix addresses it

PR #496, opened 2020-07-23
Add signed percentage error histogram to regression evaluator

nicodv:
**Related issues**
N/A

Adds a histogram of signed (as opposed to more common definition of absolute) error percentages to `OpRegressionEvaluator`. A couple of parameters are exposed to control the behavior of the calculation of this histogram, related to how to deal with label values around 0, as that would explode the percentages.

- `signedPercentageErrorHistogramBins`, an array determining the histogram bins
- `scaledErrorCutoff`: a label value cutoff below which the signed percentage error is implemented as a scaled error with a fixed denominator to avoid problems with label values around 0
- `smartCutoffRatio`: if set, `scaledErrorCutoff` is determined smartly by taking the average absolute magnitude of the data multiplied with this ratio

- Spark's `RegressionEvaluator` is based on summary statistics (e.g., SSE) that are computed in private fields, so there is no neat way to avoid going over the data once more for the percentage errors.
- Putting this in a separate evaluator is an option, but since `OpWorkflowModel.scoreAndEvaluate` takes a single evaluator, it would mean these metrics wouldn't always be available.

**Additional context**
Can be used to express error bands around regression (e.g., "given an error tolerance of 10%, you'd have this many correct, over- and under-predictions"), the need for which is the driving force behind this PR.


nicodv:
@leahmcguire , I considered that but `OpWorkflowModel.scoreAndEvaluate()` takes a single evaluator currently. I chose to add it to the existing `OpRegressionEvaluator` so that the metrics are broadly available (e.g. in our evaluation), not just in the `RegressionModelSelector`.

Let me know if you think we should reconsider that trade-off.

core/src/main/scala/com/salesforce/op/evaluators/EvaluationMetrics.scala
```diff
@@ -181,6 +179,8 @@ object RegressionEvalMetrics extends Enum[RegressionEvalMetric] {
   case object MeanSquaredError extends RegressionEvalMetric("mse", "mean square error", false)
   case object R2 extends RegressionEvalMetric("r2", "r2", true)
   case object MeanAbsoluteError extends RegressionEvalMetric("mae", "mean absolute error", false)
+  case object SignedPercentageErrorHistogram extends RegressionEvalMetric("signedpercentageerrorhistogram",
```
leahmcguire, line 182: "signedpercentageerrorhistogram" is not readable - how about "speh" or "prcntErrorHst"

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
```diff
@@ -55,6 +60,50 @@ private[op] class OpRegressionEvaluator
 
   @transient private lazy val log = LoggerFactory.getLogger(this.getClass)
 
+  final val signedPercentageErrorHistogramBins = new DoubleArrayParam(
+    parent = this,
+    name = "signedPercentageErrorHistogramBins",
+    doc = "sequence of bins for signed percentage error histogram",
```
leahmcguire, line 66: is this the error percentage bins or the prediction value bins? lets make the doc clear

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
```diff
@@ -55,6 +60,50 @@ private[op] class OpRegressionEvaluator
...
+    doc = "label value cutoff below which percentage error is implemented as a scaled error " +
+      "with a fixed denominator to avoid problems with label values around 0",
+    isValid = (d: Double) => d > 0.0
+  )
+  setDefault(scaledErrorCutoff, 1E-3)
+
+  def setScaledErrorCutoff(v: Double): this.type = set(scaledErrorCutoff, v)
+  def getScaledErrorCutoff: Option[Double] = get(scaledErrorCutoff)
+
+  final val smartCutoff = new BooleanParam(
```
leahmcguire, line 87: not sure you need this boolean param it seems overkill/harder to keep track of everything. perhaps just dont have a default for the smart cutoff value and if it is set use that instead of the scaled error cutoff.

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
```diff
@@ -86,6 +144,55 @@ private[op] class OpRegressionEvaluator
...
+   */
+  private def calculateSignedPercentageErrorHistogram(data: Dataset[_]): Array[Long] = {
+    // Prep data
+    val predictionsAndLabels = data
+      .select(col(getPredictionValueCol).cast(DoubleType), col(getLabelCol).cast(DoubleType))
+      .rdd
+      .map { case Row(prediction: Double, label: Double) => (prediction, label) }
+
+    // If we need to set the scaledErrorCutoff smartly, use the label data for that
+    if ($(smartCutoff)) {
```
leahmcguire, line 162: rather maybe check if smart cutoff value is set

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
```diff
@@ -86,6 +144,55 @@ private[op] class OpRegressionEvaluator
...
+      val cutoff = calculateSmartCutoff(predictionsAndLabels)
+      log.info(s"Smart scaledErrorCutoff was determined to be: $cutoff")
+      setScaledErrorCutoff(cutoff)
+    }
+
```
leahmcguire, line 167: get val for $(scaledErrorCutoff) here and pass into map

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
```diff
@@ -86,6 +144,55 @@ private[op] class OpRegressionEvaluator
...
+  /**
+   * Calculates the signed percentage error, with cutoff logic to avoid large results
+   * due to division by small numbers.
+   *
+   * @param prediction Predicted value
+   * @param label      Actual value
+   * @return Signed percentage error
+   */
+  private def calculateSignedPercentageError(prediction: Double, label: Double): Double = {
+    100.0 * (prediction - label) / (label.abs max $(scaledErrorCutoff))
```
leahmcguire, line 193: accessing spark parameters is actually quite inefficient - please access it once outside the map this calculation is called in and pass the value in as a param.

```
private def calculateSignedPercentageError(prediction: Double, label: Double, scldErrorCuttoff: Double): Double = {
    100.0 * (prediction - label) / (label.abs max scldErrorCutoff)
```

nicodv: Ooh, didn't know that, good call. 👍

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
nicodv: Yes, I was toying with this idea already. Let me implement it. 👍

nicodv: Done. It was easier than expected.

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
nicodv: Fixed

core/src/main/scala/com/salesforce/op/evaluators/OpRegressionEvaluator.scala
```diff
@@ -159,13 +149,14 @@ private[op] class OpRegressionEvaluator
...
-    if ($(smartCutoff)) {
+    if (isDefined(smartCutoffRatio)) {
       val cutoff = calculateSmartCutoff(predictionsAndLabels)
       log.info(s"Smart scaledErrorCutoff was determined to be: $cutoff")
       setScaledErrorCutoff(cutoff)
     }
 
-    val errors: RDD[Double] = predictionsAndLabels.map(x => calculateSignedPercentageError(x._1, x._2))
+    val errors: RDD[Double] = predictionsAndLabels
+      .map(x => calculateSignedPercentageError(x._1, x._2, $(scaledErrorCutoff)))
```
leahmcguire, line 159: since this is still in the map not sure it will fix the call
rather make a 
val scldErrorCutoff = $(scaledErrorCutoff)
and pass it into the map

nicodv: I believe this is a plain call-by-value situation, but I changed it to make that explicit.

PR #497, opened 2020-07-23
Avoid hardcoded default evaluators in ModelSelectors

nicodv:
**Related issues**
N/A

Since the model selectors hardcoded the default evaluators, we lacked the ability to set parameters on these evaluators there. This has been fixed by simply setting the evaluators as default arguments. This change was made to be backward compatible by prepending the defaults to the `Seq` of evaluators if they're not there.

You can add a duplicate of the default evaluator with different parameters to the list of evaluators, but you'd have to separate them again later on (and they'd be same class and name etc.).

**Additional context**
Related to https://github.com/salesforce/TransmogrifAI/pull/496, which introduces several parameters on `OpRegressionEvaluator`

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
@@ -148,8 +148,8 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
    * @param splitter            instance that will balance and split the data
    * @param numFolds            number of folds for cross validation (>= 2)
    * @param validationMetric    metric name in evaluation: AuROC or AuPR
-   * @param trainTestEvaluators List of evaluators applied on training + holdout data for evaluation. Default is empty
-   *                            and default evaluator is added to this list (here Evaluators.BinaryClassification)
+   * @param trainTestEvaluators List of evaluators applied on training + holdout data for evaluation. Default is
+   *                            the standard OpBinaryClassificationEvaluator and OpBinScoreEvaluator.
    * @param seed                random seed
    * @param stratify            whether or not stratify cross validation. Caution : setting that param to true might
    *                            impact the runtime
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -182,9 +183,13 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
       numFolds = numFolds, seed = seed, evaluator = validationMetric, stratify = stratify,
       parallelism = parallelism, maxWait = maxWait
     )
+    // For backwards compatibility, make sure evaluators always include the defaults
+    val allEvaluators = addDefaultEvaluators(trainTestEvaluators, Seq(
+      new OpBinaryClassificationEvaluator, new OpBinScoreEvaluator))
```
leahmcguire, line 188: this will run the same evaluators twice by default

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelectorFactory.scala
```diff
@@ -105,4 +103,13 @@ trait ModelSelectorFactory {
...
   }
 
+  protected def addDefaultEvaluators(
+    evaluators: Seq[OpEvaluatorBase[_ <: EvaluationMetrics]],
+    defaultEvaluators: Seq[OpEvaluatorBase[_ <: EvaluationMetrics]]
+  ): Seq[OpEvaluatorBase[_ <: EvaluationMetrics]] = {
+    defaultEvaluators.filter {
+      e => !evaluators.exists(_.isInstanceOf[e.type])
+    } ++ evaluators
+  }
```
leahmcguire, line 113: perhaps just a fold left to remove duplicates?

core/src/main/scala/com/salesforce/op/stages/impl/classification/BinaryClassificationModelSelector.scala
```diff
@@ -231,9 +237,13 @@ case object BinaryClassificationModelSelector extends ModelSelectorFactory {
       trainRatio = trainRatio, seed = seed, validationMetric, stratify = stratify, parallelism = parallelism,
       maxWait = maxWait
     )
+    // For backwards compatibility, make sure evaluators always include the defaults
+    val allEvaluators = addDefaultEvaluators(trainTestEvaluators, Seq(
+      new OpBinaryClassificationEvaluator, new OpBinScoreEvaluator))
```
leahmcguire, line 242: these are still functionally hardcoded. what is the benefit of doing it this way?

PR #501, opened 2020-07-29
Refactor flaky local scoring tests

Jauntbox:
**Related issues**
N/A

The current local scoring tests are flaky when xgboost models are included because the dataset used is a tiny 8-row hardcoded dataset. This can cause the train/validation splits to often contain all of a single class which causes xgboost models to throw an error.

This PR makes the dataset used a synthetic dataset of adjustable size (now 100 rows) to fix this problem.

We could also have used the full hardcoded Titanic dataset, but this was much easier for me.

**Additional context**
N/A

### DIFF
--- a/local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
+++ b/local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
@@ -40,9 +40,8 @@ import com.salesforce.op.stages.impl.classification.{BinaryClassificationModelSe
 import com.salesforce.op.stages.impl.feature.StringIndexerHandleInvalid
 import com.salesforce.op.stages.impl.selector.DefaultSelectorParams
 import com.salesforce.op.stages.impl.selector.ModelSelectorNames.EstimatorType
-import com.salesforce.op.stages.impl.tuning.DataSplitter
-import com.salesforce.op.test.{PassengerSparkFixtureTest, TestCommon, TestFeatureBuilder}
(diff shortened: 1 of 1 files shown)


Jauntbox:
Oof, didn't realize you weren't a "code owner", I guess @leahmcguire needs to approve this too

local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
```diff
@@ -57,14 +56,31 @@ import org.slf4j.LoggerFactory
 
 
 @RunWith(classOf[JUnitRunner])
-class OpWorkflowModelLocalTest extends FlatSpec with PassengerSparkFixtureTest with TestCommon {
-
+class OpWorkflowModelLocalTest extends FlatSpec with TestSparkContext with TempDirectoryTest with TestCommon {
```
tovbinm, line 59: Where do you use temp directory here?

Jauntbox: For example, here: https://github.com/salesforce/TransmogrifAI/blob/30cd461968ab9df9648afc7e6eb7f31a9d8f155f/local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala#L207

The test originally pulled it in via `PassengerSparkFixtureTest`, but I only needed things in `TempDirectoryTest`.

tovbinm: Got it.

PR #502, opened 2020-08-21
load workflows without spark context directly into MLeap transformers

leahmcguire:
Changes model loading to not require spark context and allows for local loading of spark wrapped stages into MLleap transformers rather than spark stages that are then converted to mleap for local scoring.

The final pr will fix the spark predictor wrapped stages to work with mleap loaded models.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
@@ -347,19 +347,12 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
   def train(persistEveryKStages: Int = OpWorkflowModel.PersistEveryKStages)
     (implicit spark: SparkSession): OpWorkflowModel = {
 
-    val (fittedStages, newResultFeatures) =
-      if (stages.exists(_.isInstanceOf[Estimator[_]])) {
-        val rawData = generateRawData()
-        // Update features with fitted stages
-        val fittedStgs = fitStages(data = rawData, stagesToFit = stages, persistEveryKStages)
-        val newResultFtrs = resultFeatures.map(_.copyWithNewStages(fittedStgs))
-        fittedStgs -> newResultFtrs
-      } else if (rawFeatureFilter.nonEmpty) {
-        generateRawData()
-        stages -> resultFeatures
-      } else {
-        stages -> resultFeatures
-      }
+    val rawData = generateRawData()
+    // Update features with fitted stages
+    val fittedStgs = fitStages(data = rawData, stagesToFit = stages, persistEveryKStages)
+    val newResultFtrs = resultFeatures.map(_.copyWithNewStages(fittedStgs))
+    val (fittedStages, newResultFeatures) = fittedStgs -> newResultFtrs
+
 
     val model =
       new OpWorkflowModel(uid, getParameters())
@@ -397,10 +390,6 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
       .map(_.filter(s => stagesToFit.contains(s._1)))
       .filter(_.nonEmpty)
 
-    // Search for the last estimator
-    val indexOfLastEstimator: Option[Int] =
-      dag.collect { case seq if seq.exists(_._1.isInstanceOf[Estimator[_]]) => seq.head._2 }.lastOption
-
     // doing regular workflow fit without workflow level CV
     if (!isWorkflowCV) {
       // The cross-validation job group is handled in the appropriate Estimator
@@ -410,7 +399,6 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
           train = train,
           test = test,
           hasTest = hasTest,
-          indexOfLastEstimator = indexOfLastEstimator,
           persistEveryKStages = persistEveryKStages
         ).transformers
       }
@@ -427,7 +415,6 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
             train = train,
             test = test,
             hasTest = hasTest,
-            indexOfLastEstimator = indexOfLastEstimator,
             persistEveryKStages = persistEveryKStages
           )
         }
@@ -458,7 +445,6 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
               train = trainFixed,
               test = testFixed,
               hasTest = hasTest,
-              indexOfLastEstimator = indexOfLastEstimator,
               persistEveryKStages = persistEveryKStages,
               fittedTransformers = beforeTransformers
             ).transformers
--- a/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
@@ -349,10 +349,8 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
 
     val rawData = generateRawData()
     // Update features with fitted stages
-    val fittedStgs = fitStages(data = rawData, stagesToFit = stages, persistEveryKStages)
-    val newResultFtrs = resultFeatures.map(_.copyWithNewStages(fittedStgs))
-    val (fittedStages, newResultFeatures) = fittedStgs -> newResultFtrs
-
+    val fittedStages = fitStages(data = rawData, stagesToFit = stages, persistEveryKStages)
+    val newResultFeatures = resultFeatures.map(_.copyWithNewStages(fittedStages)
 
     val model =
       new OpWorkflowModel(uid, getParameters())
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
@@ -464,7 +464,7 @@ case object OpWorkflowModel {
    * Load a previously trained workflow model from path
    *
    * @param path to the trained workflow model
-   * @param asSpark if true will load as spark models if false will load as Mlleap stages for spark wrapped stages
+   * @param asSpark if true will load as spark models if false will load as Mleap stages for spark wrapped stages
    * @return workflow model
    */
   def load(path: String, asSpark: Boolean = true): OpWorkflowModel =
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
@@ -466,6 +466,7 @@ case object OpWorkflowModel {
    * @param path to the trained workflow model
    * @return workflow model
    */
-  def load(path: String): OpWorkflowModel = new OpWorkflowModelReader(None).load(path)
+  def load(path: String, asSpark: Boolean = true): OpWorkflowModel =
+    new OpWorkflowModelReader(None, asSpark).load(path)
 
 }
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
@@ -464,6 +464,7 @@ case object OpWorkflowModel {
    * Load a previously trained workflow model from path
    *
    * @param path to the trained workflow model
+   * @param asSpark if true will load as spark models if false will load as Mlleap stages for spark wrapped stages
    * @return workflow model
    */
   def load(path: String, asSpark: Boolean = true): OpWorkflowModel =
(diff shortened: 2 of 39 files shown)


core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
```diff
@@ -466,6 +466,7 @@ case object OpWorkflowModel {
    * @param path to the trained workflow model
    * @return workflow model
    */
-  def load(path: String): OpWorkflowModel = new OpWorkflowModelReader(None).load(path)
+  def load(path: String, asSpark: Boolean = true): OpWorkflowModel =
```
tovbinm, line 469: missing docs
`asSpark if true will load as spark models if false will load as Mlleap stages for spark wrapped stages`

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -224,3 +230,30 @@ class OpWorkflowModelReader(val workflowOpt: Option[OpWorkflow]) extends MLReade
   }
 
 }
+
+private object FileReader {
+
+  def loadFile(pathString: String)(implicit conf: Configuration): String = {
```
tovbinm, line 236: I think this should be moved to `com.salesforce.op.utils.io.hadoop` in `utils` sub-project.

core/src/main/scala/com/salesforce/op/stages/sparkwrappers/specific/SparkModelConverter.scala
```diff
@@ -101,4 +100,30 @@ object SparkModelConverter {
     }
   }
 
+  /**
+   * Converts supported mleap loaded spark model of type PredictionModel[Vector, T] to an OP model
```
tovbinm, line 104: Is this going to be fixed in the next PR?

features/src/main/scala/com/salesforce/op/stages/SparkStageParam.scala
```diff
@@ -100,50 +104,83 @@ class SparkStageParam[S <: PipelineStage with Params]
...
+  /**
+   * Decodes the saved path string and uses the Pipeline.load method
+   * to recover the stage as an Mleap transformer
+   */
+  def jsonDecodeMLleap(jsonStr: String): Option[Either[S, MLeapTransformer]] = {
+    getPathUid(jsonStr) match {
+      case (None, _, _) | (_, None, _) | (_, Some(NoUID), _) =>
+        savePath = None
+        None
+      case (Some(p), Some(stageUid), asSpark) =>
```
tovbinm, line 160: please rename `Some(p)` to `Some(path)` so it would be clear what it is.

features/src/main/scala/com/salesforce/op/stages/SparkStageParam.scala
```diff
@@ -100,50 +104,83 @@ class SparkStageParam[S <: PipelineStage with Params]
...
-              case Success(mod) => mod
-            }
-          }
-          }.opt
-        dirBundle.map(_.root.asInstanceOf[S]).orElse{ // for backwards compatibility
+      case Left(spark) => Option(spark)
+    }.orElse { // for backwards compatibility
+      getPathUid(jsonStr) match {
+        case (_, Some(NoUID), _) => None
+        case (Some(p), Some(stageUid), Some(true)) =>
```
tovbinm, line 130: please rename `Some(p)` to `Some(path)` so it would be clear what it is.

core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
TuanNguyen27: are we exposing `asSpark` to be overridable in production ?

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
```diff
@@ -233,9 +240,11 @@ final class SelectedModel private[op]
     case m => setDefault(sparkMlStage, Option(m))
   }
 
-  @transient private lazy val recoveredStage: ModelType = getSparkMlStage() match {
-    case Some(m: PredictionModel[_, _]) => SparkModelConverter.toOPUnchecked(m).asInstanceOf[ModelType]
-    case Some(m: ModelType@unchecked) => m
+  @transient private lazy val recoveredStage: ModelType = (getSparkMlStage(), getLocalMlStage()) match {
+    case (Some(m: PredictionModel[_, _]), _) => SparkModelConverter.toOPUnchecked(m).asInstanceOf[ModelType]
+    case (Some(m: ModelType@unchecked), _) => m
+//    case (None, Some(m: MLeapTransformer)) => SparkModelConverter.toOPUnchecked(m).asInstanceOf[ModelType]
```
TuanNguyen27, line 246: Should this be uncommented ?

core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
leahmcguire: yes we will allow models to be loaded with and without spark

core/src/main/scala/com/salesforce/op/stages/impl/selector/ModelSelector.scala
leahmcguire: it is in the next pr when I make it work :-)

core/src/main/scala/com/salesforce/op/stages/sparkwrappers/specific/SparkModelConverter.scala
leahmcguire: yep

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
leahmcguire: this is a not general util since it only works for things saved with 1 part file. I could modify it to work in general but it loads everything up into memory which is probably not something we want for most hadoop files. @gerashegalov any thoughts?

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -347,19 +347,12 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
...
-        generateRawData()
-        stages -> resultFeatures
-      } else {
-        stages -> resultFeatures
-      }
+    val rawData = generateRawData()
+    // Update features with fitted stages
+    val fittedStgs = fitStages(data = rawData, stagesToFit = stages, persistEveryKStages)
+    val newResultFtrs = resultFeatures.map(_.copyWithNewStages(fittedStgs))
+    val (fittedStages, newResultFeatures) = fittedStgs -> newResultFtrs
```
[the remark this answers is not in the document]
nicodv: val fittedStages = fitStages(data = rawData, stagesToFit = stages, persistEveryKStages)
    val newResultFeatures = resultFeatures.map(_.copyWithNewStages(fittedStages))

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -30,19 +30,23 @@
 
 package com.salesforce.op
 
+
```
nicodv, line 33: nit: whitespace

core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
```diff
@@ -464,8 +464,10 @@ case object OpWorkflowModel {
    * Load a previously trained workflow model from path
    *
    * @param path to the trained workflow model
+   * @param asSpark if true will load as spark models if false will load as Mlleap stages for spark wrapped stages
```
nicodv, line 467: typo: MLeap

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -51,23 +55,24 @@ import scala.util.{Failure, Success, Try}
  * NOTE: The FeatureGeneratorStages will not be recovered into the Model object, because they are part of each feature.
  *
  * @param workflowOpt optional workflow that produced the trained model
+ * @param asSpark if true will load as spark models if false will load as Mlleap stages for spark wrapped stages
```
nicodv, line 58: typo: MLeap

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
nicodv: The name could make that limitation clear, e.g. `loadSinglePartFile`.

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -224,3 +230,30 @@ class OpWorkflowModelReader(val workflowOpt: Option[OpWorkflow]) extends MLReade
   }
 
 }
+
+private object FileReader {
```
nicodv, line 234: Please add tests for this

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -224,3 +230,30 @@ class OpWorkflowModelReader(val workflowOpt: Option[OpWorkflow]) extends MLReade
...
+    val fs = path.getFileSystem(conf)
+    val allFiles = fs.listFiles(path, false)
+    var partPath: Option[Path] = None
+    while (allFiles.hasNext) {
+      val p = allFiles.next().getPath
+      if (p.getName.startsWith("part")) {
+        partPath = Option(p)
+      }
+    }
+    val finalPath = partPath.getOrElse(path)
```
nicodv, line 247: If this is meant for reading single part files, would we want to drop this `.getOrElse(path)` in favor of an exception?

leahmcguire: that is there because some of our backwards compatibility tests use non part file names. this is meant to read a single workflow .json file if you use our save method this is a part file if you have passed it around it may have a name

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
leahmcguire: it is used in dozens of tests as is because it replaced the spark read

leahmcguire: https://github.com/salesforce/TransmogrifAI/blob/master/core/src/test/scala/com/salesforce/op/OpWorkflowModelReaderWriterTest.scala

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
leahmcguire: I will change the object name to indicate it is only fr workflow files

gerashegalov: I think we can wait with the cosmetics until there is a need to use it somewhere else unless we already see how to reuse this function. I want to first make sure we get the semantics right

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -224,3 +230,30 @@ class OpWorkflowModelReader(val workflowOpt: Option[OpWorkflow]) extends MLReade
...
+private object FileReader {
+
+  def loadFile(pathString: String)(implicit conf: Configuration): String = {
+    val path = new Path(pathString)
+    val fs = path.getFileSystem(conf)
+    val allFiles = fs.listFiles(path, false)
+    var partPath: Option[Path] = None
+    while (allFiles.hasNext) {
+      val p = allFiles.next().getPath
+      if (p.getName.startsWith("part")) {
```
gerashegalov, line 243: are we really trying to grab a random part file here. POSIX does not specify the order of file entries. If we just want to grab part-00000 just specify it.

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -224,3 +230,30 @@ class OpWorkflowModelReader(val workflowOpt: Option[OpWorkflow]) extends MLReade
   }
 
 }
+
+private object FileReader {
+
+  def loadFile(pathString: String)(implicit conf: Configuration): String = {
+    val path = new Path(pathString)
+    val fs = path.getFileSystem(conf)
+    val allFiles = fs.listFiles(path, false)
```
gerashegalov, line 239: When we call list, we won't be much better than Spark

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
leahmcguire: I merged everything into a single PR but I will update this in the master PR

leahmcguire: the reason I am looking for starts with is because sometimes they are compressed and sometimes not in the tests

PR #503, opened 2020-08-27
Making predictive models work with local load of wrapped spark models

leahmcguire:
**Related issues**

**Additional context**

### DIFF
--- a/core/src/main/scala/com/salesforce/op/stages/impl/classification/OpNaiveBayes.scala
+++ b/core/src/main/scala/com/salesforce/op/stages/impl/classification/OpNaiveBayes.scala
@@ -103,10 +103,10 @@ class OpNaiveBayesModel
 ) extends OpProbabilisticClassifierModel[NaiveBayesModel](
   sparkModel = sparkModel, uid = uid, operationName = operationName
 ) {
-  @transient lazy val predictRawMirror = reflectMethod(getSparkMlStage().get, "predictRaw")
-  @transient lazy val raw2probabilityMirror = reflectMethod(getSparkMlStage().get, "raw2probability")
-  @transient lazy val probability2predictionMirror =
-    reflectMethod(getSparkMlStage().get, "probability2prediction")
+  @transient lazy val predictRawMirror = getSparkOrLocalMethod("predictRaw", "predictRaw")
+  @transient lazy val raw2probabilityMirror = getSparkOrLocalMethod("raw2probability", "rawToProbability")
+  @transient lazy val probability2predictionMirror = getSparkOrLocalMethod("probability2prediction",
+    "probabilityToPrediction")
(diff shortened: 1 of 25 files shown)


core/src/main/scala/com/salesforce/op/stages/impl/classification/OpNaiveBayes.scala
```diff
@@ -103,10 +103,10 @@ class OpNaiveBayesModel
 ) extends OpProbabilisticClassifierModel[NaiveBayesModel](
   sparkModel = sparkModel, uid = uid, operationName = operationName
 ) {
-  @transient lazy val predictRawMirror = reflectMethod(getSparkMlStage().get, "predictRaw")
-  @transient lazy val raw2probabilityMirror = reflectMethod(getSparkMlStage().get, "raw2probability")
-  @transient lazy val probability2predictionMirror =
-    reflectMethod(getSparkMlStage().get, "probability2prediction")
+  @transient lazy val predictRawMirror = getSparkOrLocalMethod("predictRaw", "predictRaw")
+  @transient lazy val raw2probabilityMirror = getSparkOrLocalMethod("raw2probability", "rawToProbability")
```
nicodv, line 107: nit: any way to align these names to `rawToProbabilitySpark` / `rawToProbabilityLocal`? same for `probabilityToPrediction`

core/src/main/scala/com/salesforce/op/stages/impl/regression/OpXGBoostRegressor.scala
```diff
@@ -361,4 +364,18 @@ class OpXGBoostRegressionModel
...
-)
+) {
+  import OpXGBoost._
+  @transient private lazy val localModel = getLocalMlStage().map(s => s.model.asInstanceOf[MleapXGBoostRegressionModel])
+  @transient lazy protected val predict: Vector => Double = {
+    getSparkMlStage().map(s => s.predict(_))
+      .getOrElse{
+        features: Vector => {
+          val model = localModel.getOrElse(throw new RuntimeException(s"Could not find the wrapped Spark stage."))
+          val dm = new DMatrix(processMissingValues(Iterator(features.asXGB), 0.0F))
```
nicodv, line 375: nit: one-line doc about what pre-processing happens here and why would be nice. Maybe extract a method for it?

core/src/main/scala/com/salesforce/op/stages/impl/classification/OpNaiveBayes.scala
leahmcguire: So these are the names on the underlying spark and mleap classes. We are using reflection to access private methods (a dangerous hack but...) so we dont have any control over the names.

PR #505, opened 2020-09-02
changed to look specifically for part files rather than listing the files and searching through them

leahmcguire:
**Related issues**

**Additional context**

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
@@ -34,6 +34,7 @@ package com.salesforce.op
 import com.salesforce.op.OpWorkflowModelReadWriteShared.{FieldNames => FN}
 import com.salesforce.op.OpWorkflowModelReadWriteShared.FieldNames._
 import com.salesforce.op.OpWorkflowModelReadWriteShared.DeprecatedFieldNames._
+import com.salesforce.op.WorkflowFileReader.readAsString
 import com.salesforce.op.features.{FeatureJsonHelper, OPFeature, TransientFeature}
 import com.salesforce.op.filters.{FeatureDistribution, RawFeatureFilterResults}
 import com.salesforce.op.stages.OpPipelineStageReaderWriter._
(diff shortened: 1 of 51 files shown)


tovbinm:
Is it faster? What if another compression technique is being used, i.e. 7Zip, then we would have `.7z` extension?

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -236,15 +236,11 @@ private object WorkflowFileReader {
...
-    var partPath: Option[Path] = None
-    while (allFiles.hasNext) {
-      val p = allFiles.next().getPath
-      if (p.getName.startsWith("part-00000")) {
-        partPath = Option(p)
-      }
-    }
-    val finalPath = partPath.getOrElse(path)
+
+    val partFile = new Path(pathString, "part-00000")
```
gerashegalov, line 240: better configure the part file name on the write call path so we don't have to guess.

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -236,15 +236,11 @@ private object WorkflowFileReader {
...
+    val partZipped = new Path(pathString, "part-00000.gz")
+    val finalPath = if (fs.exists(partZipped)) partZipped else if (fs.exists(partFile)) partFile else path
```
gerashegalov, line 242: exists add redundant filesystem calls that will be even more expensive with S3

we can just call open directly and catch FileNotFound to try another filename

leahmcguire:
@gerashegalov suggested this for faster reading

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
leahmcguire: it will always be this but the compression codec will determine the extension

gerashegalov:
> Is it faster? What if another compression technique is being used, i.e. 7Zip, then we would have .7z extension?

@tovbinm it should be faster on object stores since you have fast path lookups but slow entry list performance.

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
leahmcguire: @gerashegalov fixed to just try loading

leahmcguire:
@tovbinm this covers no compression and our default settings as well as passing in the exact full path. Do you think we need to cover more possibilities?

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -234,20 +235,31 @@ class OpWorkflowModelReader(val workflowOpt: Option[OpWorkflow], val asSpark: Bo
...
+      } recoverWith { case suppressed2 => suppressed2.addSuppressed(suppressed1)
+          Try {
+            readAsString( new Path(pathString) )
+          } recoverWith {
+            case th => th.addSuppressed(suppressed2)
+              Failure(th)
+          }
+        }
+      } match {
+      case Failure(e) => throw new RuntimeException(s"Failed to load workflow because of $e")
```
gerashegalov, line 252: Let us keep `e` as the cause exception as well, i.e. provided as the second parameter to the RTE constructor.

gerashegalov (approved):
LGTM, minor comment

gerashegalov:
> this covers no compression and our default settings as well as passing in the exact full path. Do you think we need to cover more possibilities?

I don't think we need to make it that complex. At most I would make Compression Codec configurable instead of doing any guesses at all.

PR #510, opened 2020-09-08
Fix binary classification threshold metrics non-determinism

Jauntbox:
**Related issues**
N/A

The initial thresholded confusion matrix for binary classification metrics was added in https://github.com/salesforce/TransmogrifAI/pull/492, through selective overriding of Spark's BinaryClassificationMetrics class to expose only the thresholded confusion matrices. To do this, the calculation had to be done a second time.

Unfortunately, this calculation is non-deterministic. I haven't figured out exactly which piece, but I'm worried about the partitioning since a mapPartitions is used. This means that we need to go back to the original plan of copying the entire Spark class, since that's the only way to guarantee that the confusion matrices agrees with the rest of the thresholded metrics.

Reimplementing the other thresholded metrics amounts to copy/pasting the entire class so isn't really an alternative.

**Additional context**
N/A


core/src/main/scala/org/apache/spark/mllib/evaluation/RichBinaryClassificationMetrics.scala
```diff
@@ -38,16 +40,107 @@ import org.apache.spark.rdd.RDD
  *                partition boundaries.
  */
 class RichBinaryClassificationMetrics(
-  override val scoreAndLabels: RDD[(Double, Double)],
-  override val numBins: Int
-) extends BinaryClassificationMetrics(scoreAndLabels, numBins) {
+  val scoreAndLabels: RDD[(Double, Double)],
+  val numBins: Int
+) {
```
tovbinm, line 45: why the base class was dropped?

nicodv: Wondering this too. We should be able to re-use some thing, no, @Jauntbox ?

Jauntbox: We used to extend Spark's class since we were only partially overriding it, but I don't think we need to extend that class anymore. We're completely copying it here so aren't actually sharing code with it anymore.

Is there anything to gain by extending it?

tovbinm: If you arent extending it, then we should move `RichBinaryClassificationMetrics` out of that package and consider renaming it to `BinaryClassificationMetrics`

tovbinm: (this might be a breaking change though)

Jauntbox: Yeah, renaming it makes sense. I can check if moving it causes a lot of trouble.

Jauntbox: I renamed to `SparkBinaryClassifiationMetrics` to make it clearer. It has to live in org.apache.spark.mllib.evaluation for its imports.

@nicodv Unfortunately we need to replicate everything since all the other metrics are built from `cumulativeCounts` and `confusions`.

nicodv (approved):
LGTM

leahmcguire (changes requested):
@kevin lets talk about what you need here we should not be copy pasting the class. if you just need access to private methods we can use reflection to access them

gerashegalov:
Strong +1 for using reflection in the interim instead of copy&pasting. And let us make a PR to Spark

leahmcguire (approved):
: -)

core/src/main/scala/org/apache/spark/mllib/evaluation/ExtendedBinaryClassificationMetrics.scala
```diff
@@ -0,0 +1,20 @@
+package org.apache.spark.mllib.evaluation
```
nicodv, line 1: Header?

core/src/test/scala/com/salesforce/op/evaluators/ExtendedBinaryClassificationMetricsTest.scala
```diff
@@ -0,0 +1,28 @@
+package com.salesforce.op.evaluators
```
nicodv, line 1: Header?

Jauntbox: Good catch! Still barely in negative lines :P

core/src/main/scala/org/apache/spark/mllib/evaluation/ExtendedBinaryClassificationMetrics.scala
```diff
@@ -0,0 +1,50 @@
...
+  override val scoreAndLabels: RDD[(Double, Double)],
+  override val numBins: Int
+) extends BinaryClassificationMetrics(scoreAndLabels, numBins) {
+
+  /**
+   * Exposes the thresholded confusion matrices that Spark uses to calculate other derived metrics.
+   * @return RDD of (threshold, BinaryConfusionMatrix) over all thresholds calculated
+   */
+  def confusionMatrixByThreshold(): RDD[(Double, BinaryConfusionMatrix)] = {
+    val method = this.getClass.getSuperclass.getDeclaredMethod("confusions")
```
gerashegalov, line 46: let us do the method lookup once, suggestion PR #513 on top of yours

gerashegalov (changes requested):

gerashegalov (approved):
LGTM

PR #512, opened 2020-09-11
Experiment with OpenNLP Language detection

gerashegalov:


core/build.gradle
```diff
@@ -9,8 +9,7 @@ dependencies {
     compile "com.googlecode.libphonenumber:carrier:$googleCarrierVersion"
 
     // Optimaize language detection
-    compile "com.salesforce.transmogrifai:language-detector:$optimaizeLangDetectorVersion"
-
+    compile "com.optimaize.languagedetector:language-detector:$optimaizeLangDetectorVersion"
```
tovbinm, line 12: There was some fixes I did into Optimaize language detector. I think it was related to Guava.

tovbinm:
Let us know if it works better or worse than Optimaize one.

tovbinm:
It seems that it has more languages supported. Is this correct?

core/build.gradle
tovbinm: https://github.com/optimaize/language-detector/pull/74

gerashegalov: I was hoping to get rid of this fork by shading if necessary.

tovbinm: Sure. No development is happening on that repo, so not a big deal imo.

Jauntbox:
Cool - curious to see how this compares. Another one we could try is FastText, which also has a language detection module: https://github.com/facebookresearch/fastText/#full-documentation

PR #515, opened 2020-09-23
Add topk metrics for multiclass classifier

marksho-sfdc:
**Related issues**
N/A

Adds the ability to get multiclass classification metrics on topK labels

N/A

**Additional context**
N/A


core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -118,19 +129,68 @@ private[op] class OpMultiClassificationEvaluator
...
+ * @param topKs      Sequence of topK values to calculate multiclass classification metrics for
+ */
+  def calculateTopKMetrics(
+    data: RDD[(Array[Double], Double)],
+    topKs: Seq[Int]
+  ): MultiClassificationMetricsTopK = {
+    type Label = Int
+
+    def computeMetrics(scoresAndLabels: (Array[Double], Double)): (Array[Double], Double) = {
+      val scores: Array[Double] = scoresAndLabels._1
```
tovbinm, line 165: use case instead of `_1` and `_2`, ie.
```scala
{
   case (scores: Array[Double], label: Label)
}
```
same applies elsewhere in your code

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -271,7 +331,24 @@ case class MultiClassificationMetrics
...
+ * Metrics for topK multiclass classification
+ *
+ * @param PrecisionTopK
+ * @param RecallTopK
+ * @param F1TopK
+ * @param ErrorTopK
+ */
+case class MultiClassificationMetricsTopK
+(
+  PrecisionTopK: Seq[Double],
```
tovbinm, line 348: the class already has TopK in the name, no need to also include it each of the member names.

tovbinm (commented):
thanks for the contribution! see some comments

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -75,6 +75,16 @@ private[op] class OpMultiClassificationEvaluator
 
   def setTopNs(v: Array[Int]): this.type = set(topNs, v)
 
+  final val topKs = new IntArrayParam(
+    parent = this,
+    name = "topKs",
+    doc = "sequence of topK values to use for top K metrics",
+    isValid = _.forall(_ > 0)
```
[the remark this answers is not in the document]
Jauntbox: Good point, we probably want a limit on how many of these we allow (10-20?)

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -75,6 +75,16 @@ private[op] class OpMultiClassificationEvaluator
...
+  )
+  setDefault(topKs, Array(5, 10, 20, 50, 100, 200))
```
Jauntbox, line 84: Let's keep the max topK in line with the max cardinality we're setting for multiclass classification here (100): https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/stages/impl/tuning/Splitter.scala#L179

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -118,19 +129,68 @@ private[op] class OpMultiClassificationEvaluator
         thresholds = $(thresholds)
       )
 
+      val metricsTopK = calculateTopKMetrics(
+        data = dataUse.select(col(probabilityColName), col(labelColName).cast(DoubleType)).rdd.map{
```
Jauntbox, line 133: Let's do the RDD conversion once since we're sending the same one to the threshold metrics calculation

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
Jauntbox: Agreed, we also need to store the topK values themselves so it's clear what the metric sequences are coming from

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -118,19 +129,68 @@ private[op] class OpMultiClassificationEvaluator
...
+  def calculateTopKMetrics(
+    data: RDD[(Array[Double], Double)],
+    topKs: Seq[Int]
+  ): MultiClassificationMetricsTopK = {
+    type Label = Int
+
+    def computeMetrics(scoresAndLabels: (Array[Double], Double)): (Array[Double], Double) = {
+      val scores: Array[Double] = scoresAndLabels._1
+      val label: Label = scoresAndLabels._2.toInt
+      val topKsAndScores: Map[Label, Array[(Double, Int)]] = topKs.map(t => t -> scores.zipWithIndex.sortBy(-_._1)
```
Jauntbox, line 167: I don't think this is doing what we want.

This seems to be repeating the existing topN calculation in the threshold metrics (eg. allow a "correct" to correspond to the true label class being in the top 1, top 3, or top 10 of the predicted labels).

What we want is a measure of model metrics (error, precision, recall, f1) restricted to the topK most commonly occurring labels. Eg. what's the error if we only allow the top 10 most common labels to be predicted? The topN is always 1 here still - we only care about the prediction label with the highest score, but if it's outside the topK most common labels then it wouldn't count (even if it's a correct classification).

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -118,19 +129,64 @@ private[op] class OpMultiClassificationEvaluator
...
+ *
+ * @param data       Input RDD consisting of (vector of score probabilities, label), where label corresponds to the
+ *                   index of the true class and the score vector consists of probabilities for each class
+ * @param topKs      Sequence of topK values to calculate multiclass classification metrics for
+ */
+  def calculateTopKMetrics(
+    data: RDD[(Double, Double)],
+    topKs: Seq[Int]
+  ): MultiClassificationMetricsTopK = {
+    val labelCounts: RDD[(Double, Int)] = data.map { case (_, label) => (label, 1)}.reduceByKey(_ + _)
```
Jauntbox, line 166: Let's use a `Long` for the counts here just to be safe.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -271,7 +327,27 @@ case class MultiClassificationMetrics
...
+ *
+ * Each metric contains a list of metrics corresponding to each of the topK most occurring labels.  If the predicted
+ * label is outside of the topK most occurring labels, it is treated as incorrect.
+ *
+ * @param Precision
+ * @param Recall
+ * @param F1
+ * @param Error
+ */
+case class MultiClassificationMetricsTopK
```
Jauntbox, line 345: We also should keep track of the topKs here so that we don't need to get it from the evaluator params

Jauntbox (changes requested):
Method looks good - please add some tests that verify this and we're good to go

Jauntbox (approved):
LGTM - nice job!

Issue #514, opened 2020-09-23
Model saving and loading behavior changed since #475


koertkuipers:
**Describe the bug**
model.save and model.load behavior changes since #475
it seems relative paths and hadoop filesystem urls are no longer supported.

**To Reproduce**
on master branch make this change:
```
$ git diff
diff --git a/core/src/test/scala/com/salesforce/op/OpWorkflowTest.scala b/core/src/test/scala/com/salesforce/op/OpWorkflowTest.scala
index a89fcce2..3287e7a5 100644
--- a/core/src/test/scala/com/salesforce/op/OpWorkflowTest.scala
+++ b/core/src/test/scala/com/salesforce/op/OpWorkflowTest.scala
@@ -547,7 +547,7 @@ class OpWorkflowTest extends FlatSpec with PassengerSparkFixtureTest {
     val model = workflow.train()
     val expectedScoresDF = model.score()
     val expectedScores = expectedScoresDF.select(prediction.name, KeyFieldName).sort(KeyFieldName).collect()
-    model.save(workflowLocation)
+    model.save("testmodel")
 
     def assertModel(model: OpWorkflowModel): Assertion = {
       val scoresDF = model.setInputDataset(ds, keyFn).score()
@@ -557,10 +557,10 @@ class OpWorkflowTest extends FlatSpec with PassengerSparkFixtureTest {
     }
 
     withClue("Expected to load and score model with provided workflow: ") {
-      assertModel(model = workflow.loadModel(workflowLocation))
+      assertModel(model = workflow.loadModel("testmodel"))
     }
     withClue("Expected to load and score model without workflow: ") {
-      assertModel(model = OpWorkflowModel.load(workflowLocation))
+      assertModel(model = OpWorkflowModel.load("testmodel"))
     }
   }
```
now the test fails with a stacktrace that looks like this:
```
com.salesforce.op.OpWorkflowTest > OpWorkflow should train a model with features of all feature types, save, load and score it FAILED
    java.lang.NullPointerException
        at sun.nio.fs.UnixPath.normalizeAndCheck(UnixPath.java:77)
        at sun.nio.fs.UnixPath.<init>(UnixPath.java:71)
        at sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:281)
        at ml.combust.bundle.BundleFile$.apply(BundleFile.scala:59)
        at ml.combust.bundle.BundleFile$.apply(BundleFile.scala:40)
[...]

PR #516, opened 2020-10-02
Save model locally and convert to zip before moving to final path and do the reverse for loading

leahmcguire:
**Related issues**
The mleap bundle save works only with the local file system so workflow saves directly to hdfs fail after merging #475 
https://github.com/salesforce/TransmogrifAI/issues/514

Save models first to a local tmp file and then zip and move to final location. move zipped file to local and unzip for loading

**Additional context**


core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
```diff
@@ -220,8 +223,19 @@ class OpWorkflowModel(val uid: String = UID[OpWorkflowModel], val trainingParams
    * @param path      path to save the model
    * @param overwrite should overwrite if the path exists
    */
-  def save(path: String, overwrite: Boolean = true): Unit = {
-    OpWorkflowModelWriter.save(this, path = path, overwrite = overwrite)
+  def save(path: String, overwrite: Boolean = true,
+    localDir: String = "tmp/model"): Unit = {
```
tovbinm, line 227: 1. please add docs for `localDir`
2. better use something random here, e.g. `localDir: String = s"tmp/model-${System.currentTimeInMillis}"`

utils/src/main/scala/com/salesforce/op/utils/io/file/FileCompress.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.collection.JavaConverters._
+
+case object FileCompress {
+
+  /**
+   * Zip a directory
+   * @param inLoc location of directory to zip
+   * @param zipLoc output zip file name
+   */
+  def zip(inLoc: String, zipLoc: String): Unit = {
```
tovbinm, line 18: It could be beneficial to use [this](https://github.com/zeroturnaround/zt-zip) library instead. E.g
```scala
// Compress folder
ZipUtil.pack(new File("/tmp/demo"), new File("/tmp/demo.zip"))

// Decompress zip into a folder
ZipUtil.unpack(new File("/tmp/demo.zip"), new File("/tmp/demo"))
```

core/src/main/scala/com/salesforce/op/OpWorkflowModel.scala
gerashegalov: Consider calling it modelStagingDir. We don't need an explicit default, we can "borrow"/reflect into https://github.com/apache/spark/blob/065f17386d1851d732b4c1badf1ce2e14d0de338/core/src/main/scala/org/apache/spark/util/Utils.scala createTempDir and other goodies

utils/src/main/scala/com/salesforce/op/utils/io/file/FileCompress.scala
gerashegalov: The downsides are 
- it's yet another dependency to be vetted at Salesforce 
- and we lost the API that would allow to write the zip directly to the destination file system by providing an appropriate output stream [here](https://github.com/salesforce/TransmogrifAI/pull/516/commits/2b0ecd031294b9c699049b76356997e8bc85384d#diff-4fc4382e0d7736f51ed2d263d373ece5L22)

leahmcguire: Lets chat about this @gerashegalov

leahmcguire:
https://github.com/combust/mleap/issues/721

leahmcguire:
You want the link in the code?

tovbinm:
@leahmcguire It's fine to have it here.

TuanNguyen27 (approved):
LGTM

utils/src/main/scala/com/salesforce/op/utils/io/file/FileCompress.scala
nicodv: `java.util.zip` could be an alternative. zt-zip seems pretty light-weight, though.

nicodv (approved):
LGTM, but may want to avoid additional dependency

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -475,10 +475,17 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
...
+   * @param localDir local folder to copy and unpack stored model to for loading
+   *
    * @return workflow model
    */
-  def loadModel(path: String): OpWorkflowModel = {
-    new OpWorkflowModelReader(Some(this)).load(path)
+  def loadModel(
+    path: String,
+    asSpark: Boolean = true,
+    localDir: String = WorkflowFileReader.localDir
```
gerashegalov, line 486: just maybe call the parameter modelStagingDir too

gerashegalov (approved):
LGTM, minor comment

PR #517, opened 2020-10-13
Allow metadata to be preserved on features extracted from dataframes

leahmcguire:
**Related issues**

**Additional context**

### DIFF
--- a/features/src/main/scala/com/salesforce/op/features/Feature.scala
+++ b/features/src/main/scala/com/salesforce/op/features/Feature.scala
@@ -42,13 +42,15 @@ import scala.reflect.runtime.universe.WeakTypeTag
  *
  * Note: can only be created using [[FeatureBuilder]].
  *
- * @param name        name of feature, represents the name of the column in the dataframe.
- * @param isResponse  whether or not this feature is a response feature, ie dependent variable
- * @param originStage reference to OpPipelineStage responsible for generating the feature.
- * @param parents     references to the features that are transformed by the originStage that produces this feature
- * @param uid         unique identifier of the feature instance
- * @param wtt         feature's value type tag
- * @tparam O feature value type
+ * @param name          name of feature, represents the name of the column in the dataframe.
+ * @param isResponse    whether or not this feature is a response feature, ie dependent variable
+ * @param originStage   reference to OpPipelineStage responsible for generating the feature.
+ * @param parents       references to the features that are transformed by the originStage that produces this feature
+ * @param uid           unique identifier of the feature instance
+ * @param distributions distributions associated with raw feature from raw feature filter
+ * @param metadata      metadata to attach to feature in dataframe
+ * @param wtt           feature's value type tag
+ * @tparam O            feature value type
  */
 case class Feature[O <: FeatureType] private[op]
 (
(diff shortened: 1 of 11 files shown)


features/src/main/scala/com/salesforce/op/features/Feature.scala
```diff
@@ -107,6 +110,14 @@ case class Feature[O <: FeatureType] private[op]
    */
   override private[op] def withDistributions(distributions: Seq[FeatureDistributionLike]) =
     this.copy(distributions = distributions)
+
+  /**
+   * Adds metadata to feature so can override metadata on created feature
+   *
+   * @param metadataIn dataframe metadata to include in the
```
tovbinm, line 117: docs seem incomplete

features/src/main/scala/com/salesforce/op/features/FeatureJsonHelper.scala
```diff
@@ -114,6 +118,11 @@ object FeatureJsonHelper {
     val isResponse = (json \ "isResponse").extract[Boolean]
     val originStageUid = (json \ "originStage").extract[String]
     val parentUids = (json \ "parents").extract[Array[String]]
+    // TODO get this to work
```
tovbinm, line 121: what's the problem here?

readers/src/main/scala/com/salesforce/op/readers/DataReader.scala
```diff
@@ -179,21 +180,30 @@ trait DataReader[T] extends Reader[T] with ReaderKey[T] {
...
     rawData match {
       case Left(rdd) =>
-        val d = rdd.flatMap(record => generateRow(key(record), record, rawFeatures))
+        val d = rdd.flatMap(record => generateRow(key(record), record, rawFeatures, schema))
         spark.createDataFrame(d, schema)
       case Right(ds) =>
-        implicit val rowEnc = RowEncoder(schema)
-        ds.flatMap(record => generateRow(key(record), record, rawFeatures))
+        val inputSchema = ds.schema.fields
+         if (schema.forall(fn => inputSchema.exists(
```
tovbinm, line 187: does this check here validates that the dataframe schema exactly matches the input features?

readers/src/test/scala/com/salesforce/op/readers/CustomReaderTest.scala
```diff
@@ -0,0 +1,34 @@
+package com.salesforce.op.readers
```
tovbinm, line 1: The license header is missing

tovbinm (approved):
A great addition! LGTM

readers/src/main/scala/com/salesforce/op/readers/DataReader.scala
leahmcguire: it checks that the schema for all the input features matches that of the input dataframe features (there can be additional features in the dataframe). I will add a comment.

features/src/main/scala/com/salesforce/op/features/FeatureJsonHelper.scala
leahmcguire: the case class is not available in this module...I can move it to from core to features. though at that point I should probably get rid of the base class...so kind of a cascade of changes.

tovbinm: If you can move it without changing the package - do it.

leahmcguire: I will try it in a separate PR, 'cause I am on a deadline :-P It will not be moving just one file there are at least 3 other classes that would need to move to move FeatureDistribution

nicodv (approved):
LGTM

PR #521, opened 2020-10-21
Fix local file system handling

gerashegalov:
- fix  path is being deleted with via remote https://github.com/salesforce/TransmogrifAI/pull/516#discussion_r508854106
- replace manual path concatention


nicodv (approved):
LGTM, provided tests are fixed (I see "java.io.NotSerializableException: org.apache.hadoop.fs.RawLocalFileSystem")

PR #523, opened 2020-10-26
Integrate TransmogrifAI OSS with git2gus

gerashegalov:


tovbinm (approved):

tovbinm:
@gerashegalov this comment looks a bit weird though `Git2Gus App is installed but the .git2gus/config.json doesn't exists.`

gerashegalov:
it should be the last time we see it after this PR

PR #524, opened 2020-10-26
Changing blocklist policy for Sequence stages

michaelweilsalesforce:
**Related issues**
Example : If a SequenceEstimator/Transformer with input features `Seq(f1, f2, f3)` has a `f1` as a blocklist, then, because `Seq(f2, f3)` and `Seq(f1, f2, f3)` don't have the same size, the SequenceEstimator/Transformer will be removed when updating the DAG. 
However those sequence stages should ignore if one or more original inputs are missing.  

When updating the DAG, sequence stages with updated inputs with length different than 0 will be kept.

**Additional context**
This problem was addressed when we witnessed a result feature being part of the blocklist after updating the DAG. We acknowledge the possibility to change the policy in RawFeatureFilter.

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflow.scala
@@ -30,21 +30,22 @@
 
 package com.salesforce.op
 
-import com.salesforce.op.features.OPFeature
+import com.salesforce.op.features.{OPFeature, TransientFeature}
 import com.salesforce.op.filters.{FeatureDistribution, FilteredRawData, RawFeatureFilter, Summary}
 import com.salesforce.op.readers.Reader
 import com.salesforce.op.stages.OPStage
+import com.salesforce.op.stages.base.sequence.{SequenceEstimator, SequenceTransformer}
 import com.salesforce.op.stages.impl.feature.TimePeriod
 import com.salesforce.op.stages.impl.preparators.CorrelationType
 import com.salesforce.op.stages.impl.selector.ModelSelector
 import com.salesforce.op.utils.reflection.ReflectionUtils
 import com.salesforce.op.utils.spark.{JobGroupUtil, OpStep}
-import com.salesforce.op.utils.spark.RichDataset._
 import com.salesforce.op.utils.stages.FitStagesUtil
 import com.salesforce.op.utils.stages.FitStagesUtil.{CutDAG, FittedDAG, Layer, StagesDAG}
 import enumeratum.{Enum, EnumEntry}
 import org.apache.spark.annotation.Experimental
-import org.apache.spark.ml.{Estimator, Transformer}
+import org.apache.spark.ml.Transformer
+import org.apache.spark.sql.types.MetadataBuilder
 import org.apache.spark.sql.{DataFrame, SparkSession}
 
 import scala.collection.mutable.{MutableList => MList}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -149,7 +150,21 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
           }
         val inputsChanged = blocklistRemoved.map{ f => allUpdated.find(u => u.sameOrigin(f)).getOrElse(f) }
         val oldOutput = stg.getOutput()
-        Try(stg.setInputFeatureArray(inputsChanged).setOutputFeatureName(oldOutput.name).getOutput()) match {
+        Try(stg match {
+          case s: SequenceEstimator[_, _] if (inputsChanged.size > 0) => {
```
tovbinm, line 154: please create a helper function

tovbinm, line 154: `inputsChanged.size > 0` = `inputsChanged.nonEmpty`

michaelweilsalesforce:
My code still need some improvement. 
However do you folks agree with this change?

michaelweilsalesforce:
@nicodv @Jauntbox Please review. Thanks.

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -149,7 +150,16 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
           }
         val inputsChanged = blocklistRemoved.map{ f => allUpdated.find(u => u.sameOrigin(f)).getOrElse(f) }
         val oldOutput = stg.getOutput()
-        Try(stg.setInputFeatureArray(inputsChanged).setOutputFeatureName(oldOutput.name).getOutput()) match {
+        Try(stg match {
+          // For Sequence stages, we still want to keep them and remove the blocklisted inputs
+          case s @ (_ : SequenceEstimator[_, _] | _ : SequenceTransformer[_, _]) if !inputsChanged.isEmpty => {
+            s.set(s.inputFeatures, inputsChanged.map(TransientFeature(_))).setOutputFeatureName(oldOutput.name)
+            // Resetting Metadata
+            s.setMetadata(new MetadataBuilder().build())
```
leahmcguire, line 158: you cannot destroy the metadata on a sequence estimator like this - better call the set input and then call setMetadata(with s.getMetadata)

michaelweilsalesforce: That doesn't fix my issue. When setting the input a second time, it does not remove the metadata related to features that were blocklisted. This is a problem for stages that rely on column metadata E.g SanityChecker

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
```diff
@@ -149,7 +150,16 @@ class OpWorkflow(val uid: String = UID[OpWorkflow]) extends OpWorkflowCore {
           }
         val inputsChanged = blocklistRemoved.map{ f => allUpdated.find(u => u.sameOrigin(f)).getOrElse(f) }
         val oldOutput = stg.getOutput()
-        Try(stg.setInputFeatureArray(inputsChanged).setOutputFeatureName(oldOutput.name).getOutput()) match {
+        Try(stg match {
+          // For Sequence stages, we still want to keep them and remove the blocklisted inputs
+          case s @ (_ : SequenceEstimator[_, _] | _ : SequenceTransformer[_, _]) if !inputsChanged.isEmpty => {
```
nicodv, line 155: `!inputsChanged.isEmpty` => `inputsChanged.nonEmpty`

nicodv (commented):
Looks reasonable, but I don't have good knowledge of the consequences of changing the metadata.

core/src/main/scala/com/salesforce/op/OpWorkflow.scala
leahmcguire: if the problem is with a specific stage then I would suggest that you fix the getMetadata call to that stage so that it will recreate the metadata with the new input features taken into account - doing this will break everything that relies on metadata to work for example all the model info and feature importance caluculations

leahmcguire (changes requested):
Setting the metadata to empty is not the answer - you need to fix the metadata creation on the stage that is not working correctly

michaelweilsalesforce:
Weird. Not able to reproduce the bug anymore. It might have been fixed. Closing PR

PR #526, opened 2020-10-30
@W-8295960@ avoid repeated codec factory initialization

gerashegalov:
**Related issues**
#527 

move out as an object field with only default codecs (Gzip and deflate)

**Additional context**

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
@@ -264,6 +264,8 @@ private object WorkflowFileReader {
   val rawModel = "rawModel"
   val zipModel = "Model.zip"
   def modelStagingDir: String = s"modelStagingDir/model-${System.currentTimeMillis}"
+  val confWithDefaultCodec = new Configuration(false)
+  val codecFactory = new CompressionCodecFactory(confWithDefaultCodec)
 
   def loadFile(pathString: String)(implicit conf: Configuration): String = {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
```diff
@@ -268,6 +268,8 @@ private object WorkflowFileReader {
   val rawModel = "rawModel"
   val zipModel = "Model.zip"
   def modelStagingDir: String = s"modelStagingDir/model-${System.currentTimeMillis}"
+  val confWithDefaultCodec = new Configuration(false)
+  val codecFactory = new CompressionCodecFactory(confWithDefaultCodec)
```
tovbinm, line 272: Is it reusable? Is this thread-safe?

leahmcguire: model reading should be in the driver

leahmcguire (approved):

core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
gerashegalov: I think it is. 

The codec factory maps are fully populated after the constructor and val is final. 
Each usage of the WorkflowReader singleton in a different thread necessarily is synchronized with the classloader static class initialization. 

So there is proper JMM happens-before order between the consistent codec factory and the future getCodec lookups.

PR #528, opened 2020-11-11
Fix writing of model in the local file system

sanmitra:
**Related issues**
The model loading in a cluster was failing with error - *op-model.json/part-00000 did not exist.*

**Additional context**
In #516 we made the changes to save model locally and convert to zip before moving to remote path. So the path passed to 
the below method `saveImpl` is of a local file system, hence when we are trying to save it as a HadoopFile, *op-model.json/part-00000* is not being written out.
```
override protected def saveImpl(path: String): Unit = {
    JobGroupUtil.withJobGroup(OpStep.ModelIO) {
      sc.parallelize(Seq(toJsonString(path)), 1)
        .saveAsTextFile(OpWorkflowModelReadWriteShared.jsonPath(path), classOf[GzipCodec])
    }(this.sparkSession)
  }
```

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
@@ -31,15 +31,14 @@
 package com.salesforce.op
 
 import java.io.File
+import java.nio.charset.StandardCharsets
 
 import com.salesforce.op.features.FeatureJsonHelper
 import com.salesforce.op.filters.RawFeatureFilterResults
 import com.salesforce.op.stages.{OPStage, OpPipelineStageWriter}
-import com.salesforce.op.utils.spark.{JobGroupUtil, OpStep}
 import enumeratum._
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
-import org.apache.hadoop.io.compress.GzipCodec
 import org.apache.spark.ml.util.MLWriter
 import org.json4s.JsonAST.{JArray, JObject, JString}
 import org.json4s.JsonDSL._
@@ -59,11 +58,39 @@ class OpWorkflowModelWriter(val model: OpWorkflowModel) extends MLWriter {
 
   implicit val jsonFormats: Formats = DefaultFormats
 
+  protected var modelStagingDir: String = WorkflowFileReader.modelStagingDir
+
+  /**
+   * Set the local folder to copy and unpack stored model to for loading
+   */
+  def setModelStagingDir(localDir: String): this.type = {
+    modelStagingDir = localDir
+    this
+  }
+
   override protected def saveImpl(path: String): Unit = {
-    JobGroupUtil.withJobGroup(OpStep.ModelIO) {
-      sc.parallelize(Seq(toJsonString(path)), 1)
-        .saveAsTextFile(OpWorkflowModelReadWriteShared.jsonPath(path), classOf[GzipCodec])
-    }(this.sparkSession)
+    val conf = new Configuration()
+    val localFileSystem = FileSystem.getLocal(conf)
+    val localPath = localFileSystem.makeQualified(new Path(modelStagingDir))
+    localFileSystem.delete(localPath, true)
+    val raw = new Path(localPath, WorkflowFileReader.rawModel)
+
+    val rawPathStr = raw.toString
+    val modelJson = toJsonString(rawPathStr)
+    val jsonPath = OpWorkflowModelReadWriteShared.jsonPath(rawPathStr)
+    val os = localFileSystem.create(new Path(jsonPath))
+    try {
+      os.write(modelJson.getBytes(StandardCharsets.UTF_8.toString))
+    } finally {
+      os.close()
+    }
+
+    val compressed = new Path(localPath, WorkflowFileReader.zipModel)
+    ZipUtil.pack(new File(raw.toUri.getPath), new File(compressed.toUri.getPath))
+
+    val finalPath = new Path(path, WorkflowFileReader.zipModel)
+    val destinationFileSystem = finalPath.getFileSystem(conf)
+    destinationFileSystem.moveFromLocalFile(compressed, finalPath)
   }
 
   /**
@@ -207,21 +234,9 @@ object OpWorkflowModelWriter {
     overwrite: Boolean = true,
     modelStagingDir: String = WorkflowFileReader.modelStagingDir
   ): Unit = {
-    val localPath = new Path(modelStagingDir)
-    val conf = new Configuration()
-    val localFileSystem = FileSystem.getLocal(conf)
-    if (overwrite) localFileSystem.delete(localPath, true)
-    val raw = new Path(modelStagingDir, WorkflowFileReader.rawModel)
-
-    val w = new OpWorkflowModelWriter(model)
+    val w = new OpWorkflowModelWriter(model).setModelStagingDir(modelStagingDir)
     val writer = if (overwrite) w.overwrite() else w
-    writer.save(raw.toString)
-    val compressed = new Path(modelStagingDir, WorkflowFileReader.zipModel)
-    ZipUtil.pack(new File(raw.toString), new File(compressed.toString))
-
-    val finalPath = new Path(path, WorkflowFileReader.zipModel)
-    val destinationFileSystem = finalPath.getFileSystem(conf)
-    destinationFileSystem.moveFromLocalFile(compressed, finalPath)
+    writer.save(path)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
```diff
@@ -59,11 +57,36 @@ class OpWorkflowModelWriter(val model: OpWorkflowModel) extends MLWriter {
...
+
   override protected def saveImpl(path: String): Unit = {
-    JobGroupUtil.withJobGroup(OpStep.ModelIO) {
-      sc.parallelize(Seq(toJsonString(path)), 1)
-        .saveAsTextFile(OpWorkflowModelReadWriteShared.jsonPath(path), classOf[GzipCodec])
-    }(this.sparkSession)
+    val localPath = new Path(modelStagingDir)
+    val conf = new Configuration()
+    val localFileSystem = FileSystem.getLocal(conf)
+    if (shouldOverwrite) localFileSystem.delete(localPath, true)
```
Jauntbox, line 74: Is there a way to do the overwriting in an atomic transaction? Eg. so we can't get into a state where the old model got deleted, but the new model failed to save?

core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
```diff
@@ -59,11 +57,36 @@ class OpWorkflowModelWriter(val model: OpWorkflowModel) extends MLWriter {
...
+    val raw = new Path(localPath, WorkflowFileReader.rawModel)
+
+    val modelJson = toJsonString(raw.toString)
+    val jsonPath = OpWorkflowModelReadWriteShared.jsonPath(raw.toString)
+    val out = localFileSystem.create(new Path(jsonPath, "part-00000"), shouldOverwrite)
```
Jauntbox, line 79: It feels weird to need to hardcode the file name here to `part-00000`. Is there a way to let the hadoop filesystem determine the naming for us? Would we run into problems if the file is too large and hadoop wants to break it into many part-xxxxx files?

sanmitra: I initially thought that model reader needs the `part-0000` but looks like it does not https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala#L282. So, I will remove it and then it will be just `op-model.json`.

We are writing `modelJson` to a local file system, zipping it and then moving to remote s3 path. The bug here was that we were trying to write `modelJson` by doing
`sc.parallelize(Seq(modelJson)).saveAsTextFile(jsonPath)`

The above code will only work if the file is of the hadoop file system. So to write to local file system I am using standard file operations. So, there is no breaking it into many part-xxxx files as the `modelJson` is not big data.

core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
nicodv, line 74: I think this line is no longer necessary now that we have `val out = localFileSystem.create(new Path(jsonPath), shouldOverwrite)` a bit below.

core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
```diff
@@ -59,11 +57,36 @@ class OpWorkflowModelWriter(val model: OpWorkflowModel) extends MLWriter {
...
-    }(this.sparkSession)
+    val localPath = new Path(modelStagingDir)
+    val conf = new Configuration()
+    val localFileSystem = FileSystem.getLocal(conf)
+    if (shouldOverwrite) localFileSystem.delete(localPath, true)
+    val raw = new Path(localPath, WorkflowFileReader.rawModel)
+
+    val modelJson = toJsonString(raw.toString)
+    val jsonPath = OpWorkflowModelReadWriteShared.jsonPath(raw.toString)
+    val out = localFileSystem.create(new Path(jsonPath), shouldOverwrite)
```
nicodv, line 79: Should we set `shouldOverwrite` to `true`? If `destinationFileSystem.moveFromLocalFile(compressed, finalPath)` below ever fails to delete after copying we'll need manual intervention.

core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
```diff
@@ -59,11 +57,36 @@ class OpWorkflowModelWriter(val model: OpWorkflowModel) extends MLWriter {
...
+    val os = new BufferedOutputStream(out)
+    os.write(modelJson.getBytes("UTF-8"))
+    os.close()
```
nicodv, line 82: We need to make sure to close even in case of exceptions. Something along the lines of try...finally.

sanmitra: sure

core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
sanmitra: i think we should always do `localFileSystem.delete(localPath, true)` irrespective of `shouldOverwrite`. shouldOverwrite should be applicable to the remote path i.e the passed to `saveImpl` fn.

nicodv: 👍 Right, they're independent

core/src/main/scala/com/salesforce/op/OpWorkflowModelWriter.scala
sanmitra: This whole `saveImpl` function is called inside `save` function which takes cares of if destinationFileSystem.moveFromLocalFile(compressed, finalPath) ever fails to delete.

```
def save(path: String): Unit = {
    new FileSystemOverwrite().handleOverwrite(path, shouldOverwrite, sc)
    saveImpl(path)
  }


private[ml] class FileSystemOverwrite extends Logging {

  def handleOverwrite(path: String, shouldOverwrite: Boolean, sc: SparkContext): Unit = {
    val hadoopConf = sc.hadoopConfiguration
    val outputPath = new Path(path)
    val fs = outputPath.getFileSystem(hadoopConf)
    val qualifiedOutputPath = outputPath.makeQualified(fs.getUri, fs.getWorkingDirectory)
    if (fs.exists(qualifiedOutputPath)) {
      if (shouldOverwrite) {
        logInfo(s"Path $path already exists. It will be overwritten.")
        // TODO: Revert back to the original content if save is not successful.
        fs.delete(qualifiedOutputPath, true)
      } else {
        throw new IOException(s"Path $path already exists. To overwrite it, " +
          s"please use write.overwrite().save(path) for Scala and use " +
          s"write().overwrite().save(path) for Java and Python.")
      }
    }
  }
}

```

nicodv: I was referring to doing `localFileSystem.create(new Path(jsonPath), true)` -- which you already changed -- not the overwriting on the s3 path.

gerashegalov:
@ijeri can you give an example of the local path? Do you mean because the path is not absolute?

Saving RDD to a local path should work even if the local filesystem is not the default one:
```
scala> sc.hadoopConfiguration.set("fs.defaultFS", "hdfs://127.0.0.1/")
scala> sc.parallelize(Seq(1,2,3)).saveAsTextFile("file:///tmp/test5")
```

gerashegalov:
+1 to gzipping op-model json since we now zip the whole bundle

sanmitra:
@gerashegalov here is an example.- file:/mnt/yarn/usercache/automl_experimentation_user/appcache/application_1605059393023_53414/container_1605059393023_53414_01_000001/modelStagingDir/model-1605147909424/rawModel/op-model.json/

I tried saving the RDD but it didn't work, apparently only` _SUCCESS` file was written, and no `part-xxxx` files.


see logs -> http://emr-shared.dev.platform.einstein.com:19888/jobhistory/logs/ip-10-43-53-155.us-west-2.compute.internal:8041/container_1605059393023_53414_01_000001/container_1605059393023_53414_01_000001/automl_experimentation_user/stdout/?start=0


Screen Shot 2020-11-12 at 4 13 00 PM

sanmitra:
@gerashegalov the file path is a qualified one for ex- `file:/a.../b..../op-model.json`

gerashegalov:
makes sense: after rdd is distributed partitions are written out to the worker node local dirs.

core/src/test/scala/com/salesforce/op/OpWorkflowRunnerTest.scala
```diff
@@ -232,7 +232,12 @@ class OpWorkflowRunnerTest extends AsyncFlatSpec with PassengerSparkFixtureTest
       dirFile.isDirectory shouldBe true
       // TODO: maybe do a thorough files inspection here
       val files = FileUtils.listFiles(dirFile, null, true)
-      files.asScala.map(_.toString).exists(_.contains("_SUCCESS")) shouldBe true
+      if (outFile.getAbsolutePath.endsWith("/model")) {
```
gerashegalov, line 235: maybe:
```
      val fileNames = files.asScala.map(_.getName)
      if (outFile.getAbsolutePath.endsWith("/model")) {
        fileNames should contain ("op-model.json")
      }
      else {
        fileNames should contain ("_SUCCESS")
      }
```

gerashegalov (commented):
LGTM, some comments

sanmitra:
@gerashegalov I have addressed your comments. Please take a look again.

gerashegalov (approved):
LGTM

PR #530, opened 2020-11-17
Add confusion matrix metrics

feifjiang:
Added confusion matrix for multi-class classification, including:
- confusion matrix for the topK most occurring labels
- the topN predictions/labels for each label/prediction category

### DIFF
--- a/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
+++ b/core/src/main/scala/com/salesforce/op/OpWorkflowModelReader.scala
@@ -43,7 +43,7 @@ import com.salesforce.op.stages._
 import org.zeroturnaround.zip.ZipUtil
 import org.apache.commons.io.IOUtils
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.Path
+import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.hadoop.io.compress.CompressionCodecFactory
 import org.json4s.JsonAST.{JArray, JNothing, JValue}
 import org.json4s.jackson.JsonMethods.parse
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -94,6 +98,25 @@ private[op] class OpMultiClassificationEvaluator
 
   def setThresholds(v: Array[Double]): this.type = set(thresholds, v)
 
+  def setTopKCm(v: Int): this.type = set(topKCm, v)
```
Jauntbox, line 101: I think we should call these something different since we already have confusion between `topK` and `topN` in the multiclass metrics. How about a more self-documenting name like `confMatrixNumClasses` for `topKCm` and `confMatrixMinSupport` for `topNCm`?

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +160,200 @@ private[op] class OpMultiClassificationEvaluator
...
+ * function to find the lower bound of the confidence threshold bin in which the probability score falls.
+ * E.g. if thresholds(i) <= probability < thresholds(i+1), return thresholds(i)
+ *
+ * @param probability the probability score of the predicted class
+ * @param sortedThresholds thresholds sorted in ascending order
+ * @return lower bound of the
+ */
+  def findConfidenceThresholdBin(probability: Double, sortedThresholds: Seq[Double]): Double = {
+
+    if (probability < sortedThresholds(0)) {0.0}
```
Jauntbox, line 196: We shouldn't need to implement our own binary search here. If you cast `sortedThresholds` to an `indexedSeq`, then you can use `scala.collection.Searching` here: https://github.com/scala/scala/blob/v2.11.0-M3/src/library/scala/collection/Searching.scala#L61-L99

It can find both exact matches and insertion points for elements not in the thresholds.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -352,6 +560,56 @@ case class MultiClassificationMetricsTopK
...
+ * @param topK cutoff value for the topK most frequent labels
+ * @param thresholds thresholds for predicted probabilities. Values must be between 0.0 and 1.0
+ * @param confusionMatrixTopKByThreshold a 3D array that stores the counts. 1st dim is thresholds, 2nd dim is labels
+ *                                       and 3rd dim is predictions
+ */
+case class MulticlassConfusionMatrixTopKByThreshold
+(
+  topK: Int,
+  thresholds: Seq[Double],
+  confusionMatrixTopKByThreshold: Seq[Seq[Seq[Long]]]
```
Jauntbox, line 577: I think we should try and keep this close to the way Spark treats the multiclass confusion matrices, if possible. They use their `Matrix` type for the actual confusion matrix, see eg. https://github.com/apache/spark/blob/v2.4.0/mllib/src/main/scala/org/apache/spark/mllib/evaluation/MulticlassMetrics.scala#L67. I'd suggest a `Seq[Matrix]` type for the confusion matrices by threshold.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -112,7 +135,11 @@ private[op] class OpMultiClassificationEvaluator
       log.warn("The dataset is empty. Returning empty metrics.")
       MultiClassificationMetrics(0.0, 0.0, 0.0, 0.0,
         MulticlassThresholdMetrics(Seq.empty, Seq.empty, Map.empty, Map.empty, Map.empty),
-        MultiClassificationMetricsTopK(Seq.empty, Seq.empty, Seq.empty, Seq.empty, Seq.empty))
+        MultiClassificationMetricsTopK(Seq.empty, Seq.empty, Seq.empty, Seq.empty, Seq.empty),
+        MulticlassConfusionMatrixTopKByThreshold($(topKCm), $(thresholds), Seq.empty),
```
Jauntbox, line 139: I'm not sure we want to use the same set of thresholds for the topK confusion matrix as we do for the existing threshold metrics since they're just 3 * (topN = 3) sets of counts per threshold, while this may end up being somewhere in the 10*10 - 30*30 range for each threshold.

I think we should have a separate set of thresholds for the confusion matrices, and be pretty sparing with them.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: I tried returning a Matrix type as the confusion matrix, but it's causing Json deserialization error in both `MultiClassificationModelSelectorTest` and `ModelSelectorSummaryTest` (error msg: Caused by: com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of org.apache.spark.mllib.linalg.Matrix, problem: abstract types either need to be mapped to concrete types, have custom deserializer, or be instantiated with additional type information). Adding a Json serialization annotation doesn't help. I think it needs to be a concrete type in order to be extracted correctly.

In the updated PR, I keep the returned type as an Array. @Jauntbox

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +169,176 @@ private[op] class OpMultiClassificationEvaluator
         topKs = $(topKs)
       )
 
+      val rddCm = dataUse.select(col(labelColName), col(predictionColName), col(probabilityColName)).rdd.map{
+        case Row(label: Double, pred: Double, prob: DenseVector) => (label, pred, prob.toArray)
```
gerashegalov, line 173: consider staying in the Dataframe API as long as possible before resorting to RDD

PR #532, opened 2020-12-02
Add multi-class confusion matrix metrics

feifjiang:
Added confusion matrix calculation for multi-class classification, including:
1) confusion matrix for the top n most frequent labels
2) misclassification per class

PR #533, opened 2020-12-04
Add confusion matrix

feifjiang:
This PR replaces #532 #530

### DIFF
--- a/core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
+++ b/core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
@@ -210,11 +210,20 @@ private[op] class OpMultiClassificationEvaluator
     }
   }
 
-  object SearchHelper extends Serializable{
+  private[evaluators] object SearchHelper extends Serializable{
 
+    /**
+     * function to search the confidence threshold corresponding to a probability score
+     *
+     * @param arr a sorted array of confidence thresholds
+     * @param element the probability score to be searched
+     * @return the confidence threshold corresponding of the element. It equals to the element if there is an exact
+     *         match. Otherwise it's the element right before the insertion point.
+     */
     def findThreshold(arr: IndexedSeq[Double], element: Double): Double = {
+      require(!arr.isEmpty, "Array of confidence thresholds can't be empty!")
       if (element > arr.last) arr.last
-      else if (element < arr(0)) 0.0
+      else if (element < arr.head) 0.0
       else {
         val insertionPoint = new SearchImpl(arr).search(element).insertionPoint
         val insertionPointValue = arr(insertionPoint)
@@ -228,9 +237,8 @@ private[op] class OpMultiClassificationEvaluator
  * function to calculate confusion matrix for TopK most occurring labels by confidence threshold
  *
  * @param data RDD of (label, prediction, prediction probability vector)
- * @return a MulticlassConfMatrixMetricsByThreshold object
+ * @return a MulticlassConfMatrixMetricsByThreshold instance
 */
-
   def calculateConfMatrixMetricsByThreshold(
     data: RDD[(Double, Double, Array[Double])]): MulticlassConfMatrixMetricsByThreshold = {
 
@@ -251,7 +259,7 @@ private[op] class OpMultiClassificationEvaluator
       }
     }.reduceByKey(_ + _)
 
-    labelPredictionConfidenceCountRDD.cache()
+    labelPredictionConfidenceCountRDD.persist()
 
     val cmByThreshold = sortedThresholds.map( threshold => {
       val filteredRDD = labelPredictionConfidenceCountRDD.filter {
@@ -260,11 +268,13 @@ private[op] class OpMultiClassificationEvaluator
       constructConfusionMatrix(filteredRDD, cmClasses)
     })
 
+    labelPredictionConfidenceCountRDD.unpersist()
+
     MulticlassConfMatrixMetricsByThreshold(
-      confMatrixNumClasses = $(confMatrixNumClasses),
-      confMatrixClassIndices = cmClasses,
-      confMatrixThresholds = $(confMatrixThresholds),
-      confMatrices = cmByThreshold
+      ConfMatrixNumClasses = $(confMatrixNumClasses),
+      ConfMatrixClassIndices = cmClasses,
+      ConfMatrixThresholds = $(confMatrixThresholds),
+      ConfMatrices = cmByThreshold
     )
   }
 
@@ -272,7 +282,7 @@ private[op] class OpMultiClassificationEvaluator
  * function to calculate the mostly frequently mis-classified classes for each label/prediction category
  *
  * @param data RDD of (label, prediction)
- * @return a MisClassificationMetrics class
+ * @return a MisClassificationMetrics instance
  */
   def calculateMisClassificationMetrics(data: RDD[(Double, Double)]): MisClassificationMetrics = {
 
@@ -281,10 +291,10 @@ private[op] class OpMultiClassificationEvaluator
       .reduceByKey(_ + _)
 
     val misClassificationsByLabel = labelPredictionCountRDD.map {
-      case ((label, prediction), count) => (label, (prediction, count))
-    }.groupByKey()
+      case ((label, prediction), count) => (label, Seq((prediction, count)))
+    }.reduceByKey(_ ++ _)
       .map { case (label, predictionCountsIter) => {
-        val misClassificationCtMap = predictionCountsIter.toSeq
+        val misClassificationCtMap = predictionCountsIter
           .filter { case (pred, _) => pred != label }
           .sortBy(-_._2)
           .take($(confMatrixMinSupport)).toMap
@@ -295,19 +305,19 @@ private[op] class OpMultiClassificationEvaluator
           .reduceOption(_ + _).getOrElse(0L)
 
         MisClassificationsPerCategory(
-          category = label,
-          totalCount = labelCount,
-          correctCount = correctCount,
-          misClassifications = misClassificationCtMap
+          Category = label,
+          TotalCount = labelCount,
+          CorrectCount = correctCount,
+          MisClassifications = misClassificationCtMap
         )
       }
-    }.sortBy(-_.totalCount).collect()
+    }.sortBy(-_.TotalCount).collect()
 
     val misClassificationsByPrediction = labelPredictionCountRDD.map {
-      case ((label, prediction), count) => (prediction, (label, count))
-    }.groupByKey()
+      case ((label, prediction), count) => (prediction, Seq((label, count)))
+    }.reduceByKey(_ ++ _)
       .map { case (prediction, labelCountsIter) => {
-        val sortedMisclassificationCt = labelCountsIter.toSeq
+        val sortedMisclassificationCt = labelCountsIter
           .filter { case (label, _) => label != prediction }
           .sortBy(-_._2)
           .take($(confMatrixMinSupport)).toMap
@@ -318,13 +328,13 @@ private[op] class OpMultiClassificationEvaluator
           .reduceOption(_ + _).getOrElse(0L)
 
         MisClassificationsPerCategory(
-          category = prediction,
-          totalCount = predictionCount,
-          correctCount = correctCount,
-          misClassifications = sortedMisclassificationCt
+          Category = prediction,
+          TotalCount = predictionCount,
+          CorrectCount = correctCount,
+          MisClassifications = sortedMisclassificationCt
         )
       }
-    }.sortBy(-_.totalCount).collect()
+    }.sortBy(-_.TotalCount).collect()
 
     MisClassificationMetrics(
       ConfMatrixMinSupport = $(confMatrixMinSupport),
@@ -547,14 +557,14 @@ case class MultiClassificationMetricsTopK
  * @param confMatrixNumClasses value of the top n most occurring classes in the dataset
  * @param confMatrixClassIndices label index of the top n most occuring classes
  * @param confMatrixThresholds a sequence of thresholds
- * @param confMatrices a Matrix object that stores the confusion matrix for each threshold in confMatrixThresholds
+ * @param confMatrices a sequence of counts that stores the confusion matrix for each threshold in confMatrixThresholds
  */
 case class MulticlassConfMatrixMetricsByThreshold
 (
-  confMatrixNumClasses: Int,
-  confMatrixClassIndices: Seq[Double],
-  confMatrixThresholds: Seq[Double],
-  confMatrices: Seq[Seq[Long]]
+  ConfMatrixNumClasses: Int,
+  ConfMatrixClassIndices: Seq[Double],
+  ConfMatrixThresholds: Seq[Double],
+  ConfMatrices: Seq[Seq[Long]]
 ) extends EvaluationMetrics
 
 /**
@@ -580,11 +590,11 @@ case class MisClassificationMetrics
  */
 case class MisClassificationsPerCategory
 (
-  category: Double,
-  totalCount: Long,
-  correctCount: Long,
+  Category: Double,
+  TotalCount: Long,
+  CorrectCount: Long,
   @JsonDeserialize(keyAs = classOf[java.lang.Double])
-  misClassifications: Map[Double, Long]
+  MisClassifications: Map[Double, Long]
 )
 
 /**
--- a/core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
+++ b/core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
@@ -37,14 +37,18 @@ import com.twitter.algebird.Operators._
 import com.twitter.algebird.Tuple2Semigroup
 import com.salesforce.op.utils.spark.RichEvaluator._
 import org.apache.spark.ml.evaluation.MulticlassClassificationEvaluator
-import org.apache.spark.ml.linalg.Vector
-import org.apache.spark.ml.param.{DoubleArrayParam, IntArrayParam}
+import org.apache.spark.ml.linalg. {Vector, DenseVector}
+import org.apache.spark.ml.param.{DoubleArrayParam, IntArrayParam, IntParam, ParamValidators}
 import org.apache.spark.mllib.evaluation.MulticlassMetrics
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.functions.col
 import org.apache.spark.sql.types.DoubleType
 import org.apache.spark.sql.{Dataset, Row}
 import org.slf4j.LoggerFactory
+
+import scala.collection.immutable.HashSet
+import scala.collection.Searching._
+
 
 /**
  * Instance to evaluate Multi Classification metrics
@@ -94,6 +98,35 @@ private[op] class OpMultiClassificationEvaluator
 
   def setThresholds(v: Array[Double]): this.type = set(thresholds, v)
 
+  final val confMatrixNumClasses = new IntParam(
+    parent = this,
+    name = "confMatrixNumClasses",
+    doc = "# of the top most frequent classes used for confusion matrix metrics",
+    isValid = ParamValidators.inRange(1, 30, lowerInclusive = true, upperInclusive = true)
+  )
+  setDefault(confMatrixNumClasses, 10)
+
+  def setConfMatrixNumClasses(v: Int): this.type = set(confMatrixNumClasses, v)
+
+  final val confMatrixMinSupport = new IntParam(
+    parent = this,
+    name = "confMatrixMinSupport",
+    doc = "# of the top most frequent misclassified classes in each label/prediction category",
+    isValid = ParamValidators.inRange(1, 10, lowerInclusive = false, upperInclusive = true)
+  )
+  setDefault(confMatrixMinSupport, 5)
+
+  def setConfMatrixMinSupport(v: Int): this.type = set(confMatrixMinSupport, v)
+
+  final val confMatrixThresholds = new DoubleArrayParam(
+    parent = this,
+    name = "confMatrixThresholds",
+    doc = "sequence of threshold values used for confusion matrix metrics",
+    isValid = _.forall(x => x >= 0.0 && x <= 1.0)
+  )
+  setDefault(confMatrixThresholds, (0 to 10).map(_ / 10.0).toArray)
+  def setConfMatrixThresholds(v: Array[Double]): this.type = set(confMatrixThresholds, v)
+
   override def evaluateAll(data: Dataset[_]): MultiClassificationMetrics = {
     val labelColName = getLabelCol
     val dataUse = makeDataToUse(data, labelColName)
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +168,176 @@ private[op] class OpMultiClassificationEvaluator
...
+    val confusionMatrixMap = labelPredictionCtRDD.map {
+      case ((label, prediction, _), count) => ((label, prediction), count)
+    }.reduceByKey(_ + _)
+      .map {
+        case ((label, prediction), count) => (label, prediction) -> count
+      }.collectAsMap()
+
+    cmClasses.map(label => {
+      cmClasses.map(prediction => {
+        confusionMatrixMap.get((label, prediction)).getOrElse(0L)
```
gerashegalov, line 212: there is syntactic sugar for this 
```scala
confusionMatrixMap.getOrElse((label, prediction), 0L)
```

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +168,176 @@ private[op] class OpMultiClassificationEvaluator
...
+      })
+    }).flatten
```
gerashegalov, line 214: flatten of a map means we want a `flatMap` on Line 210 or we can use the for comprehension
```
    for {
      label <- cmClasses
      prediction <- cmClasses
    } yield {
      confusionMatrixMap.getOrElse((label, prediction), 0L)
    }
```

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +168,176 @@ private[op] class OpMultiClassificationEvaluator
...
+ * @param data RDD of (label, prediction, prediction probability vector)
+ * @return sequence of Matrix that corresponds to the confMatrixThresholds
+*/
+
+  def calculateConfMatrixMetricsByThreshold(
+    data: RDD[(Double, Double, Array[Double])]): MulticlassConfMatrixMetricsByThreshold = {
+
+    val labelCountsRDD = data.map { case (label, _, _) => (label, 1L) }.reduceByKey(_ + _)
+    val cmClasses = labelCountsRDD.sortBy(-_._2).map(_._1).take($(confMatrixNumClasses)).toSeq
+    val cmClassesSet = new HashSet() ++ cmClasses
```
gerashegalov, line 243: Let us do
```
val cmClassesSet = cmClasses.toSet
```

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +168,176 @@ private[op] class OpMultiClassificationEvaluator
...
+
+    val dataTopNLabels = data.filter { case (label, _, _) => cmClassesSet.contains(label) }
```
gerashegalov, line 245: Is this equivalent to?
```
    val dataTopNLabels = data.filter { case (label, prediction, _) => 
      cmClassesSet.contains(label) || cmClassesSet.contains(prediction)
    }
```

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +168,176 @@ private[op] class OpMultiClassificationEvaluator
...
+      case ((label, prediction), count) => (label, (prediction, count))
+    }.groupByKey()
+      .map { case (label, predictionCountsIter) => {
+        val misClassificationCtMap = predictionCountsIter.toSeq
+          .filter { case (pred, _) => pred != label }
+          .sortBy(-_._2)
+          .take($(confMatrixMinSupport)).toMap
+
+        val labelCount = predictionCountsIter.map(_._2).reduce(_ + _)
+        val correctCount = predictionCountsIter.filter {case (pred, _) => pred == label }
```
gerashegalov, line 296: filter + map usually calls for a single collect
```
        val correctCount = predictionCountsIter
          .collect { case (pred, count) if pred == label => count }
          .reduceOption(_ + _)
          .getOrElse(0L)
```

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +168,176 @@ private[op] class OpMultiClassificationEvaluator
...
+
+        MisClassificationsPerCategory(
+          category = prediction,
+          totalCount = predictionCount,
+          correctCount = correctCount,
+          misClassifications = sortedMisclassificationCt
+        )
+      }
+    }.collect()
+      .sortBy(-_.totalCount).toSeq
```
gerashegalov, line 332: can we reverse sortBy and collect to give Spark more work? 

toSeq is redundant

gerashegalov (changes requested):

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: It equals to &&. I like this simpler syntax - will change it

gerashegalov (approved):
LGTM, question re confMatrixThresholds parameter upper bound

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +166,173 @@ private[op] class OpMultiClassificationEvaluator
...
   }
 
+/**
+ * function to construct the confusion matrix for the top n most occurring labels
+ * @param labelPredictionCtRDD RDD of ((label, prediction, confidence), count)
+ * @param cmClasses the top n most occurring classes, sorted by counts in descending order
+ * @return an array of counts
+ */
+  def constructConfusionMatrix(
+    labelPredictionCtRDD: RDD[((Double, Double, Double), Long)],
```
tovbinm, line 198: Why not define a `case class LabelPredConf(label: Double, prediction: confidence: Double, count: Long)`?

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +166,173 @@ private[op] class OpMultiClassificationEvaluator
...
+
+    for {
+      label <- cmClasses
+      prediction <- cmClasses
+    } yield {
+      confusionMatrixMap.getOrElse((label, prediction), 0L)
+    }
+  }
+
+  object SearchHelper extends Serializable{
```
tovbinm, line 213: `private[evaluators] SearchHelper extends Serializable {`

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +166,173 @@ private[op] class OpMultiClassificationEvaluator
...
+
+    def findThreshold(arr: IndexedSeq[Double], element: Double): Double = {
+      if (element > arr.last) arr.last
```
tovbinm, line 216: 1. please check that arr is non empty
2. use arr.head instead of arr(0)

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +166,173 @@ private[op] class OpMultiClassificationEvaluator
...
+
+    val labelPredictionCountRDD = data.map {
+      case (label, prediction) => ((label, prediction), 1L) }
+      .reduceByKey(_ + _)
+
+    val misClassificationsByLabel = labelPredictionCountRDD.map {
+      case ((label, prediction), count) => (label, (prediction, count))
+    }.groupByKey()
+      .map { case (label, predictionCountsIter) => {
+        val misClassificationCtMap = predictionCountsIter.toSeq
```
tovbinm, line 287: Here and below - instead of iterating through predictionCountsIter twice for cases where `pred != label` and `pred == label`, better use `predictionCountsIter.partition(...)`

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +166,173 @@ private[op] class OpMultiClassificationEvaluator
...
+          misClassifications = misClassificationCtMap
+        )
+      }
+    }.sortBy(-_.totalCount).collect()
+
+    val misClassificationsByPrediction = labelPredictionCountRDD.map {
+      case ((label, prediction), count) => (prediction, (label, count))
+    }.groupByKey()
+      .map { case (prediction, labelCountsIter) => {
+        val sortedMisclassificationCt = labelCountsIter.toSeq
```
tovbinm, line 310: This code block is almost identical to the one in misClassificationsByLabel. Please introduce a local helper function to DRY it.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -352,6 +539,54 @@ case class MultiClassificationMetricsTopK
...
+ */
+case class MisClassificationMetrics
+(
+  ConfMatrixMinSupport: Int,
+  MisClassificationsByLabel: Seq[MisClassificationsPerCategory],
+  MisClassificationsByPrediction: Seq[MisClassificationsPerCategory]
+)
+
+/**
+ * Case class that stores the most frequently mis-classified classes for each label/prediction category
```
tovbinm, line 573: replace `Case class that stores` with `Container to store`

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -133,20 +166,173 @@ private[op] class OpMultiClassificationEvaluator
...
+        else arr(insertionPoint-1)
+      }
+    }
+  }
+
+/**
+ * function to calculate confusion matrix for TopK most occurring labels by confidence threshold
+ *
+ * @param data RDD of (label, prediction, prediction probability vector)
+ * @return a MulticlassConfMatrixMetricsByThreshold object
```
tovbinm, line 231: `@return a MulticlassConfMatrixMetricsByThreshold instance`

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -352,6 +539,54 @@ case class MultiClassificationMetricsTopK
...
+ * Metrics for multi-class confusion matrix. It captures confusion matrix of records of which
+ * 1) the labels belong to the top n most occurring classes (n = confMatrixNumClasses)
+ * 2) the top predicted probability exceeds a certain threshold in confMatrixThresholds
+ *
+ * @param confMatrixNumClasses value of the top n most occurring classes in the dataset
+ * @param confMatrixClassIndices label index of the top n most occuring classes
+ * @param confMatrixThresholds a sequence of thresholds
+ * @param confMatrices a Matrix object that stores the confusion matrix for each threshold in confMatrixThresholds
+ */
+case class MulticlassConfMatrixMetricsByThreshold
```
tovbinm, line 552: 1. Prefixing everything with `contMatrix` seems redundant 
2. Why is everything lowercased here? Why not:
```scala
case class MulticlassConfMatrixMetricsByThreshold
(
  NumClasses: Int,
  ClassIndices: Seq[Double],
  Thresholds: Seq[Double],
  ConfMatrices: Seq[Seq[Long]]
) extends EvaluationMetrics
```

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -352,6 +539,54 @@ case class MultiClassificationMetricsTopK
...
+ *
+ * @param category a category which a record's label or prediction equals to
+ * @param totalCount total # of records in that category
+ * @param correctCount # of correctly predicted records in that category
+ * @param misClassifications the top n most frequently misclassified classes (n = confMatrixMinSupport) and
+ *                           their respective counts in that category. Ordered by counts in descending order.
+ */
+case class MisClassificationsPerCategory
+(
+  category: Double,
```
tovbinm, line 583: same here - shouldn't the members be capitalized?

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -352,6 +539,54 @@ case class MultiClassificationMetricsTopK
...
+ * 2) the top predicted probability exceeds a certain threshold in confMatrixThresholds
+ *
+ * @param confMatrixNumClasses value of the top n most occurring classes in the dataset
+ * @param confMatrixClassIndices label index of the top n most occuring classes
+ * @param confMatrixThresholds a sequence of thresholds
+ * @param confMatrices a Matrix object that stores the confusion matrix for each threshold in confMatrixThresholds
+ */
+case class MulticlassConfMatrixMetricsByThreshold
+(
+  confMatrixNumClasses: Int,
```
tovbinm, line 554: Don't you need to add `@JsonDeserialize` annotations here?

local/src/test/scala/com/salesforce/op/local/OpWorkflowModelLocalTest.scala
```diff
@@ -123,7 +123,7 @@ class OpWorkflowModelLocalTest extends FlatSpec with TestSparkContext with TempD
     }
     log.info(s"Scored ${expectedScores.length * numOfRuns} records in ${elapsed}ms")
     log.info(s"Average time per record: ${elapsed.toDouble / (expectedScores.length * numOfRuns)}ms")
-    elapsed should be <= 25000L
+    elapsed should be <= 50000L
```
tovbinm, line 126: This change is not related. Let's omit this from this PR.

feifjiang: @tovbinm  this is a flapper test that sometimes fails with 25000L as the threshold. That's why I increased it to 50000L

tovbinm: Let's not double it. Let's increase it a little, e.g. 35000

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -122,9 +120,9 @@ private[op] class OpMultiClassificationEvaluator
     parent = this,
     name = "confMatrixThresholds",
     doc = "sequence of threshold values used for confusion matrix metrics",
-    isValid = _.forall(x => x >= 0.0 && x <= 1.0)
+    isValid = _.forall(x => x >= 0.0 && x < 1.0)
```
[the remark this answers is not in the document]
feifjiang: For multinomial logistic regression, the scenario where top probability equals to 1.0 will never happen. For random forest it's quite unlikely, unless the record falls onto one and the only one leaf node of the same class in every tree. Therefore it doesn't really make sense to have confidence threshold of 1.0

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: @tovbinm this case class is only used within this class for confusion matrix calculation and not sent outbound, so I didn't explicitly define it. In a future PR we can use Dataframe APIs/Row object instead of RDD tuples (suggestion from @gerashegalov ) in OpMultiClassificationEvaluator.scala, so that it's more explicit than the tuples and we don't have to declare it every time in the anonymous functions

tovbinm: this can be a private case class for this module.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: 1. There's already a `thresholds` defined for the threshold metrics, so adding a `confMatrix` prefix is necessary to avoid confusion. Also `confMatrixNumClasses` and `confMatrixClassIndices` specifically refer to the cutoff classes for confusion matrix calculation, so I think it's better to be explicit about them. Otherwise `numClasses` could mean number of classes for other metrics or for all the data.

2. Yes I can change all the element to uppercase. There doesn't seem to be a consistency of whether to use uppercase or camel case in the case classes that extends `EvaluationMetrics` here.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: @tovbinm  followed gera's suggestion and changed it to  `.collect { case (pred, count) if pred == label => count }`

tovbinm: ok.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: This case class will be further transformed in E1 before saving the metrics, so there doesn't seem to be a need to add an annotation here. The JsonDeserialize tag was added to the Map as `ModelSelectorSummaryTest` will break without it.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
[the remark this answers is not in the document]
gerashegalov: just wondering how much value is there to restrict it here?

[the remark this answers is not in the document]
feifjiang: @gerashegalov are you concerned about too many values present in the `confMatrixThresholds` array? That's a valid concern, I can add a check for that. It's a

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
tovbinm: I think it's still a good idea to add deser annotations, in case someone would need to parse json into classes later on.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: @tovbinm  Changed case class data fields name to upper case. Please see updated PR

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: @tovbinm changed the members to capitalized.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: @tovbinm added @JsonDeserialize annotation to both `ConfMatrixClassInidces` and `ConfMatrixThresholds` in the updated PR.

tovbinm (approved):

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -597,6 +600,13 @@ case class MisClassificationsPerCategory
   MisClassifications: Map[Double, Long]
 )
 
+case class labelPredictionConfidence
```
tovbinm, line 603: @feifjiang are you planning to use LabelPredictionConfidenceCt and labelPredictionConfidence somewhere?

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: If defining a case class `LabelPredConf(label: Double, prediction: confidence: Double, count: Long)`, it would require an additional map function to map the key-val pairs to this case class after the ```    val labelPredictionConfidenceCountRDD = dataTopNLabels.map{
      case (label, prediction, proba) => {
        ( (label, prediction, SearchHelper.findThreshold(sortedThresholds, proba.max)), 1L )
      }
    }.reduceByKey(_ + _) ```

And in then in `constructConfusionMatrix`, it will be `    val confusionMatrixMap = labelPredictionCtRDD.map( e => ((e.Label, e.Prediction), e.Count)).reduceByKey(_ + _).collectAsMap() `

tovbinm: It's not mandatory. If it only scoped within one function / file - no need to create a case class.

feifjiang: If defining a case class here, the new constructConfusionMatrix would look like 

 ```
 def constructConfusionMatrix(
    labelPredictionCtRDD: RDD[LabelPredictionConfidenceCt],
    cmClasses: Seq[Double]): Seq[Long] = {

    val confusionMatrixMap = labelPredictionCtRDD.map( r => ((r.Label, r.Prediction), r.Count))
      .reduceByKey(_ + _).collectAsMap()
```
And an additional map function will be needed  after `reduceByKey` in 

```
    val labelPredictionConfidenceCountRDD = dataTopNLabels.map{
      case (label, prediction, proba) => {
        ( (label, prediction, SearchHelper.findThreshold(sortedThresholds, proba.max)), 1L )
      }
    }.reduceByKey(_ + _)
      .map {
        case ((label, pred, confidence), ct) => LabelPredictionConfidenceCt(label, pred, confidence, ct)
      }
```
This case class is only used twice in the entire class. It doesn't seem to make the code more succinct. What would be the benefit of doing it? @tovbinm

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
feifjiang: @tovbinm  sorry - forgot to remove it. Just updated the PR again.

feifjiang:
@tovbinm @gerashegalov Can either one of you merge my PR with master? I don't have write access.

tovbinm:
@feifjiang thank you for your contribution!

feifjiang:
@tovbinm  Thank you for the comments and being so responsive!!

PR #534, opened 2020-12-09
@W-8522881@ Invalidate input URL if it contains special characters

WintersLu:
Due to latest java.net.URL security constrain, we invalidate URL containing special characters, for example '＠', this issue should be addressed globally and properly by normalizing special characters into single form.

### DIFF
--- a/features/src/main/scala/com/salesforce/op/features/types/Text.scala
+++ b/features/src/main/scala/com/salesforce/op/features/types/Text.scala
@@ -39,6 +39,8 @@ import org.apache.commons.httpclient.URI
 import org.apache.commons.io.input.CharSequenceInputStream
 import org.apache.commons.validator.routines.UrlValidator
 
+import scala.util.Try
+
 /**
  * Text value representation
  *
@@ -172,7 +174,7 @@ class URL(value: Option[String]) extends Text(value){
    * RFC2396 (http://www.ietf.org/rfc/rfc2396.txt)
    * Default valid protocols are: http, https, ftp.
    */
-  def isValid: Boolean = value.exists(UrlValidator.getInstance().isValid)
+  def isValid: Boolean = value.exists(v => UrlValidator.getInstance().isValid(v) && Try(new java.net.URL(v)).isSuccess)
   /**
    * Verifies if the url is of correct form of "Uniform Resource Identifiers (URI): Generic Syntax"
    * RFC2396 (http://www.ietf.org/rfc/rfc2396.txt)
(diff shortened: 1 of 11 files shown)


features/src/main/scala/com/salesforce/op/features/types/Text.scala
```diff
@@ -181,11 +182,11 @@ class URL(value: Option[String]) extends Text(value){
   /**
    * Extracts url domain, i.e. 'salesforce.com', 'data.com' etc.
    */
-  def domain: Option[String] = value map (new java.net.URL(_).getHost)
+  def domain: Option[String] = value map (s => new java.net.URL(new URI(s, false).toString).getHost)
   /**
    * Extracts url protocol, i.e. http, https, ftp etc.
    */
-  def protocol: Option[String] = value map (new java.net.URL(_).getProtocol)
+  def protocol: Option[String] = value map (s => new java.net.URL(new URI(s, false).toString).getProtocol)
```
tovbinm, line 189: why do you need to escape the url when extracting protocol?

features/src/main/scala/com/salesforce/op/features/types/Text.scala
```diff
@@ -181,11 +182,11 @@ class URL(value: Option[String]) extends Text(value){
   /**
    * Extracts url domain, i.e. 'salesforce.com', 'data.com' etc.
    */
-  def domain: Option[String] = value map (new java.net.URL(_).getHost)
+  def domain: Option[String] = value map (s => new java.net.URL(new URI(s, false).toString).getHost)
```
tovbinm, line 185: I would rather make it a parameter with a default value, e.g.
```scala
/**
 * Extracts url domain, i.e. 'salesforce.com', 'data.com' etc.
 *
 * @param escaped true if URI character sequence is in escaped form. false otherwise. 
 */ 
def domain(escaped: Boolean = false): Option[String] = value map (s => new java.net.URL(new URI(s, escaped).toString).getHost)
```

features/src/main/scala/com/salesforce/op/features/types/Text.scala
WintersLu: cuz the way how we extracting protocol here always throw exception if any part of input contains special character

tovbinm: Oh, got it.

WintersLu: correct myself, it happens when host part of URL contains special character, i fixed this in test case

gerashegalov:
I still don't see an exception:
```
scala> val url = new java.net.URL("http://specialchars.＠.com/")
url: java.net.URL = http://specialchars.＠.com/

scala> url.getHost
res5: String = specialchars.＠.com
```

WintersLu:
Looks like a JDK related issue, with `openjdk version "1.8.0_192"` issue cannot be reproduced, however with `openjdk version "1.8.0_232"` issue can be reproduced on local.  Thanks to @gerashegalov checking this

gerashegalov:
it looks like it was a CVE patch: http://hg.openjdk.java.net/jdk-updates/jdk11u/rev/2c5fa7382e75

tovbinm (approved):

gerashegalov:
instead of working around the security patch I suggest we simply return None for illegal URI.

tovbinm:
@gerashegalov that might be hiding the problem under the carpet. None would be indistinguishable from a `null` value in this case, or any other illegal URL. I this particular case the solution exists.

gerashegalov:
> @gerashegalov that might be hiding the problem under the carpet. None would be indistinguishable from a `null` value in this case, or any other illegal URL. I this particular case the solution exists.

I am not convinced that replacing the URL parsing library is an acceptable solution in this case.

WintersLu:
@tovbinm  I traced to this change in [jdk](http://hg.openjdk.java.net/jdk-updates/jdk11u/rev/2c5fa7382e75#l3.182), and also the reason it is been added. Line http://hg.openjdk.java.net/jdk-updates/jdk11u/rev/2c5fa7382e75#l3.39 declared there is a CRLF concern has been masked out. 
@gerashegalov I can return None, but still, its not given reasonable response and hiding the problem

tovbinm:
@WintersLu +1

gerashegalov:
This PR in the current form "sweeps the problem under the rug", does it not?

crupley (approved):

gerashegalov:
note that implementation no longer match the PR description

WintersLu:
@crupley @gerashegalov To merge and close this PR, i have created an issue (feature request) https://github.com/salesforce/TransmogrifAI/issues/539 to log additional works that would handle string/url input in this bug, by replacing special characters with standard form.

tovbinm:
Thank you.

PR #537, opened 2021-01-05
Add multiclassification topk and confmatrix metrics to model insights serialization format

feifjiang:


core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
```diff
@@ -232,8 +232,8 @@ class RecordInsightsLOCOTest extends FunSpec with TestSparkContext with RecordIn
       info("Each feature vector should only have either three or four non-zero entries. One each from country and " +
         "picklist, while currency can have either two (if it's null the currency column will be filled with the mean)" +
         " or just one if it's not null.")
-      it("should pick between 1 and 4 of the features") {
-        all(parsed.map(_.size)) should (be >= 1 and be <= 4)
+      it("should pick between 0 and 4 of the features") {
```
tovbinm, line 235: why did this change?

feifjiang: @tovbinm  this test is quite flaky. In one of the failed runs, the value was 0 and not between 1 and 4.

tovbinm: @leahmcguire @Jauntbox do you have any ideas?

WintersLu:
@crupley @nicodv include someone from the team to prioritize review this PR, changes in this PR will fix test that's currently blocking https://github.salesforceiq.com/einstein/einstein1/pull/4010 for bumping version in automl

core/src/test/scala/com/salesforce/op/stages/impl/insights/RecordInsightsLOCOTest.scala
nicodv: There's still an open PR to address this: https://github.com/salesforce/TransmogrifAI/pull/423 

We could use a better test, but for now we can start by getting rid of the flakiness.

nicodv (approved):
LGTM

PR #542, opened 2021-02-18
Get Hadoop config from Spark session

nicodv:
**Related issues**
N/A

We were getting a fresh Hadoop configuration, instead of getting it from the SparkSession. Hence, it will miss out on any configurations that were set.

N/A

**Additional context**
N/A


tovbinm (approved):
lgtm!

PR #545, opened 2021-03-05
Update confusion_matrix

swkasula:
**Related issues**
De-serializing the ConfMatrices: Seq[Seq[Long]] resulting in ClassCastException as there was no @JsonDeserialize 
 annotation used for this attribute: https://github.com/salesforce/TransmogrifAI/pull/533/files

**Describe the proposed solution**
Fix is it update the ConfMatrices format from Seq[Seq[Long]] -> Seq[ConfusionMatrixPerThreshold] to de-serialize in the right format

PR #549, opened 2021-03-11
Confusion matrix branch

swkasula:
Related issues
De-serializing the ConfMatrices: Seq[Seq[Long]] resulting in ClassCastException as there was no @JsonDeserialize
annotation used for this attribute: https://github.com/salesforce/TransmogrifAI/pull/533/files

Describe the proposed solution
Fix is it update the ConfMatrices format from Seq[Seq[Long]] -> Seq[ConfusionMatrixPerThreshold] to de-serialize in the right format

Copy of: https://github.com/salesforce/TransmogrifAI/pull/545


core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
```diff
@@ -596,6 +599,13 @@ case class MisClassificationsPerCategory
   MisClassifications: Seq[ClassCount]
 )
 
+case class ConfusionMatrixPerThreshold
+(
+  Threshold: Double,
+  @JsonDeserialize(contentAs = classOf[java.lang.Long])
+  ConfusionMatrixCounts: Seq[Long]
```
crupley, line 606: lowerCamelCase on these 2 params

feifjiang:
lgtm @swkasula

crupley:
@swkasula can we add a test to confirm the serialization/deserialization is working? That would probably belong in the ModelSelectorSummaryTest

swkasula:
> probably

that's a good point, will add it.

swkasula:
> @swkasula can we add a test to confirm the serialization/deserialization is working? That would probably belong in the ModelSelectorSummaryTest

That's a good point, will add it.

core/src/main/scala/com/salesforce/op/evaluators/OpMultiClassificationEvaluator.scala
swkasula: Seems like that there was a suggestion in the past to use the upper camel case to follow conventions in that file, so leaving as-is.

crupley (approved):
LGTM

swkasula:
@crupley @tovbinm @nicodv Could you please assist how these checks can be completed: 
security/snyk (Salesforce OSS )
license/snyk (Salesforce OSS

nicodv (approved):
LGTM

amateurhuman:
@swkasula I removed the Snyk checks from the repo which cleared them from this PR

Issue #336, opened 2019-06-18
Support Spark 3.0.0


koertkuipers:
Spark 3.0.0 is actively being developed (its the current master branch for Spark).

I know its early days but i figured i start commenting here on what i ran in to so far to safe others time in the future.

My goal is to get a branch working against Spark 3.0.0 when it comes out or a Spark 3.0.0-SNAPSHOT before that. The intention is not to get this merged in: I assume Transmogrify will wait for Spark 3.0.1 before even considering that.

This builds on https://github.com/salesforce/TransmogrifAI/issues/184 and https://github.com/salesforce/TransmogrifAI/issues/332

koertkuipers:
https://issues.apache.org/jira/browse/SPARK-26133
in spark 3 OneHotEncoderEstimator was renamed to OneHotEncoder

koertkuipers:
Jackson has to be upgraded to 2.9.8. EitherModule was moved.
See also https://github.com/salesforce/TransmogrifAI/issues/162 and https://github.com/salesforce/TransmogrifAI/issues/173

koertkuipers:
https://issues.apache.org/jira/browse/SPARK-26127
Spark removed a bunch of setters on regression and classification models. Since in TransmogrifAI
these are overridden this no longer compiles.

koertkuipers:
in-house we are now using/testing transmogrifai with spark 3.0.0 snapshots. all transmogrifai tests pass. 
i did run into issues that are perhaps not covered by tests. for example BinaryClassificationModelSelector throws an error in the titanic example... probably due to some changes in the params of spark-ml models it uses that i haven't tracked down yet.

koertkuipers:
it seems OpDefaultScalaModule is broken with jackson 2.9.9 (which is what spark 3 uses)
this took me a bit of time to track down because the error was hidden in `JsonUtils.fromString` which uses a Try/Recover that recovers from all errors masking the issue. 

anyhow this issue:
```
scala> val m = new ObjectMapper()
scala> m.registerModule(OpDefaultScalaModule)
scala> m.readValue("""{"a": "b"}""", classTag[Map[String, String]].runtimeClass)
java.lang.NullPointerException
  at com.fasterxml.jackson.databind.type.TypeBindings.<init>(TypeBindings.java:61)
  at com.fasterxml.jackson.databind.type.TypeBindings.createIfNeeded(TypeBindings.java:191)
[...]

PR #550, opened 2021-03-18
[WIP] Scala 2.12 / Spark 3 upgrade

nicodv:
**Related issues**
https://github.com/salesforce/TransmogrifAI/issues/336
https://github.com/salesforce/TransmogrifAI/issues/332

Upgrade to Scala 2.12 and Spark 3

Living in the past, suffering from security issues and missing out on feature and speed improvements

**Additional context**


readers/src/main/scala/com/salesforce/op/readers/JoinedDataReader.scala
```diff
@@ -346,6 +346,8 @@ private[op] class JoinedAggregateDataReader[T, U]
 }
 
 
+// TODO: UserDefinedAggregateFunction is now deprecated in favor of Aggregator,
```
nicodv, line 349: @tovbinm / @leahmcguire , I could use some pointers/help here. Since this is largely unused anyway, I'm even considering dropping the join functionality.

build.gradle
```diff
@@ -7,6 +7,7 @@ buildscript {
     dependencies {
         classpath 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.11:1.0.1'
         classpath 'com.commercehub.gradle.plugin:gradle-avro-plugin:0.16.0'
+        classpath 'com.adtran:scala-multiversion-plugin:1.+'
```
nicodv, line 10: Initially I wanted to go with cross-compilation for 2.11/2.12, but now that we're upgrading to Spark 3 too we'll only build for 2.12.

This cross-compilation plugin seems to work well, though, so keeping it around for future 2.13 support.

core/build.gradle
```diff
@@ -21,13 +21,13 @@ dependencies {
...
     // Scopt
-    compile "com.github.scopt:scopt_$scalaVersion:$scoptVersion"
+    compile "com.github.scopt:scopt_%%:$scoptVersion"
 
     // Zip util
     compile 'org.zeroturnaround:zt-zip:1.14'
 
     // XGBoost
-    compile ("ml.dmlc:xgboost4j-spark:$xgboostVersion") { exclude group: 'com.esotericsoftware.kryo', module: 'kryo' }
+    compile ("ml.dmlc:xgboost4j-spark_%%:$xgboostVersion") { exclude group: 'com.esotericsoftware.kryo', module: 'kryo' }
```
TuanNguyen27, line 30: I think we need both `xgboost4j` and `xgboost4j-spark` here, [according to this](https://github.com/dmlc/xgboost/tree/master/jvm-packages#add-maven-dependency). However, can we also use the gpu-enabled version of xgboost now ? The artifact names are `xgboost4j-gpu_2.12` and `xgboost4j-spark-gpu_2.12`

nicodv: `xgboost4j` already gets pulled in, but let me add it to be explicit.

W.r.t. GPUs: let's leave that baby alone for now and get this upgrade done first. 😅

readers/src/main/scala/com/salesforce/op/readers/JoinedDataReader.scala
tovbinm: Good idea. I would also use this as an opportunity to get rid of other unused code, e.g cli with codegen.

build.gradle
```diff
@@ -80,15 +80,15 @@ configure(allProjs) {
...
         pegdownVersion = '1.4.2'
         commonsValidatorVersion = '1.6'
         commonsIOVersion = '2.6'
         scoveragePluginVersion = '1.3.1'
-        xgboostVersion = '0.90'
-        akkaSlf4jVersion = '2.3.11'
-        mleapVersion = '0.16.0'
+        xgboostVersion = '1.3.1'
+        akkaSlf4jVersion = '2.5.23'
+        mleapVersion = '0.16.0' // TODO: upgrade to Spark 3-compatibel 0.17 when ready: https://github.com/combust/mleap/issues/727
```
tovbinm, line 91: typo

build.gradle
```diff
@@ -149,7 +149,6 @@ configure(allProjs) {
                 "-language:implicitConversions", "-language:existentials", "-language:postfixOps"
         ]
     }
-    compileScala.scalaCompileOptions.additionalParameters += "-optimize"
```
tovbinm, line 152: why remove optimization option?

features/build.gradle
```diff
@@ -16,11 +16,12 @@ dependencies {
...
     // Json4s extensions (needed for Joda time)
-    compile "org.json4s:json4s-ext_$scalaVersion:$json4sVersion"
+    compile "org.json4s:json4s-ext_%%:$json4sVersion"
 
     // MLeap serialization & runtime for Spark models
-    compile "ml.combust.mleap:mleap-spark_$scalaVersion:$mleapVersion"
-    compile "ml.combust.mleap:mleap-runtime_$scalaVersion:$mleapVersion"
-    compile "ml.combust.mleap:mleap-xgboost-spark_$scalaVersion:$mleapVersion"
-    compile "ml.combust.mleap:mleap-xgboost-runtime_$scalaVersion:$mleapVersion"
+    // TODO: upgrade 2.11 to %% when 0.17 is out
```
tovbinm, line 22: Please reference the ticket - https://github.com/combust/mleap/issues/727

features/src/main/scala/com/salesforce/op/features/FeatureSparkTypes.scala
```diff
@@ -263,12 +263,7 @@ case object FeatureSparkTypes {
    */
   def udf1[I <: FeatureType : TypeTag, O <: FeatureType : TypeTag](
     f: I => O
-  ): UserDefinedFunction = {
-    val inputTypes = Some(FeatureSparkTypes.sparkTypeOf[I] :: Nil)
-    val outputType = FeatureSparkTypes.sparkTypeOf[O]
-    val func = transform1[I, O](f)
-    UserDefinedFunction(func, outputType, inputTypes)
-  }
+  ): UserDefinedFunction = udf(transform1[I, O](f))
```
tovbinm, line 266: I think there was a reason why udf-s were done this way, but I dont recall the reason. I guess it will either work, or you would discover why ;-)

tovbinm (commented):
Thank you for the effort!

Did anything change in model serialization? I wonder how serialization of transformers would work.

build.gradle
nicodv: `-optimize` is deprecated. I've now added the new optimization flags that replace it.

core/src/test/scala/com/salesforce/op/filters/FeatureDistributionTest.scala
```diff
@@ -192,7 +192,7 @@ class FeatureDistributionTest extends FlatSpec with PassengerSparkFixtureTest wi
     val fd2 = FeatureDistribution("A", None, 20, 20, Array(2, 8, 0, 0, 12), Array.empty)
     fd1.hashCode() shouldBe fd1.hashCode()
     fd1.hashCode() shouldBe fd1.copy(summaryInfo = fd1.summaryInfo).hashCode()
-    fd1.hashCode() should not be fd1.copy(summaryInfo = Array.empty).hashCode()
+    fd1.hashCode() shouldBe fd1.copy(summaryInfo = Array.empty).hashCode()
```
emitc2h, line 195: @tovbinm I just want to make sure this is correct. In principle hashCode equality and equals should be consistent and this is what I'm trying to accomplish here, but I figured you might have had a reason for wanting something different.

tovbinm: I think this test was invalid.

leahmcguire (commented):
What is the reason for deleting all the join functionality from data readers in this PR?

tovbinm:
@leahmcguire I think it just not being used - https://github.com/salesforce/TransmogrifAI/pull/550#discussion_r597091426

leahmcguire:
We should be careful in how we define unused in a public project. Also that functionality would be needed to migrate projects on Transmogrifai V0...

readers/src/main/scala/com/salesforce/op/readers/JoinedDataReader.scala
nicodv: @leahmcguire ^^^

Besides the immediate problems of upgrading the UDAF approach, we're trying to reduce the burden of maintenance a bit here too . And since we're all in on Spark SQL on our end, we have no need for it anymore.

At this point, it's still open for discussion/contributions, though.

nicodv: @emitc2h , we could look into hacking it like we did here: https://github.com/salesforce/TransmogrifAI/pull/550/commits/a946ffb01e6eb7a1acd5636774ed19967f9479f4

emitc2h:
Hey @tovbinm,there's a unit test failure I've been investigating that's the result of a bug in Spark: https://issues.apache.org/jira/browse/SPARK-34805?focusedCommentId=17337491&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17337491.

I'm wondering why `testData` in `SanityCheckerTest.scala` ([L137](https://github.com/salesforce/TransmogrifAI/blob/90a0f298f14506a27c84a71de414d53a30cf687f/core/src/test/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerTest.scala#L137)) is constructed the way it is, with the metadata for the features column added manually. The fact that the metadata isn't passed along in `DataFrame.select` anymore is discovered by [this assertion](https://github.com/salesforce/TransmogrifAI/blob/90a0f298f14506a27c84a71de414d53a30cf687f/features/src/main/scala/com/salesforce/op/test/OpPipelineStageSpec.scala#L159).

I'm assuming Spark won't fix this any time soon, and I'm having trouble finding an alternative way of putting in the metadata in the schema of `testData`. I've tried `.withColumn`, but it still relies on `.select` under the hood. What's your take on this?

emitc2h:
Hey @tovbinm,there's a unit test failure I've been investigating that's the result of a bug in Spark: https://issues.apache.org/jira/browse/SPARK-34805?focusedCommentId=17337491&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17337491.

I'm wondering why `testData` in `SanityCheckerTest.scala` ([L137](https://github.com/salesforce/TransmogrifAI/blob/90a0f298f14506a27c84a71de414d53a30cf687f/core/src/test/scala/com/salesforce/op/stages/impl/preparators/SanityCheckerTest.scala#L137)) is constructed the way it is, with the metadata for the features column added manually. The fact that the metadata isn't passed along in `DataFrame.select` anymore is discovered by [this assertion](https://github.com/salesforce/TransmogrifAI/blob/90a0f298f14506a27c84a71de414d53a30cf687f/features/src/main/scala/com/salesforce/op/test/OpPipelineStageSpec.scala#L159).

I'm assuming Spark won't fix this any time soon, and I'm having trouble finding an alternative way of putting in the metadata in the schema of `testData`. I've tried `.withColumn`, but it still relies on `.select` under the hood. What's your take on this?

nicodv:
Also pinging @Jauntbox (we know you're out there!) for question above.

tovbinm:
This is a known issue indeed. We have been copying over the metadata between fields each time we apply our transformers, e.g [`OpTransformer1.transform`](https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/stages/base/unary/UnaryTransformer.scala#L79)

emitc2h:
> This is a known issue indeed. We have been copying over the metadata between fields each time we apply our transformers, e.g [`OpTransformer1.transform`](https://github.com/salesforce/TransmogrifAI/blob/master/features/src/main/scala/com/salesforce/op/stages/base/unary/UnaryTransformer.scala#L79)

I mean that there is a new problem with Spark 3.1. Even `OpTransformer1.transform` is broken now since it relies on `.select` to pass back the metadata into the output dataframe. `SelectedModelCombinerTest` [tests](https://github.com/salesforce/TransmogrifAI/blob/90a0f298f14506a27c84a71de414d53a30cf687f/core/src/test/scala/com/salesforce/op/stages/impl/selector/SelectedModelCombinerTest.scala#L98) the `.transform` function directly and also fails for the same reason.

tovbinm:
`StructField` [still has the metadata in it](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/types/StructField.scala#L43), it's just `ExpressionEncoder` in Spark 3.x does not allow passing it anymore. Oh, it's a true bummer. We rely heavily on this feature.

core/build.gradle
gerashegalov: When you finish this PR you can aim at an e2e on GPU throughout data [transformations](https://github.com/NVIDIA/spark-rapids) and [Xgboost](https://github.com/NVIDIA/spark-xgboost-examples) :sunglasses:

TuanNguyen27: yes @gerashegalov i was trying to get people to use rapids ;)

gerashegalov: On the Spark side we require 3.x , so you are on the right track @TuanNguyen27

hedibejaoui:
Hello, any estimation on when we can get this PR ready? Thank you!

nicodv:
@hedibejaoui , we are running internal forks of TransmogrifAI and MLeap on Spark 3.1.1, so the bulk of the work has been done.

For public release, the MLeap dependency needs to be upgraded now that they're on Spark 3 too: https://github.com/combust/mleap/pull/765

But since they've upgraded to Spark 3.0.2 and TransmogrifAI to 3.1.1, we have some testing left to do.

hedibejaoui:
@nicodv Thanks for the information. Actually, we are using Spark 3.0.x because of some internal dependencies, any chance we get a public release of TransmogrifAI for that version?

Fatma-abdel:
Hello, When do you think this PR will merged for the public use? Thank you!

EhsanSadr:
Hi, 
This PR adds important functionality that I need for my project. When will this PR merge ?

Thank you

MeriamAffes:
Hi, we are waiting for the new PR adds. When it will be available ? Thanks

PR #552, opened 2021-04-22
Cr/fixed codecov ver

crupley:
**Related issues**
Refer to issue(s) addressed in this pull request from Issues page.

As the result of the recent CodeCov security breach, it would be preferred to use a fixed version rather than automatically upgrade to another potentially insecure version.

Add the CodeCov script to repo and execute the static version.

This is the advised solution from CodeCov in response to my email request:

    We are working on a new version that will compile into binaries. For now you can vendor the bash script into your repo and execute it there. That way it will not change until/unless you update it.
    Ticket: https://codecov.freshdesk.com/helpdesk/tickets/4907

**Additional context**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,9 +7,6 @@ machine-config: &machine-config
     TERM: dumb
     GRADLE_OPTS: -Xmx256m
 
-orbs:
-  codecov: codecov/codecov@1.1.4
-
 shared-steps:
   open-jdk: &open-jdk
     run:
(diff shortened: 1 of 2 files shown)


nicodv (approved):
LGTM

tovbinm:
@crupley is this the official bash script? It's worth mentioning in the ticket where the script is coming from.

crupley:
> @crupley is this the official bash script? It's worth mentioning in the ticket where the script is coming from.

@tovbinm Yes, this is the same script that was being used previously. I've updated the PR description to reflect this.

PR #560, opened 2021-08-11
Unit test for auroc=1 when TP=0

mrunfeldt:
Added a unit test that captures a scenario that was _suspected_ to be a bug in BinaryClassification metric calculation. Verified that there is no bug, behavior is expected.


crupley (approved):
LGTM

nicodv (approved):
LGTM

